PR #38, opened 2017-02-13
source path

richardalligier:
Hi,

I made a modification on where the .ptx and .cl files are taken. One can specify where to look using the ``set_source_path`` method. This can be useful to package a library in a "cleaner" way.

Here is an example, where we use of the kernel ``foo`` in a library ``bar``.
```
kernel foo: int -> unit
let path_lib = Findlib.package_directory "bar" in
foo#set_source_path path_lib;
foo#reload_sources();
```

If ``set_path_source`` is not called, the expected behaviour is not changed by this commit.
By the way, I did not changed ``cuda_load`` nor ``opencl_load``. These two use also ``Sys.argv.(0)``. A modification of these functions might be necessary.


mathiasbourgoin:
Commit 175fbfeb4093871ab02e2a037ad28c68e0a86fb6 has been merged into master

PR #148, opened 2026-06-02
Phase 0B PR-2: pure intrinsic registry + CUDA sinf fix

mathiasbourgoin:
## Phase 0B PR-2 of 5 — pure-codegen-rollout

Makes the intrinsic registry resolvable from the pure (no-`Device.t`) codegen path, and lands the intake-approved CUDA `sinf` fix.

### Changes
- **`spoc/ir/Sarek_pure_registry.ml`** (new, in `sarek_ir`, ctypes-free): maps `(module-path, name) -> (framework:string -> string)` device-name templates. Populated via a **static table** covering `Float32`/`Float64`/`Math` math intrinsics.
- **All 4 generators** (`Sarek_ir_{cuda,opencl,metal}.ml`, `Sarek_ir_glsl.ml`): for path-qualified intrinsics (`path <> []`) query the pure registry first, passing the backend framework string; otherwise fall through to the existing arms. Thread-index/atomic/barrier intrinsics remain handled by each generator's own switch.
- **`sinf` fix** (intake-approved behavior change): path-qualified `Float32.sin` now emits CUDA `sinf` (correct single-precision) instead of `main`'s erroneous double `sin`. OpenCL/Metal/GLSL keep `sin` (those languages don't suffix) — so the only behavior change is CUDA.
- **Goldens**: new `float32_sin_path` kernel across all 4 backends (CUDA `sinf`, others `sin`). Pre-existing `sin_kernel` (path=`[]`) goldens byte-unchanged.
- **e2e**: `test_float32_sin_pure.ml` (new) — builds IR directly, runs `Float32.sin` on device, numeric check vs OCaml `sin` (tol 1e-4). **PASSES on Vulkan / RX 7900 XTX.**

### Scope boundary (per brief)
No library split (PR-3), no `sarek_codegen` extraction (PR-4), no `Sarek_transpile` (PR-5). Native/interpreter behavior identical.

### Authorized escalation
The brief preferred extending the `%sarek_intrinsic` PPX to emit pure registrations alongside FFI ones (single source of truth). That proved larger than PR-2's bound, so the **static-table fallback** was used as the brief explicitly authorized. This leaves the FFI table (`Sarek_stdlib/*.ml`) and the pure table as two hand-maintained sources — **tracked follow-up: fold the pure registration into the PPX in PR-3/4 and delete the hand table.**

### Gates
- `@sarek/tests/runtest` — pass (goldens incl. unchanged `sin_kernel` on 4 backends)
- `@sarek-vulkan/all` — pass
- `test_float32_sin_pure --vulkan` — PASSED on RX 7900 XTX
- `test_math_intrinsics --native` / `--interpreter` — pass (FFI path unchanged)
- `ocamlformat --check` — clean
- (`-lnvrtc` full-build link error pre-existing; CI e2e-fast matrix-mul segfault known-flaky.)

## Summary by CodeRabbit

* **New Features**
  * Added a pure-registry for path-qualified intrinsics and updated intrinsic dispatch to prefer registry-resolved device names across CUDA, Metal, OpenCL, and GLSL.

* **Tests**
  * Added golden and end-to-end GPU tests validating Float32 math intrinsic codegen and runtime correctness.

* **Chores**
  * Wired the new registry into the build module list so it’s available to all backends.

### DIFF
--- a/sarek-cuda/Sarek_ir_cuda.ml
+++ b/sarek-cuda/Sarek_ir_cuda.ml
@@ -231,6 +231,30 @@ and gen_intrinsic buf path name args =
   let full_name =
     match path with [] -> name | _ -> String.concat "." path ^ "." ^ name
   in
+  (* For path-qualified intrinsics, query the pure registry first.
+     This handles Float32.sin -> sinf on CUDA, sin on others, etc. *)
+  let pure_registry_hit =
+    match path with
+    | [] -> None
+    | _ ->
+        let framework = Option.value ~default:"CUDA" !current_framework in
+        (match
+           Sarek_pure_registry.fun_device_template ~module_path:path name
+         with
+        | Some f -> Some (f ~framework)
+        | None -> None)
+  in
+  match pure_registry_hit with
+  | Some device_name ->
+      Buffer.add_string buf device_name ;
+      Buffer.add_char buf '(' ;
+      List.iteri
+        (fun i e ->
+          if i > 0 then Buffer.add_string buf ", " ;
+          gen_expr buf e)
+        args ;
+      Buffer.add_char buf ')'
+  | None ->
   (* Try thread intrinsics *)
   if
     List.mem
--- a/sarek-cuda/Sarek_ir_cuda.ml
+++ b/sarek-cuda/Sarek_ir_cuda.ml
@@ -236,11 +236,11 @@ and gen_intrinsic buf path name args =
   let pure_registry_hit =
     match path with
     | [] -> None
-    | _ ->
+    | _ -> (
         let framework = Option.value ~default:"CUDA" !current_framework in
-        (match
-           Sarek_pure_registry.fun_device_template ~module_path:path name
-         with
+        match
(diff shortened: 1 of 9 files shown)


mathiasbourgoin:
Addressed CodeRabbit review in 37e32240:
- **Math.Float32 completeness** — added the 5 missing entries (`abs_float`, `expm1`, `log1p`, `hypot`, `copysign`) so `Math.Float32.*` mirrors `Float32.*`.
- **var_id** — distinct ids in the e2e IR builder.
- **tolerance comment** — now matches the 1e-4 check.
- **Backend_error** — logged on SKIP via `to_string`.

**Deferred (tracked):** Float32 `min`/`max` → `fminf`/`fmaxf` on CUDA vs `min`/`max` elsewhere has differing IEEE NaN semantics, but `min`/`max` is unexercised by any test — changing it without a golden would be unverified. Folded into the PR-3/4 PPX-dual-registration follow-up alongside the FFI-table/pure-table single-source-of-truth work.

PR #218, opened 2026-07-08
feat(interp): Addition of math intrinsic functions

matburnx:
# Overview

This PR introduces support for mathematical intrinsic operations within the interpreter. It adds mappings for mathematical trigonometric primitives (such as `atan`, `cosh`, `sinh`, `atan2`...) as well as additional mathematical primitives in `float64` (`min`, `max`, `floor`, `ceil`, `pow`).

Such primitives can be used in signal processing or spatial simulations.

## Summary by CodeRabbit

* **New Features**
  * Expanded math support for floating-point values, including inverse trig, hyperbolic, and additional common math operations.
  * Added more built-in handling for `Float32` and `Float64` calculations, with support for both single- and two-argument forms where applicable.
* **Bug Fixes**
  * Improved error reporting when math functions are called with the wrong number of arguments, making unsupported cases clearer.

### DIFF
--- a/sarek/interp/Sarek_ir_interp_intrinsics.ml
+++ b/sarek/interp/Sarek_ir_interp_intrinsics.ml
@@ -133,6 +133,55 @@ let eval_float32_math_intrinsic name args =
           Interp_error.raise_error
             (Unsupported_operation
                {operation = "tan"; reason = "requires 1 argument"}))
+  | "asin" -> (
+      match args with
+      | arg :: _ -> Some (VFloat32 (F32.asin (to_float32 arg)))
+      | [] ->
+          Interp_error.raise_error
+            (Unsupported_operation
+               {operation = "asin"; reason = "requires 1 argument"}))
+  | "acos" -> (
+      match args with
+      | arg :: _ -> Some (VFloat32 (F32.acos (to_float32 arg)))
+      | [] ->
+          Interp_error.raise_error
+            (Unsupported_operation
+               {operation = "acos"; reason = "requires 1 argument"}))
+  | "atan" -> (
+      match args with
+      | arg :: _ -> Some (VFloat32 (F32.atan (to_float32 arg)))
+      | [] ->
+          Interp_error.raise_error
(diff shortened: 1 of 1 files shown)


mathiasbourgoin (approved):
Thanks @matburnx! Reviewed and tested locally.

Verified:
- Branch builds cleanly, full build OK.
- Ran `sarek/tests/e2e/test_float64_math_intrinsics.exe` on this branch vs `main`: coverage goes from 20/115 to 42/115 (device × function) combinations, no regressions. All the newly mapped intrinsics now pass on the Interpreter backend.
- Naming matches the PPX registry (`Sarek_core_primitives`) and the GPU codegens (CUDA/OpenCL/Metal/GLSL/WGSL already map these), so this fills a real interpreter gap.

Two small observations, neither blocking:

1. `min`/`max` (float64) use OCaml's polymorphic `min`/`max`, whose NaN behavior is asymmetric (`min 1.0 nan` → `nan`, but `min nan 1.0` → `1.0`) and differs from GPU `fmin`/`fmax` (which return the non-NaN operand). The existing float32 arms behave the same way, so this is consistent with the file — just noting the cross-backend divergence on NaN inputs in case we want a `min_num`-style semantics pass later.

2. Still uncovered on the interpreter after this PR: `expm1`, `log10`, `log1p`, `rsqrt`, `abs_float`, `hypot`, `copysign` (the remaining ERRORs in the e2e report). Fine as a follow-up — mentioning it so it doesn't get lost.

mathiasbourgoin:
Merci @matburnx — ce travail est intégré à main via #234 (cherry-pick de tes 3 commits, paternité préservée, + complétion des 7 fonctions f64 restantes expm1/log10/log1p/rsqrt/abs_float/hypot/copysign pour que l'oracle de l'interpréteur couvre toute la surface Float64 des backends GPU). Il était bloqué ici par un check CI obsolète et une branche de fork non poussable de notre côté. Je ferme celle-ci au profit de #234 (déjà mergée).

PR #220, opened 2026-07-22
PTX emitter: shared arrays, conversions, int32 atomics, helper inlining

mathiasbourgoin:
## Overview

Closes the three PTX-emitter coverage gaps identified by ZLUDA benchmarking (follow-up to #219), plus two soundness fixes they made observable. The CUDA/PTX backend now runs the reduction/sort/fractal/transpose benchmark families.

## What's new

1. **Shared arrays** — `let%shared` lowers to `SLet (v, EArrayCreate (elt, N, Shared), _)` (the PPX never emits `DShared`); the emitter now declares these in `.shared` space and reuses the existing 32-bit shared addressing.
2. **Conversion intrinsics** — `float`, `of_int`, `int_of_float`, `float64_of_int`, … delegate to `emit_cast`; `of_int` picks f32/f64 from the module path.
3. **int32 atomics** — `atom.{shared,global}.add.s32` for `atomic_add_int32` / `atomic_add_global_int32`.
4. **Helper functions (`EApp`)** — inlined at the call site with a recursion guard (recursive helpers keep the old typed-error fallback). PTX `.func` would need a per-function register frame the single-pass emitter doesn't model, and NVCC inlines these anyway.

## Soundness fixes

- **Register aliasing**: `let mutable acc = y` (and inlined scalar params) bound the initializer register directly, so `acc <- …` clobbered `y`. Now copied via `copy_reg`.
- **Effectful `if` expressions**: the eager evaluate-both-branches + `selp` strategy was sound only while PTX expressions were pure; branches containing atomics or helper calls now emit real control flow (pure branches keep `selp` — golden output unchanged).

## Validation (RX 7900 XTX via ZLUDA v7-preview.3)

| Benchmark | CUDA/PTX before | after |
|---|---|---|
| dot_product, reduction, reduction_max | rejected (EArrayCreate) | **verified ✓** |
| histogram | rejected (EArrayCreate + atomics) | **verified ✓** |
| scan, radix_sort | rejected | **verified ✓** (bonus) |
| mandelbrot | rejected (`float` intrinsic) | **verified ✓**, fastest backend (0.41 ms vs 0.43 Vulkan) |
| transpose, transpose_tiled | rejected (EApp) | **verified ✓** |

- ZLUDA e2e unsupported-kernel errors: **29 → 10** (all remaining are the documented records/variants gap).
- Full suite green without ZLUDA; byte-exact PTX golden unchanged; 4 new snapshot tests (shared decl, inlining without `.func`, `cvt` emission, recursive-helper rejection).
- Residual (out of scope): `bitonic_sort` still hits a ZLUDA runtime memory fault (host address in the fault message — same class as the stream_triad@50M fault; ZLUDA bug, not codegen).

## Summary by CodeRabbit

* **New Features**
  * Added shared-memory array support in PTX generation, including declarations, accesses, and synchronization.
  * Inlined non-recursive helper calls into generated kernels.
  * Improved conditional codegen to use real branching when needed for guarded evaluation.
  * Extended PTX intrinsics with float conversions and int32 atomic add.
* **Bug Fixes**
  * Prevented register aliasing in mutable bindings.
  * Improved benchmark output filenames by sanitizing device identifiers and avoiding collisions.
* **Tests**
  * Expanded PTX snapshot coverage for shared memory, helper inlining, float conversion, conditional branching/guarding, and rejection of recursive helpers.

### DIFF
--- a/benchmarks/bench_mandelbrot.ml
+++ b/benchmarks/bench_mandelbrot.ml
@@ -176,15 +176,22 @@ let benchmark_device dev size config =
 
     (* Generate image if requested *)
     (if config.generate_images then
-       let safe_name =
-         String.map (fun c -> if c = ' ' then '_' else c) dev.Device.name
+       let sanitize s =
+         String.map (fun c -> if c = ' ' || c = '/' then '_' else c) s
+       in
+       (* Distinct devices can sanitize name+framework to the same string
+          (e.g. differing only in '/'); a hash of the raw identity keeps the
+          filenames unique. *)
+       let tag =
+         Hashtbl.hash (dev.Device.name, dev.Device.framework) land 0xffff
        in
        let img_filename =
          Printf.sprintf
-           "%s/mandelbrot_%s_%s_%dx%d.ppm"
+           "%s/mandelbrot_%s_%s_%04x_%dx%d.ppm"
            config.output_dir
-           safe_name
-           dev.Device.framework
+           (sanitize dev.Device.name)
+           (sanitize dev.Device.framework)
+           tag
            width
            height
        in
--- a/sarek/codegen/Sarek_ir_ptx_expr.ml
+++ b/sarek/codegen/Sarek_ir_ptx_expr.ml
@@ -23,6 +23,42 @@ let is_f64_reg r = String.length r >= 3 && r.[1] = 'f' && r.[2] = 'd'
 
 let is_f32_reg r = String.length r >= 2 && r.[1] = 'f' && not (is_f64_reg r)
 
+(** Whether an expression must NOT be evaluated speculatively — i.e. an [EIf]
+    with such a branch must emit real control flow rather than the eager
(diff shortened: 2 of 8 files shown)


mathiasbourgoin:
Addressed CodeRabbit review: the guarded-array-read-in-if-expression OOB (generalized the branch-guard predicate to treat array reads as non-speculative, with a snapshot test), and collision-proof mandelbrot image filenames. Note the reviewer's guess that this caused the bitonic ZLUDA fault is incorrect — that was the kernarg-lifetime UAF fixed in #221 (bitonic uses a statement `if`, already branch-lowered); this fix addresses a distinct latent OOB in if-*expressions*.

mathiasbourgoin:
Fixed: `block_barrier`/`warp_barrier`/`memory_fence` are now treated as needing a branch guard in `expr_needs_branch_guard`, so a barrier/fence in an if-expression branch is never emitted on the not-taken path. History squashed to one commit per PR (rebased #222 on top).

PR #221, opened 2026-07-22
fix(cuda): retain kernel-arg buffers until the launch stream drains

mathiasbourgoin:
## Summary

Fixes an intermittent GPU page fault (host-range address) on repeated rapid kernel launches under ZLUDA — root-caused to a **use-after-free of the host kernel-argument buffers**, not codegen or argument marshalling.

## Root cause (captured under `AMD_LOG_LEVEL=4`)

`cuLaunchKernel` is asynchronous. The NVIDIA driver copies `kernel_params` synchronously before returning, so SPOC freed the host arg buffers (the `CArray` of arg pointers + per-arg ctypes cells) as soon as `Kernel.launch` returned — legal per the CUDA contract. **ZLUDA/HIP instead reads `kernel_params` later, at dispatch time.** Once the OCaml GC reclaimed those cells, the GPU read recycled host memory as the kernel's arguments.

On a faulting `bitonic_sort` launch, ROCr wrote `Arg0 = 0x55f137aad4e0` (a host-heap address from SPOC's own kernarg arena) instead of the device pointer; the GPU faulted at `0x55f137aad000` — the page of that corrupt argument. The values were always correct at the `hipModuleLaunchKernel` boundary, and the fault was timing-dependent (a per-launch `stderr` flush or `HIP_LAUNCH_BLOCKING=1` hid it) — confirming a ZLUDA/HIP kernarg-lifetime issue.

## Fix

Retain `(params, refs)` in a mutex-guarded `pending_kernargs` list keyed by `(device_id, stream_key)`, released when the stream that ran the launch is drained:
- `Device.synchronize` → `retire_device` (full context drains every stream)
- `Stream.synchronize` → that stream's key only
- blocking memcpys + `device_to_device` → null-stream key (what SPOC uses)
- `Device.destroy` → the device's entries

**No added GPU synchronization** — async overlap is preserved (bitonic @4096 stays ~1.9 ms vs ~4.2 ms under `HIP_LAUNCH_BLOCKING=1`). Keyed by stream so an explicit stream's args are never freed by a sibling stream's sync.

## Validation (RX 7900 XTX / ZLUDA v7-preview.3)

- `bitonic_sort` faulted **~45% of runs at 32768** (and intermittently at 16384/65536/131072) before → **0 faults across 30+ runs and all sizes** after.
- Full test suite green **with and without** ZLUDA; `dune build @fmt` clean.
- Adversarially reviewed twice; a first design (device-keyed) that would have freed sibling-stream args early was caught and revised to stream-keyed + mutex.

## Known minor (documented, non-blocking)

An explicit stream synchronized *only* via events over-retains its arg buffers (memory-safe over-retention, not a UAF; outside SPOC's default null-stream flow).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of asynchronous CUDA kernel launches by preserving kernel argument data until GPU execution completes.
  * Prevented potential memory access errors during deferred GPU reads, including on ZLUDA/HIP environments.
  * Added safer cleanup after stream synchronization, device synchronization, blocking transfers, and device destruction.

### DIFF
--- a/sarek-cuda/Cuda_api.ml
+++ b/sarek-cuda/Cuda_api.ml
@@ -51,6 +51,55 @@ let check (ctx : string) (result : cu_result) : unit =
     without a circular module dependency. *)
 let device_destroy_hooks : (int -> unit) list ref = ref []
 
+(** Host-side kernel-argument buffers (the [CArray] of argument pointers plus
+    the per-argument value cells) for launches that may still be in flight,
+    keyed by device id.
+
+    [cuLaunchKernel] is asynchronous. The NVIDIA driver snapshots the kernel
+    parameters synchronously before returning, so a caller may free them right
+    after the call — but the ZLUDA/HIP stack reads them later, at dispatch time.
+    Freeing them when [Kernel.launch] returns is therefore a use-after-free once
+    the OCaml GC reclaims the cells: the GPU then reads recycled host memory as
+    the kernel's arguments (observed as an intermittent GPU page fault at a
+    host-range address on ZLUDA).
+
+    We retain the buffers here until the stream that ran the launch is drained.
+    Each entry is keyed by [(device_id, stream_key)] where [stream_key] is the
+    raw address of the stream handle (0 for the default/null stream, which is
+    what SPOC uses). Values are kept as [Obj.t] purely for liveness and are
+    never inspected. A mutex guards the list so launches/retires from different
+    OCaml domains cannot lose an entry (drop = use-after-free) or a removal. *)
+let pending_kernargs : (int * nativeint * Obj.t) list ref = ref []
+
+let pending_lock = Mutex.create ()
+
+let retain_kernargs device_id stream_key keepalive =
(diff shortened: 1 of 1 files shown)


mathiasbourgoin:
Addressed both findings:

- **current_id domain-local** — done (`Domain.DLS`), mirroring CUDA's thread-local current context. A shared ref could mis-attribute a null-stream launch across domains.
- **Serialize retention with draining** — the described window (a concurrent drain retiring a key between `cuLaunchKernel` returning and the keepalive insert) is a *memory-safe over-retention* (leak), not a use-after-free: the completed launch's buffers are simply held until the next drain. It only occurs under concurrent multi-domain launch+synchronize on the same device, which SPOC does not do (all GPU work is issued from one domain on the default stream). Retaining *before* the launch would instead open a genuine UAF (a drain between retain and submit would free args the not-yet-submitted kernel will read), and locking across the CUDA calls would serialize all launches and kill stream overlap. Keeping the safe ordering and documenting the bounded multi-domain leak.

PR #226, opened 2026-07-22
feat(ptx): depth-bounded recursive helper inlining via sarek.inline pragma

mathiasbourgoin:
Follow-up 2 of the aggregate work: closes the last remaining ZLUDA e2e gap (`test_inline_pragma`).

- **PTX inliner:** recursive helpers carrying `pragma ["sarek.inline N"]` are depth-unrolled — each re-entry consumes one unit of a per-helper budget (restored on exit, so sibling calls like fib's two see the same depth). At exhaustion the residual call's arguments are still evaluated and its result is a typed zero: the pragma is the author's contract that N levels cover all runtime inputs, so the residual site is dynamically unreachable and only needs to be well-formed PTX. Recursion without the pragma stays rejected, with the pragma suggested as the workaround.
- **PPX fix:** `Sarek_tailrec.transform_kernel` already pre-unrolls N times but stripped the pragma, leaving residual recursive calls indistinguishable from unbounded recursion. It now re-wraps the unrolled body in `pragma ["sarek.inline 0"]` — semantically exact (zero remaining budget → residual calls lower as unreachable typed zeros). C-family backends emit `#pragma sarek.inline 0`, ignored per C99; all other consumers pass pragmas through.

e2e under ZLUDA (RX 7900 XTX, CUDA/PTX): pow2(5)=32 and fib(0/1/5/7)=0/1/5/13 all PASS; inliner regression checks (ktype_helper, nbody_ppx) stay green. fib at depth 6 lands at ~3.8k virtual registers — well within ptxas/ZLUDA territory. Snapshot suite extended (depth-3 unroll markers; no-pragma rejection unchanged).

With this, the full ZLUDA e2e sweep has **zero** `generate_source returned None` failures left — every construct either lowers or reports a precise unsupported message (the f64 transcendentals report, by design).

## Summary by CodeRabbit

* **New Features**
  * Added depth-limited inlining for recursive helpers using the `sarek.inline N` annotation.
  * Recursive helper expansion now respects the requested depth limit and ensures well-formed, zero-valued results when the limit is reached.
  * Enhanced PTX generation for recursive helpers to better support inlined/unrolled execution.

* **Tests**
  * Added a new PTX snapshot test covering a recursive helper annotated with `sarek.inline 3`, verifying unrolling to depth and absence of standalone function emission.

### DIFF
--- a/sarek/codegen/Sarek_ir_ptx_expr.ml
+++ b/sarek/codegen/Sarek_ir_ptx_expr.ml
@@ -244,12 +244,23 @@ let emit_match_arms buf alloc (env : env) (scrut : binding)
     truth for the "sarek.inline N" string format) — the PPX is a separate
     library not linkable from the codegen. *)
 let helper_inline_budget (hf : helper_func) : int option =
+  (* A negative depth would never reach the [Some 0] exhaustion arm of
+     [emit_app_recursive] (the budget decrements past zero), making the
+     expansion non-terminating — reject it loudly instead of parsing it. *)
+  let checked n_str =
+    match int_of_string_opt n_str with
+    | Some n when n < 0 ->
+        unsupported
+          ("pragma [\"sarek.inline " ^ n_str
+         ^ "\"]: the inline depth must be >= 0")
+    | v -> v
+  in
   let parse = function
     | [opt] -> (
         match String.split_on_char ' ' opt with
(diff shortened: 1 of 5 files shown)


mathiasbourgoin:
@coderabbitai review

mathiasbourgoin:
@coderabbitai review

PR #243, opened 2026-07-23
L12: Tier-0 defunctionalization of function values in Sarek kernels

mathiasbourgoin:
Implements **L12 Tier 0** (design: `roster/ptx-limits-campaign/L12-defunctionalization.md`): a typed-AST→typed-AST pass (`sarek/ppx/Sarek_defunc.ml`) spliced between monomorphization and lowering that eliminates first-class function *values* bound to `let` variables in kernel/helper bodies.

Before this, the two headline patterns from the design doc §1 fail at codegen with `EApp to unknown function 'f'` / `EApp with non-variable callee`. Now they compile and run:

- `let f = g in f x`  → direct call `g x` (single static candidate, §4.1)
- `let f = if c then g else h in f x` → `if c then g x else h x` (genuinely runtime-dynamic, §4.2)

## Approach — distribution, not a tag variant
The pass distributes the application into the selector's `if`/`match` leaves rather than synthesizing a Reynolds tag variant. Since the §5 escape rules keep a Tier-0 function value in callee position only, distribution is complete for the capture-free world and needs **zero new IR and zero emitter changes** — it reuses ordinary `if`/`match`-expression + direct-call lowering that every backend already has. `if`/`match` arms are mutually exclusive, so arguments are never re-evaluated.

## Escape rules (§5)
New `Function_value_escapes` error: a function-typed value in a record field, tuple/vector element, variant payload, created-array element, assignment, return, or bare non-callee position is rejected with a located diagnostic instead of an opaque downstream failure. Negative test `test_fun_escape`.

## Test matrix — `sarek/tests/e2e/test_defunc.ml` (self-verifying vs OCaml reference)

| Backend | static | dynamic op=0 | dynamic op=1 |
|---|---|---|---|
| CUDA/PTX — RX 7900 XTX (ZLUDA) | PASS | PASS | PASS |
| CUDA/PTX — Ryzen (ZLUDA) | PASS | PASS | PASS |
| OpenCL — RX 7900 XTX / Ryzen | PASS | PASS | PASS |
| Vulkan — RADV NAVI31 / RAPHAEL | PASS | PASS | PASS |
| Native | PASS | PASS | PASS |
| Interpreter (seq + parallel) | PASS | PASS | PASS |

This empirically confirms the design's "zero emitter changes" claim on the real PTX path plus every other available backend.

## Gates
- `dune build @install` clean
- `dune build @fmt` clean on all changed files (pre-existing drift in `Sarek_float32.ml` / `test_ir_layout.ml` untouched)
- `dune runtest sarek/tests` fully green (no regression)
- GPU validated under ZLUDA on CUDA/PTX + OpenCL + Vulkan

## Deferred (documented follow-ups, not started)
- Higher-order **helper-parameter** specialization (`sort ~cmp:my_less`, §4.1/§4.3) — a separable Sarek_mono-style helper-duplication pass; not needed for the headline examples.
- Tier 1 scalar captures / Tier 2 aggregate captures — as designed.

Also surfaced (not fixed, out of scope): a pre-existing typer bug where a `[@sarek.module]` helper doing int32 `+` mis-unifies float32/int32 when used in a kernel (reproduces on a direct call, no function value involved). The test uses kernel-local helpers to avoid it.

## Summary by CodeRabbit

* **New Features**
  * Added a Tier-0 defunctionalization stage to support kernels using locally selected function values, including both static and runtime-dependent selections.
* **Bug Fixes**
  * Improved diagnostics and rejection for escaping function-value usage, preventing invalid IR/codegen paths.
* **Tests**
  * Added an end-to-end defunctionalization test executed across available backends.
  * Added negative compilation coverage for escaping function values, including Makefile verification and Dune integration in the negative test profile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,8 @@ test_negative:
 	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_tuple_param.cma > "$$out" 2>&1; if grep -q "Tuple-typed kernel parameters are not supported" "$$out"; then echo "  PASS: tuple param rejected"; else cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
 	@echo "Testing function-typed kernel parameter rejection..."
 	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_fun_param.cma > "$$out" 2>&1; if grep -q "Function-typed kernel parameters are not supported" "$$out"; then echo "  PASS: function param rejected"; else cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
+	@echo "Testing escaping function-value rejection (L12 defunctionalization)..."
+	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_fun_escape.cma > "$$out" 2>&1; if grep -q "Function value escapes" "$$out"; then echo "  PASS: escaping function value rejected"; else cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
 	@echo "All negative tests checked (see KNOWN-ISSUE line above for the one non-blocking gap, if any)"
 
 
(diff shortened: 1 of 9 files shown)


mathiasbourgoin:
@coderabbitai review

PR #244, opened 2026-07-23
feat(sarek): tuple-typed vectors (L13)

mathiasbourgoin:
## Summary

Implements L13 — `('a * 'b) vector` tuple-typed vectors (Design A from the
validated design doc). A tuple vector element is lowered to a synthesized
packed record (positional fields `_0`, `_1`, ...), reusing the existing
record/aggregate codegen; the host builds the matching `custom_type` with a
new `sarek.tuple_vec` library, layout taken from the shared `Sarek_ir_layout`
authority so host and device agree byte-for-byte.

Before this, a tuple vector param was rejected by the `lower_param` guard (which
had already turned an earlier silent-`TInt32` miscompile into a clear error).

## What changed
- **`Sarek_lower_ir`**: accept `TVec (TTuple prims)`, map the element to
  `Ir.TRecord`; lower tuple literals to `Ir.ERecord` (typed compound literal
  for struct backends); rewrite a single-arm tuple-pattern `match` into
  field-access `let`s (the C-like backends otherwise mis-emit a variant
  `switch (x.tag)`); register the synthesized record in the codegen types table.
- **`sarek.tuple_vec`** (new lib): host `pair`/`triple` `custom_type` builders +
  a process-global shape registry so the Native path shares one `Type_id` with
  generated kernel code.
- **`Sarek_native_gen_base`**: resolve tuple-shape elements to the shared
  registry descriptor (restricted to float32/float64/int32/int64 components).
- **`sarek/tests/e2e/test_tuple_vector.ml`** (new): self-verifying vs a pure
  OCaml reference on every device.

## Test matrix (RX 7900 XTX host)
`(float32 * int32) vector` read-modify-write kernel, checked vs host reference:

| Device | Result |
|---|---|
| CUDA/PTX (ZLUDA) — GPU + CPU | PASS |
| OpenCL — GPU + CPU | PASS |
| Vulkan — GPU + CPU | PASS |
| Native | PASS |
| Interpreter | NOT-YET-SUPPORTED (value-model unification; documented follow-up) |

**7/7 must-pass devices green.** Full `dune runtest sarek/tests` green — no
regression. `@install` and `@fmt` clean on changed files.

## Follow-ups (see design doc §9)
Interpreter support (tuple/record value-model unification), non-primitive tuple
components, and `int`/`char`/`bool` components on Native.

## Summary by CodeRabbit

- **New Features**
  - Added support for vectors containing primitive tuples, including pairs and triples.
  - Added APIs for defining tuple vector element types using supported scalar components.
  - Added host/device layout handling for reading and writing tuple elements.
  - Added tuple vector execution coverage across supported backends.

- **Bug Fixes**
  - Improved tuple lowering and destructuring to preserve consistent memory layout and code generation.

- **Tests**
  - Added end-to-end validation for tuple vectors with floating-point and integer components.

### DIFF
--- a/sarek/ppx/Sarek_lower_ir.ml
+++ b/sarek/ppx/Sarek_lower_ir.ml
@@ -68,6 +68,70 @@ let rec elttype_of_typ (ty : typ) : Ir.elttype =
          not supported. Annotate the parameter with a concrete type (e.g. \
          float32, int32)."
 
+(* L13: tuple-typed vector elements. A tuple used as a vector element type is
+   lowered to a synthesized packed record with positional fields [_0], [_1],
+   ..., reusing the record/aggregate codegen path unchanged (the layout is
+   computed by Sarek_ir_layout, byte-for-byte identically on host and device).
+   Only scalar-primitive components are supported in this tier; nested tuples,
+   records, vectors or functions as components are rejected. *)
+let elttype_prim_tag (e : Ir.elttype) : string option =
+  match e with
+  | Ir.TInt32 -> Some "int32"
+  | Ir.TInt64 -> Some "int64"
+  | Ir.TFloat32 -> Some "float32"
+  | Ir.TFloat64 -> Some "float64"
+  | Ir.TBool -> Some "bool"
+  | _ -> None
+
+let tuple_field_name i = Printf.sprintf "_%d" i
+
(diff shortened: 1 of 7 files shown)


mathiasbourgoin:
@coderabbitai review

PR #255, opened 2026-07-23
fix(glsl): C-truncated integer Mod on Vulkan via single-eval sarek_smod helper

mathiasbourgoin:
GLSL `%` lowers to OpSMod (divisor-signed) on RADV: `-7 mod 2` returned +1 instead of C/OCaml/PTX/OpenCL's dividend-signed -1 (found by #252's signed-arith probe, previously tolerated as DIVERGES).

**Fix (3 review rounds, roster pipeline):**
- `Ir.Mod` on GLSL now lowers to a preamble helper `int sarek_smod(int a, int b) { return a - b * (a / b); }` — function arguments evaluate exactly once, so effectful operands (value-returning atomics, helper calls) are safe. An inline `(a - b*(a/b))` rewrite was rejected in review: it double-fires atomics (proven by mutation: counter=8192 vs 4096, Vulkan-only flip).
- Helper emission gated on `Sarek_ir_analysis.kernel_uses_int_mod` (complete traversal incl. LRecordField-nested lvalue indices — round-3 fix + unit pin, red-on-mutation).
- Probe `test_ptx_signed_arith`: Vulkan flipped from reported-DIVERGES to gated must-pass; 14/14 (incl. b=±1, INT_MIN mod 3) on both RADV devices; INT_MIN/-1 documented as pre-existing signed-Div UB.
- New positive control `test_glsl_mod_single_eval`: atomic-as-mod-operand, counter=n on every device.

No PTX/OpenCL regression (14/14 under ZLUDA). Pipeline: implement→review(NO-GO)→implement→review(NO-GO)→implement→review GO→QA GO.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected GLSL integer remainder semantics to match C/OCaml-style truncated, dividend-signed behavior for negative operands.
  * Prevented `mod` operands from being evaluated more than once, preserving results when operands have side effects.
  * Fixed signed remainder consistency across backends and avoided GLSL helper name collisions.
* **Tests**
  * Added e2e coverage for single-evaluation of `mod` with side effects and for backend-correct signed remainder edge cases.
  * Expanded IR analysis tests and kernel-level detection to cover integer remainder usage inside nested expressions and lvalue indices.

### DIFF
--- a/benchmarks/descriptions/generated/matrix_mul_generated.md
+++ b/benchmarks/descriptions/generated/matrix_mul_generated.md
@@ -72,10 +72,12 @@ layout(push_constant) uniform PushConstants {
 #define n pc.n
 #define k pc.k
 
+int sarek_smod(int a, int b) { return a - b * (a / b); }
+
 void main() {
   int tid = int(gl_GlobalInvocationID.x);
   int row = (tid / n);
-  int col = (tid % n);
+  int col = sarek_smod(tid, n);
   if (((row < m) ? (col < n) : false)) {
     precise float sum = 0.0;
     for (int i = 0; i <= (k - 1); i++) {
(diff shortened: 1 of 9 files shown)


mathiasbourgoin:
@coderabbitai review

PR #274, opened 2026-07-24
feat(sarek-hip): native ROCm/HIP backend (scalar + rocWMMA) with transparent device detection

mathiasbourgoin:
## Summary

Adds a native **ROCm/HIP backend** for Sarek. AMD GPU support already existed via the portable **OpenCL and Vulkan** backends — this backend is **not** about "running on AMD". Its value is specific:

1. **Native replacement for the ZLUDA hack on the CUDA-family codegen path.** The CUDA/PTX codegen previously reached AMD only via ZLUDA; HIP runs it natively through hiprtc JIT (`libhiprtc` → `hipModuleLaunchKernel`), reusing `Sarek_ir_cuda.generate_with_types` verbatim. So the mature CUDA-family code generator (shared-mem tiling, etc.) now targets AMD directly.
2. **Only tensor-core path Sarek implements today (rocWMMA).** The GLSL/Vulkan and OpenCL backends do not *currently* emit to AMD matrix units. NOTE: this is not the *only possible* path — `VK_KHR_cooperative_matrix` is present on RDNA3 (verified: revision 2, `cooperativeMatrix=true`, compute stage on the RX 7900 XTX) and would let the Vulkan/GLSL backend reach the same WMMA units cross-vendor; OpenCL has no such extension. HIP+rocWMMA gives finer matrix-core control + rocBLAS-family access. ⚠️ rocWMMA path **not yet reachable from `[%kernel]`** — hardware PoC only; f16-as-DSL-type is follow-up #57, and that frontend work is backend-independent.
3. **Native-runtime performance edge — narrow and kernel-dependent (full sweep done).** See `docs/benchmarks/hip-vs-opencl-vulkan-2026-07-24.md`. Over 21 comparable benchmarks on the RX 7900 XTX (isolated per-backend runs): HIP wins 13, loses 4, ties 4; geomean 1.25× vs best portable (1.02× = parity on pure DRAM streaming). The earlier '715 vs 467 GB/s' figure was a cross-backend-contention artifact and does NOT reproduce — isolated OpenCL/Vulkan reach ~750-790 GB/s on vector_add. HIP's real wins: launch-bound (bitonic_sort 3.6×, scan 2.7×), LDS/atomics (transpose_tiled 2.2×, reduction 1.9×, histogram), cache-resident (conv2d, stencil). **HIP LOSES on all compute/GEMM kernels** (matrix_mul naive 0.53×, tiled 0.67×, reduction_max 0.79×) — Mesa compilers win by up to 1.9×. Not a blanket win.

Fully additive (+2327/−4).

### What's in it
- **Étape A — scalar backend.** Full `Framework_sig.BACKEND` at CUDA-C parity (alloc/free/H2D/D2H/D2D, streams, events, compile/launch, 13 device intrinsics). Registered in `Device.ml` at priority 80.
- **Étape B — rocWMMA tensor-core GEMM.** RDNA3 (gfx1100) path, hardware PoC; see caveat above.
- **Gate-tolerance fix.** The étape-B rocWMMA correctness gate previously used an absolute tolerance ~5 orders too loose (0.131@256³, 0.515@1024³ — would pass a 51%-wrong result). Now a derived mixed tolerance `|got-ref| ≤ 4·√K·2⁻¹¹·|ref| + floor` against an f64 reference. Red-on-mutation proven on real hardware.
- **Example harness wiring.** `sarek_hip` linked into `benchmarks/` (mirroring the CUDA/OpenCL/Vulkan/Metal `select` + `Backend_hip.init()` idiom, `SPOC_DISABLE_HIP`/`SPOC_DISABLE_GPU` guards). All 22 benchmarks transparently detect and run on HIP.

### Verified on real hardware (AMD RX 7900 XTX, gfx1100, ROCm 7.2.4)
- `test_hip_backend` (étape-A): vector-add + 6 tiled-SGEMM cases pass on both the 7900 XTX and the 7950X iGPU.
- `test_hip_rocwmma` (étape-B): passes on correct output, fails on mutation.
- Transparent auto-detection: bench_vector_add 714.9 GB/s, bench_matrix_mul_tiled 89.3 GFLOPS ✓, bench_reduction 49.99 GB/s ✓ (HIP auto-selected, no override).

### Known caveats (gate fix)
1. The rocWMMA gate's test data (`fill`, multiples of 0.125) are f16-exact, so f16 rounding isn't actually exercised — the tolerance is sized for the precision class but this dataset doesn't stress it.
2. At 1024³ a single wrong tile lands under that size's per-entry ceiling; the aggregate gate still fails via 256³/512³. Correct-by-construction but a per-size 5%-catch at 1024³ would need a larger C.

Skip-clean off-ROCm: both gates print `[SKIP]` and exit 0 when no HIP device is enumerated.

## Summary by CodeRabbit

* **New Features**
  * Added experimental AMD ROCm/HIP GPU backend support as an optional plugin, including device selection, memory transfers, streams/events, and HIPRTC JIT kernel execution.
  * HIP backend is now included in backend initialization and GPU classification, with automatic enable/disable via environment settings.
* **Tests**
  * Added HIP end-to-end backend tests (vector add, GEMM) and rocWMMA correctness coverage.
* **Documentation**
  * Added a HIP vs OpenCL vs Vulkan performance report (with noted measurement caveats).


mathiasbourgoin:
**Re: the `Device.best()` finding (OpenCL preferred over native HIP) — this is intentional, not a defect.**

The full per-backend perf sweep (`docs/benchmarks/hip-vs-opencl-vulkan-2026-07-24.md`, RX 7900 XTX) shows HIP is **not** a blanket win on AMD:
- Wins launch-bound / LDS-atomics / memory-bound kernels: 13/21, geomean 1.25× (bitonic 3.6×, scan 2.7×, transpose_tiled 2.2×, reduction 1.9×).
- **Loses all compute/GEMM kernels 1.5–1.9×** (matrix_mul 0.53×, tiled 0.67×) — verified to be ROCm's LLVM-amdgpu device compiler vs Mesa ACO, **not** a missing `-O3` (tested: hiprtc already optimizes by default; adding `-O3 -ffp-contract=fast` changed nothing).

So blanket-preferring HIP in `best()` would regress GEMM/compute workloads to the slowest of the three backends. `best()` deliberately keeps the conservative incumbent (Mesa/OpenCL) default, which never loses catastrophically. The correct fix is a **kernel-aware device selector** (capability filter + per-kernel perf affinity, seeded from this sweep — HIP for its class, Mesa for compute), which is tracked as follow-up work. Until then HIP is available via explicit device selection. No `best()` change in this PR.

mathiasbourgoin:
**Re: the `Hip_api.ml` kernel-cache concurrency finding — the premise doesn't hold.**

The finding assumes the CUDA sibling guards its kernel cache while HIP doesn't. It doesn't: `sarek-cuda/Cuda_api.ml` (`cache`/`keys_by_device` at 465-470, `with_cache` 629-635, `evict_device` 480-492, `clear_cache` 648-655) uses **no mutex** on the compilation cache either. The only `Mutex` in either backend (`pending_lock`) guards the pending-launch list, not the cache; both track the current device via `Domain.DLS` — consistent with a single-domain execution model for the compile/cache path.

So `Hip_api.ml` faithfully mirrors `Cuda_api.ml`. Guarding HIP's cache alone would diverge it from its sibling. If cross-domain kernel compilation becomes a supported use case, the correct fix is to guard the cache in **all** backends as one change — tracked as a follow-up, out of scope for this PR (which mirrors the established CUDA design). No change here.

The two documentation findings (unverified rows in the aggregate; benchmark-count wording) are addressed in 347091cb.

PR #283, opened 2026-07-25
fix(sarek/codegen): WGSL local shadowing a scalar param no longer reads the push-constant (#72)

mathiasbourgoin:
## Root cause

`gen_expr` decides per-`EVar` whether a name is a scalar kernel param by checking the **global** `scalar_param_names` ref (`Sarek_ir_wgsl.ml` ~line 194), which ignores local scope. A body local `let width` / `let mut width` that **shadows** a scalar param `width` therefore has every body reference emitted as `params.width` (the uniform), while `gen_var_decl` and `gen_lvalue LVar` use the bare name (the local).

- Immutable self-binding local (`let width = params.width`): accidentally correct — both sides read the same value.
- **Mutated** shadowing local: **silent wrong result** (valid WGSL, no error) — writes hit a local nobody reads, reads see the never-updated uniform.

## Fix

Mirror the GLSL backend's `rename_pc_shadowing_locals` with a WGSL-specific `rename_scalar_shadowing_locals` IR pre-pass. Any `SLet`/`SLetMut`/`SFor` binder, plus match pattern binders (`SMatch` and `EMatch`), whose name collides with a scalar param is alpha-renamed to a fresh `sarek_scalar_shadow_<name>_<n>`. The initializer is evaluated in the outer scope so it still expands to `params.<name>`; body reads/writes follow the fresh name; `gen_expr`'s `scalar_param_names` check never matches the renamed local, so the bare fresh name is emitted for both reads and writes. Wired into `generate` and `generate_with_types` with a per-kernel counter reset for deterministic output.

**Why the pre-pass (over scope-threading):** stays consistent with GLSL and reuses the proven, reviewed #65/#71 design. No-op unless a binder genuinely shadows a scalar param, so every existing golden is byte-identical.

**Vector `_len` is N/A for WGSL:** GLSL's `_len` collision is macro-specific (`#define <vec>_len pc.<vec>_len`); WGSL emits vector length as the field access `params.sarek_<arr>_length` (`EArrayLen`), hardcoded with a `params.` prefix independent of any local, so a local cannot alias it. Verified in code, skipped.

## Regression test (codegen-golden, red-on-mutation — naga absent)

`sarek/tests/unit/test_shader_recursion_vector.ml`: a kernel with a **mutated** local shadowing a scalar param (init from a constant, so `params.width` can only appear via the buggy body reads). Asserts the emitted WGSL contains the renamed local `sarek_scalar_shadow_width_1` and does **not** contain `params.width`. Red-on-mutation confirmed: reverting the pre-pass wiring emits `width = (params.width + 1i)` / `out[tid] = f32(params.width)` and the test fails. naga is absent on this machine, so the assertion is on emitted WGSL text, not runtime; were naga present the kernel would also be a semantic positive control.

## Gates

- `dune build` clean; `ocamlformat --check` clean on both touched files.
- Golden codegen tests 75/75 pass, WGSL sweep byte-identical.
- Full `sarek/tests/unit` runtest green (incl. 10/10 shader_recursion_vector).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed WGSL generation when local variables reuse scalar parameter names.
  - Preserved correct variable references across scoped bindings, pattern matches, and loops.
  - Prevented local mutations from being incorrectly redirected to immutable kernel parameters.

- **Tests**
  - Added regression coverage for mutable locals shadowing scalar parameters.
  - Added validation of generated WGSL output, including optional shader validation.

### DIFF
--- a/sarek/codegen/Sarek_ir_wgsl.ml
+++ b/sarek/codegen/Sarek_ir_wgsl.ml
@@ -947,6 +947,159 @@ let gen_bindings buf params =
   Buffer.add_char buf '\n' ;
   List.map (fun (v : var) -> v.var_name) scalars
 
+(** {1 Scalar-param shadow renaming} *)
+
+(** Counter for fresh scalar-param-shadowing local names (see
+    {!rename_scalar_shadowing_locals}). Reset per kernel in {!generate} /
+    {!generate_with_types} for deterministic output. *)
+let scalar_shadow_counter = ref 0
+
+(** Alpha-rename kernel-body binders whose name collides with a scalar kernel
+    param.
+
+    Scalar params are accessed in the body as [params.<name>]; {!gen_expr}
+    decides this per-[EVar] by checking the {e global} [scalar_param_names] ref,
+    which ignores local scope. A local [let width = …] (or [let mut width = …])
+    that shadows a scalar param [width] therefore has every body reference to
+    [width] wrongly emitted as [params.width] — reading the uniform instead of
+    the local. For an immutable self-binding local ([let width = params.width])
+    this is accidentally correct; for a {e mutated} shadowing local it is a
+    silent wrong result (valid WGSL, no error): the declaration uses the bare
+    name ([var width : i32 = params.width;]) so writes hit the local, but every
+    read is redirected to the immutable uniform.
+
+    This mirrors the GLSL backend's {!Sarek_ir_glsl.rename_pc_shadowing_locals}:
+    each colliding binder (and its in-scope references) is rewritten to a fresh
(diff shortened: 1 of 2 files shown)


mathiasbourgoin:
Review note (non-blocking, for future readers): WGSL EMatch (match-expression) lowers to nested select() and discards pattern-payload bindings — the SAME pre-existing limitation as GLSL (tracked in the EMatch fail-loud follow-up). A match-payload binder whose name collides with a scalar param was already silently wrong pre-#72 (read the uniform, payload inaccessible); this pre-pass surfaces that narrow sub-case as a loud undefined-identifier compile error instead. Not a #72 regression — common tag-only/PWild EMatch shapes have no binders and are byte-identical (75/75 goldens).

Issue #44, opened 2019-01-05
My system only has libOpenCL.so.1


tari3x:
When running Samples/src/Game.asm:
```
 ./Game.asm 
dlopen "libcuda.so" failed!
dlopen "libOpenCL.so" failed!
```

My system only has /usr/lib64/libOpenCL.so.1. I created a link and it all worked. 

Is my system broken, or should spoc be trying to load a different file?

PR #78, opened 2026-01-11
feat: Add benchmark results from Intel Arc Graphics + Core Ultra 9 185H

mathiasbourgoin:
## Summary

Adds first benchmark results to populate the interactive benchmark viewer with real-world performance data from Intel Arc Graphics GPU.

## Hardware

- **GPU**: Intel(R) Arc(TM) Graphics (OpenCL + Vulkan)
- **CPU**: Intel(R) Core(TM) Ultra 9 185H (OpenCL)  
- **System**: cachyos

## Benchmarks Collected

All benchmarks run with **20 iterations** and **10 warmup iterations**:

### 1. **Matrix Multiplication (Naive)** - 4 sizes
- Tests compute-bound workloads with arithmetic intensity

### 2. **Vector Addition** - 4 sizes  
- **Peak bandwidth: 80 GB/s** on Intel Arc (OpenCL)
- Tests memory bandwidth ceiling

### 3. **Parallel Reduction (Sum)** - 4 sizes
- Tests shared memory and synchronization
- All results verified for correctness

### 4. **Matrix Transpose (Naive)** - 6 sizes
- Baseline memory access patterns

### 5. **Matrix Transpose (Tiled)** - 6 sizes
- **Up to 26 GB/s** with 16×16 shared memory tiles
- Demonstrates 2-3× speedup from memory optimization

### 6. **Mandelbrot Set** - 4 sizes  
- Tests arithmetic intensity and branch divergence
- Generated PPM images for visual verification

## Results

- **28 JSON result files** in `benchmarks/results/`
- **Updated** `gh-pages/benchmarks/data/latest.json` for web viewer
- Results include full system metadata for multi-machine aggregation

## Preview

Once merged, results will be visible at:
https://mathiasbourgoin.github.io/Sarek/benchmarks/

PR preview will be available at:
https://mathiasbourgoin.github.io/Sarek/preview/pr-<number>/benchmarks/

## Summary by CodeRabbit

* **Chores**
  * Added comprehensive benchmark results for multiple operations (Mandelbrot, matrix multiplication, vector addition, transpose, and reduction algorithms) across different devices and backends.
  * Updated the consolidated benchmark data file with the latest performance metrics, including per-device throughput measurements and timing statistics.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Issue #77, opened 2026-01-11
Update Prism.js to 1.30.0 to fix CVE-2024-53382 (DOM clobbering XSS)


mathiasbourgoin:
## Security Vulnerability

**CVE-2024-53382** affects Prism.js versions ≤ 1.29.0  
**CVSS Score**: 4.9 (Medium)  
**Type**: DOM clobbering leading to potential XSS

## Current State

`gh-pages/benchmarks/index.md` uses Prism.js 1.29.0 (lines 435-441):

```html
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/prismjs@1.29.0/themes/prism.css">
<script src="https://cdn.jsdelivr.net/npm/prismjs@1.29.0/prism.js"></script>
<script src="https://cdn.jsdelivr.net/npm/prismjs@1.29.0/components/prism-json.js"></script>
```

### 1. Update Prism.js to 1.30.0 (Security Fix)

`gh-pages/benchmarks/index.md` (lines ~435-441)

Replace all occurrences of `prism@1.29.0` with `prism@1.30.0`:

```html
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/prismjs@1.30.0/themes/prism.css">
<script src="https://cdn.jsdelivr.net/npm/prismjs@1.30.0/prism.js"></script>
<script src="https://cdn.jsdelivr.net/npm/prismjs@1.30.0/components/prism-json.js"></script>
```

### 2. Optional: Update Chart.js to 4.4.9

While not a security issue, Chart.js 4.4.1 is 8 patch versions behind latest (4.4.9).

Replace `chart.js@4.4.1` with `chart.js@4.4.9`:

```html
<script src="https://cdn.jsdelivr.net/npm/chart.js@4.4.9"></script>
```

## Files to Update

- `gh-pages/benchmarks/index.md` (lines ~435-441)
- Check for other files using Prism.js: `grep -r 'prismjs@' gh-pages/`

## References

- CVE-2024-53382: https://www.cve.org/CVERecord?id=CVE-2024-53382
- Prism.js 1.30.0 release (patched version)
- CodeRabbit security scan from PR #68

## Priority

🔴 **Medium Priority** - XSS vulnerability with CVSS 4.9, but requires attacker-injected HTML

mathiasbourgoin:
Resolved by PR #83 - Prism.js updated to 1.30.0 to fix CVE-2024-53382

PR #285, opened 2026-07-25
DESIGN SPEC FOR REVIEW: f16 (half-precision) DSL element type (#57) — do not merge

mathiasbourgoin:
## This is a DESIGN SPEC, not an implementation — do not merge without review

Adds one file: `docs/design/f16-dsl-element-type.md`. No source changes. The purpose
is to design adding **f16 (half-precision float)** as a Sarek DSL element type — the
shared prerequisite that unlocks every tensor-core path (#62 Vulkan coop-matrix,
#63 Metal simdgroup_matrix, #44 HIP rocWMMA-from-`[%kernel]`).

The spec maps f16 end-to-end (per layer, per backend, with `file:line` anchors) using
float32/float64 as the template, then states a recommendation for each design fork.

### Headline findings
- **Host storage is the EASY part, not the hard part.** Verified by compiling +
  running a probe in this switch (OCaml 5.3.0): `Bigarray.float16_elt` /
  `Bigarray.Float16` exist natively and round to binary16 on store
  (`3.14159 → 3.14062`). This contradicts a stale note in
  `docs/optimization/opt-expressivity-gaps.md:82`; the risk profile flips
  accordingly.
- **The invasive parts are PTX** (register class detected by string-prefix sniffing,
  not a type match) **and breadth** (two IR representations + typer vocabulary,
  ~40 exhaustive `elttype` matches, per-plugin codegen copies).
- Reserved-name groundwork already exists (`float16`, `half*`, WGSL `f16`).

### Recommendations (details in the doc)
- Surface: `float16 vector` annotation (+ optional `half` alias); no f16 literal in slice 1.
- Semantics: storage type, compute in f32, round to binary16 (cross-backend consistency).
- Conversions: explicit `Float16` intrinsic module → `ECast`.
- Gating (#64): add `supports_fp16` + a `kernel_uses_float16` detector + launch gate,
  reusing the exact `kernel_uses_float64` pragma/extension machinery.

### Slice plan
1. host + type system + native/interp + CUDA/HIP (M / low-med).
2. remaining GPU backends incl. PTX (M / med — PTX is the risk).
3. f16 matrix/fragment types = tensor cores #62/#63/#44 (XL / high; **own spec**, out of scope for #57).

### Open questions for the reviewer (spec §10)
1. Is legacy `Kirc_Ast` still load-bearing, or may f16 target only `Sarek_ir`? (Halves cost.)
2. Is slice 3 funded? If not, f16 scalar buys little beyond storage (near-zero current demand).
3. Launch gate: hard error vs warn-and-emulate on f16-lacking devices.
4. `half` alias, or `float16` only?
5. f16 scalar-by-value params in slice 1, or vectors first?

## Summary by CodeRabbit

* **Documentation**
  * Added a design specification for introducing half-precision floating-point (`f16`) support.
  * Documented proposed type-system, storage, conversion, arithmetic, and backend integration requirements.
  * Outlined capability checks, implementation phases, compatibility considerations, and open design questions.
  * Clarified that matrix and tensor-core types are outside the initial scope.

### DIFF
--- a/docs/design/f16-dsl-element-type.md
+++ b/docs/design/f16-dsl-element-type.md
@@ -1,11 +1,64 @@
 # Design Spec: f16 (half-precision float) as a Sarek DSL element type
 
 **Task:** #57 (`f16-dsl-element-type`) — roster RESEARCH + SPEC phase.
-**Status:** DESIGN SPEC FOR REVIEW. No implementation. Design forks are documented
-inline with a recommendation each; the human reviews before any code is written.
-**Date:** 2026-07-25.
+**Status:** DESIGN SPEC, **AMENDED POST-IMPLEMENTATION**. The original text was
+written before any code existed; slice 1 was then built (PR #290) and proved parts
+of this spec wrong. Those parts are corrected **in place**, next to the claim they
+refute, and flagged with a `CORRECTION` callout. The spec is deliberately kept as a
+design document rather than rewritten into an implementation report — the point of
+keeping the wrong claims visible is that the *next* width's spec gets written
+better.
+**Date:** written 2026-07-25; amended 2026-07-25 (post PR #290 / #291).
 **Method:** float32 / float64 traced end-to-end as the template; each layer mapped
 to what f16 needs, with `file:line` anchors verified against the worktree.
+
+**Reading the callouts:**
+
+- **`CORRECTION`** — the spec asserted X, implementation proved Y. Every correction
+  in this document is **VERIFIED BY EXECUTION** (a compiled probe, a run kernel, a
+  compiler diagnostic) unless it explicitly says otherwise.
+- **`HELD`** — a design call the implementation exercised and did not have to revisit.
+- Unmarked prose is the original, reasoned, pre-implementation text. Treat it as
+  reasoned-not-verified unless a callout upgrades it.
+
+**Anchor hygiene.** All `file:line` anchors in the original text were verified
+against the worktree at writing time. They have since drifted: PR #77 factorized
+the C-family codegen, PR #78 (`ba375a36`) deleted `Kirc_Ast.ml` and
+`Sarek_lower.ml` outright, and PR #290 itself moved things. Anchors re-verified
+during this amendment carry their current line; anchors that could not be
+re-verified are marked *(approx.)* rather than guessed at.
+
+---
+
+## Post-implementation corrections — read first
+
+Slice 1 shipped as **PR #290** (`feat/f16-dsl-slice1`, three commits: initial
+implementation, review-round-1 must-fixes, review-round-2 duplication collapse +
+soundness holes). **PR #291 / issue #78** (`ba375a36`) separately deleted the dead
+legacy `Kirc_Ast` path (~2,215 lines removed). Together they change five things in
+this spec:
+
+1. **§3.3 "transfer needs no change" was WRONG.** `Bigarray.Float16` exists
+   (OCaml 5.3, verified) but **Ctypes 0.24.0 has no `Float16` bigarray kind**. Both
+   `typ_of_bigarray_kind Float16` and `bigarray_start Float16` raise
+   `Failure "Unsupported bigarray kind"`. The spec conflated the Bigarray layer
+   (fine) with the Ctypes layer (not fine). See the CORRECTION in §3.3.
+2. **§4's `#include <cuda_fp16.h>` was wrong twice.** It would break every HIP
+   kernel, and it did not work on CUDA either. See the CORRECTION in §4.
+3. **§6.3's `%sarek_intrinsic Float16` module is not buildable as specified** —
+   `%sarek_intrinsic` type registration requires a `ctype`, and Ctypes has no half.
+   See the CORRECTION in §6.3.
+4. **The cost model was wrong in both directions.** The legacy `Kirc_Ast` vocabulary
+   was assumed to be a significant duplication tax; it cost exactly one line, and it
+   no longer exists. The *real* tax is the `Sarek_ir_types` ⇄ `Sarek_ir_ppx` mirror,
+   which turns out to have no dependency justification. See §9.
+5. **§7's capability gating was not built in slice 1** and the shape it recommended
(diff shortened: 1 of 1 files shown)


mathiasbourgoin:
⚠️ **Three claims in this spec were proven FACTUALLY WRONG during slice-1 implementation (PR #290).** Flagging them here so this doc isn't read as-is — the corrections are implemented in #290 and this doc should be amended before it lands.

1. **§3.3 "transfer needs no change" — WRONG.** `Bigarray.Float16` does exist (OCaml 5.3, verified), but **Ctypes has no `Bigarray.Float16` kind**, so `bigarray_start` raises on the first transfer. #290 works around it with a kind-independent address (`Memory.bigarray_void_ptr`, restricted to f16) plus `typ_of_bigarray_kind` in `Hip_api.alloc`. The spec conflated the Bigarray layer (fine) with the Ctypes layer (not fine).

2. **§4 `#include <cuda_fp16.h>` — would BREAK every HIP kernel.** Probed on gfx1100/hiprtc: ROCm ships **neither** `cuda_fp16.h` **nor** `hip/hip_fp16.h` (both fail file-not-found), while bare `__half` / `__float2half` **do compile**. #290 emits the include under a negative `__HIP__` guard instead. Note the CUDA/NVIDIA side of that guard is **not testable on this hardware** — residual risk, reviewed by reading only.

3. **§6.3's `%sarek_intrinsic Float16` module — not buildable as specified.** Its type registration requires a `ctype`, and Ctypes has no half. #290 uses `float16_of_float32` / `float32_of_float16` core primitives special-cased in `Sarek_lower_ir` to lower to `ECast`.

Also superseded by implementation reality: the spec's line anchors predate PR #77's codegen factorization, and the Kirc_Ast cost estimate was too high (a later investigation proved Kirc_Ast is fully DEAD — one rejecting arm was enough, and PR #78 deletes that path entirely).

What held up well: the store-as-f16/compute-in-f32 semantics, the explicit-conversions decision, the slice plan, and the reserved-names groundwork.

mathiasbourgoin:
## Amended post-implementation (`0c340b6d`)

Kept as a design doc rather than closed as superseded — the corrections are integrated **in place**, next to the claim they refute, so a reader cannot reach a wrong claim without its correction. Wrong claims are not deleted: knowing the spec asserted X and reality was Y is how the next width's spec gets written better. A new "Post-implementation corrections" section near the top summarises everything and points at #290 / #291. Every callout is marked `CORRECTION` (verified by execution unless stated) or `HELD`.

**Three factual errors**

1. **§3.3 "transfer needs no change" — wrong, and the spec's largest error.** `Bigarray.Float16` exists (OCaml 5.3) but **Ctypes 0.24.0 has no `Float16` bigarray kind**: `typ_of_bigarray_kind` and `bigarray_start` both raise `Failure "Unsupported bigarray kind"`. The §3.1 probe only exercised the Bigarray layer and the spec generalised from it. Cost: kind-independent helpers in `Memory.ml` plus every alloc/transfer site on CUDA, OpenCL, Vulkan (staging included), Metal (`Metal_api` too) and HIP, plus the missing `Float16` arm in `Execute.ml`'s exec-arg `set` without which f16 could not run on the Interpreter device at all.
2. **§4's `#include <cuda_fp16.h>` — wrong twice.** Would break every HIP kernel (ROCm ships neither `cuda_fp16.h` nor `hip/hip_fp16.h`; bare `__half` compiles — hence the negative `__HIP__` guard), and did not work on CUDA either because `Cuda_nvrtc.ml` passed no `-I`. The generalizable lesson is recorded prominently: **`nvrtc` and `ptxas` are host-side compilers**, so "we have no NVIDIA GPU" never justified leaving generated CUDA uncompiled.
3. **§6.3's `%sarek_intrinsic Float16` module — not buildable as specified.** Type registration needs a `ctype`; Ctypes has no half. Shipped as core primitives special-cased in `Sarek_lower_ir` to lower to `ECast`.

**The cost model was wrong in both directions** (§9 — the most useful part for whoever adds bf16). `Kirc_Ast` was priced far too high: 4 constructors, 4 use sites, exactly **one line** forced — and #291 has since deleted it. The real tax was omitted entirely: the `Sarek_ir_types` ⇄ `Sarek_ir_ppx` mirror, which an architect review showed has **no dependency justification** (`sarek/ppx/dune` already links `sarek_ir`, a dependency-free pure library; the two files are identical except for `SNative`'s payload). Cheap first step: share the five leaf vocabularies by type equation, making the corresponding `conv_*` the identity. Measured f16 cost recorded for calibration — ~20 forced match sites (not ~40), 4 round-1 must-fixes, 1 HIGH, 1 MEDIUM; **none of them in a match arm**.

**What held up** is stated just as explicitly: compute-in-f32 semantics, explicit conversions, the slice plan, the `is_numeric`/`is_float` exclusion (three bounded special-case sites), the reserved names.

**Two divergences beyond the review list**, found while reading the doc against the implementation: the recommended `VFloat16` value variant was deliberately *not* added (f16 rides as `VFloat32` — a storage-only type needs no value-domain representation), and §7's `supports_fp16` + launch gate were deferred in favour of static codegen-time rejection.

**New §11 — forward guidance.** bf16 is the next width (f32's 8-bit exponent → no scaling machinery; in `VK_KHR_cooperative_matrix`; RADV-on-7900XTX testability caveat noted). FP8/FP4 deliberately deferred and *not element types* — FP4 is a block format by definition, FP8 without per-tensor scaling hands users silent garbage, and both live in the `wgmma`/`tcgen05` programming model rather than the type vocabulary. And: do **not** build a table-driven `float_width` descriptor from one instance — add bf16 first, and note the ceiling is ~60%, since the GADT arms carry type-level information a value-level table cannot express.

**Anchors.** Re-verified against current mainline. Drifted ones carry their current line (CUDA `cuda_header` 656→631, OpenCL 755-759→713-717, Metal 909→875); dead ones are called out as void (`Kirc_Ast.ml`, `Sarek_lower.ml`, WGSL `:1233` past EOF, `Sarek_ir_ptx_expr.ml:436/440/530`); unverifiable ones are marked approximate rather than guessed at.

PR stays open — this is an amendment, not a merge request.

PR #291, opened 2026-07-25
chore(sarek): retire dead Kirc_Ast lowering/quoting path + orphan test-common library (#78)

mathiasbourgoin:
Retires the legacy `Kirc_Ast` compile-time IR and everything that fed it. Closes #78.

## Why this code was dead, not merely vestigial

`Sarek_quote.quote_elttype` emitted `Sarek.Kirc_Ast.EInt32`, but the `sarek` runtime library never had a `Kirc_Ast` module (see `sarek/sarek/dune`) — the only `Kirc_Ast.ml` in the tree belonged to `sarek_native_gen`. **Any code that path emitted could therefore never have compiled.** Empirically, 0 of 85 PPX-expanded artifacts referenced `Kirc_Ast` (59/63 reference `Kirc_types`, 61/63 `Sarek_ir`). The live pipeline is `Sarek_ppx.expand_kernel` → `Sarek_lower_ir.lower_kernel` → `Sarek_quote_ir`, and there is no `Kirc` module at all — zero repo-wide hits for `Kirc.run`, `Kirc.gen`, `Spoc.Kernel.run`.

## Method: the compiler enumerated the dependents

Rather than bulk-deleting and hoping, `Kirc_Ast` was first dropped from the `sarek_native_gen` module list so the compiler would name every real dependent. It named exactly two — `Sarek_lower` and `Sarek_quote` — matching the prior read-only investigation with nothing unexpected.

## What was deleted

| File | Lines |
|---|---|
| `sarek/ppx/Kirc_Ast.ml` | 297 |
| `sarek/tests/unit/test_lower.ml` (covered only the dead path) | 476 |
| `sarek/tests/common/` — the orphan `sarek_test_common` library | 526 |

The orphan library claim was re-verified independently with a fresh grep across all `dune`, `.ml` and `.mli` files: nothing outside `sarek/tests/common/` named `sarek_test_common`, `Ir_compare`, `Source_gen` or `Test_kernels`. `Source_gen.ml` even emitted `open Kirc` for a module that does not exist. `common` was also dropped from the `(dirs …)` stanza in `sarek/tests/dune`.

## What was trimmed

- **`sarek/ppx/Sarek_lower.ml` → `sarek/ppx/Sarek_ctype_gen.ml`.** The dead lowering (~540 lines) is gone. Its one live half — the C struct/union/builder string generator reached from `Sarek_ppx.expand_sarek_type` — is kept, and the file is renamed so the name matches what it now does.
- **`sarek/ppx/Sarek_quote.ml`.** The Kirc quoting family (`quote_elttype`, `quote_memspace`, `quote_case`, `quote_k_ext`) is gone. The generic quoting helpers and `quote_kernel` remain live.
- **`sarek/tests/unit/test_quote.ml`.** Dropped 4 vacuous tests: they accepted `Pexp_extension _` as a pass, so they asserted nothing.

- **Two further definition-only bindings** (review follow-up, commit 2). `Sarek_quote.quote_array` — killed by this PR's own trim, since both its callers were inside the deleted `quote_k_ext`; the first pass kept it on a *group* claim about the generic helpers that was never checked per symbol. And `Sarek_ctype_gen.c_type_of_core_type`, already dead at base. Neither can be caught by a compiler warning: `sarek_native_gen` is `(wrapped false)` with `(public_name sarek.native_gen)` and neither file has an `.mli`, so every binding is nominally exported and unused-value analysis never fires. Liveness here must be established by per-symbol grep.

**Net: 2,215 lines removed** (2,387 deletions, 172 insertions — the insertions being the surviving half of the renamed file).

## Verification

- Whole-repo `dune build` is unchanged from `origin/main`. On the dev switch used here both refs stop at the same pre-existing failure — the bytecode link of `sarek/transpile/test/test_transpile_proof.bc` (`(modes byte exe)`), where `unix.cma` wants `caml_unix_sigwait` and this switch's `ocamlrun` does not provide it. It is a switch/environment defect, not a repo one, it is unrelated to `js_of_ocaml`, and it reproduces identically at `7c4bbdef`.
- Every `runtest` alias green: `unit`, `codegen_golden`, `new_runtime`, `spoc/{framework,ir,registry}/test`, `sarek/core/test`, `sarek/ppx/test`.
- **Generated code is byte-identical.** All 182 surviving PPX-expanded artifacts hash identically before and after; the only three `.pp.ml` that differ are the edited sources themselves. No golden snapshot was promoted.
- `e2e`: 203 PASSED before and after (forced uncached re-run). All 70 e2e executables exit identically on both refs. The one failure, `test_static_tag_erasure`, is pre-existing and ref-independent: its deterministic defect is a Vulkan/RADV numeric mismatch (`got 0.000 exp 2.000`), and on this particular machine it degrades further into a GPU-driver segfault at the same leg (4/4 runs) that also hits `test_inline_pragma`. Verified identical at `7c4bbdef` by stashing and re-running.
- `@fmt` clean on every touched file (checked individually; the repo's pre-existing drift in `Device.ml`, `Sarek_float32.ml`, `test_ir_layout.ml`, `Sarek_ir_glsl.ml`, `Sarek_ir_ptx_expr.ml`, `test_ptx_snapshot.ml`, `Sarek_ir_softmath.ml`, `test_ptx_stride_spike.ml` is untouched).

## Caveat, on the record

`sarek_native_gen` is `(wrapped false)` with `(public_name sarek.native_gen)`, and `Kirc_Ast.ml` had no `.mli` — so `Kirc_Ast` was **nominally exported** from the installed `sarek` opam package. Removing it is a theoretical API break for an out-of-repo consumer. Such a consumer could not have used the quoting path anyway, since the code it emits does not compile. This break was explicitly greenlit.

## Deliberately left alone

- `Sarek_quote.quote_sarek_type_decl` is also definition-only, but it is dead **at base** and is the visible tip of a separate, production-unreachable `quote_sarek_*` family that is reachable only from `test_quote.ml`. Deleting just the tip would be arbitrary; whether that whole test-only family should go deserves its own liveness argument, so it is not folded in here.
- `gh-pages/spoc_docs/sarek/Sarek_ppx_lib/Kirc_Ast/index.html` is stale generated odoc output, but it is part of a bundle, not a standalone artifact: the sibling `Sarek_ppx_lib/index.html` links to `Kirc_Ast/index.html`, so removing the one file would leave a dangling link, and `Sarek_lower/` is now stale for the same reason. `.github/workflows/docs.yml` regenerates the whole tree from `_build/default/_doc/_html` and publishes it to the `gh-pages` branch, so the committed snapshot has no effect on published docs. Correcting it properly means a full regen, which does not belong in this PR.

## Note for the concurrent f16 branch

`lower_reg_elttype` in `Sarek_lower.ml` — where the f16 work was expected to add a rejecting arm — is inside the deleted region. Once this lands, that arm is unnecessary; the f16 branch can rebase and drop it.

## Summary by CodeRabbit

* **New Features**
  * Added automatic C structure, union, and builder generation for supported record and variant types.
  * Improved handling of type declarations during native code generation.

* **Refactor**
  * Replaced the legacy intermediate lowering pipeline with the updated native type-generation flow.
  * Removed obsolete intermediate representation and test-generation components.

* **Documentation**
  * Updated compilation pipeline documentation to reflect the current generation process.

* **Tests**
  * Removed tests and fixtures covering the retired lowering and legacy AST functionality.


mathiasbourgoin:
**All three findings are PRE-EXISTING code that this PR only relocated — but the Critical one is REAL and now tracked.**

This PR renames `sarek/ppx/Sarek_lower.ml` → `sarek/ppx/Sarek_ctype_gen.ml` and deletes its dead lower half. The surviving body is **byte-for-byte** `Sarek_lower.ml` lines 1–176 at the merge base (`7c4bbdef`); the only edits are the rewritten header comment and dropping two now-unneeded `open`s. Because a rename shows as a large addition, the review reads the relocated code as new. Verified identical on `origin/main`:

- `c_type_of_typ`'s `| _ -> "int"` wildcard — present verbatim at `Sarek_lower.ml:32` on main.
- `typ_of_core_type` not handling `bool` — same, pre-existing.
- **Critical: single-argument variant payload computed then discarded** — verbatim at `Sarek_lower.ml:162-164` on main:
  ```ocaml
  | Pcstr_tuple [ct] ->
      let _ = typ_of_core_type ~loc ct in
      (cd.pcd_name.txt, None)
  ```
  and the consumer maps `None` to `"  int " ^ … ^ "_t;"`, so a `Circle of float` payload field is emitted as `int`.

The Critical finding is **correct and worth fixing** — it is the same silent-`int`-fallback family as the wrong-width bugs already fixed this cycle (PRs #244, #261, #273). It is tracked as a follow-up together with the other two, and it also settles the direction of the pending `Sarek_ctype_gen` / `Sarek_lower_ir` de-duplication (keep whichever twin types payloads correctly).

Not fixing it here, because this PR's contract is *pure deletion with zero behaviour change*, evidenced by: 181 of 184 common `.pp.ml` PPX-expansion artifacts byte-identical (only the 3 edited sources differ), no golden snapshot promoted, and all 70 e2e executables returning identical exit codes on both refs. Changing type mapping here would break that guarantee and hide a real behaviour change inside a 2,215-line deletion.

Dismissing so the deletion can land; the findings are on the record and queued.

PR #279, opened 2026-07-24
refactor(sarek/codegen): split emit_intrinsic_native + remove dead fns + fix odoc (#50)

mathiasbourgoin:
Behaviour-preserving cleanup of the Sarek codegen (task #50). All generated output is byte-identical: golden codegen tests pass with zero diffs and the ptxas gate still assembles the regression kernels.

## Part 1 — split `emit_intrinsic_native` (was 723 lines)
`Sarek_ir_ptx_expr.ml`'s intrinsic emitter is now a **15-line dispatcher** that tries cohesive per-category emitters in order. The categories partition disjoint intrinsic-name sets, so first-`Some`-wins is equivalent to the original single `match`:

| new sub-function | responsibility | approx lines |
|---|---|---|
| `emit_intrinsic_index` | thread/block/grid/global indices, barrier | 92 |
| `emit_intrinsic_transcendental` | sin/cos/tan/exp/log/pow/sinh/…/sqrt/rsqrt, softmath widen | 166 |
| `emit_intrinsic_float_ops` | fabs/copysign/fmod/hypot/fma/min/max/floor/ceil | 91 |
| `emit_intrinsic_bitcast` | f64_bits / bits_f64 | 21 |
| `emit_intrinsic_convert` | float/float64/of_int/to_int casts | 17 |
| `emit_intrinsic_atomic` | all `atomic_*` RMW/CAS/inc-dec | 231 (flat dispatch table) |
| `emit_intrinsic_native` | dispatcher | 15 |

Shared helpers (f32 builders, arg extractors, `intr_atomic_addr/rmw/cas/incdec`, minmax, round, cast, operand checks) are extracted as `intr_*` members of the recursive group. `intr_no_f64` is a module-level `let` so its raising return type stays generalized (a recursive-group binding would monomorphize it and break the `string` vs `string*string` call sites). Every helper and match-arm body is a **verbatim slice** of the original — no logic changed. The public entry point `emit_intrinsic` is unchanged.

## Part 2 — removed 2 dead functions
The audit expected ~3; a rigorous whole-repo grep (word-boundary, excluding `_build`/`.git`/generated `gh-pages` HTML) confirms exactly **2** with zero callers:
- `count_vec_params` (`Sarek_ir_glsl.ml`) — only occurrence was its own definition.
- `generate_with_fp64` (`Sarek_ir_metal.ml`) — never referenced qualified (`Sarek_ir_metal.generate_with_fp64`) nor re-exported.

[...]

PR #292, opened 2026-07-25
fix(sarek/codegen): emit cvt.f64.f32 for exact f32->f64 widening + sweep the ptxas gate over all intrinsics (#87)

### LINKED ISSUE #87: Fix markdown linting issues in benchmark documentation
## Context
CodeRabbit flagged several markdown linting issues (markdownlint) in the benchmark description files.

### 1. Missing language specifiers on fenced code blocks (MD040)
Files affected:
- `benchmarks/descriptions/vector_copy.md` (4 blocks at lines 23, 76, 103, 109)
- `benchmarks/descriptions/stream_triad.md` (6 blocks at lines 53, 80, 89, 127, 144, 179)
- `benchmarks/descriptions/reduction_max.md` (1 block at line 15)
- `gh-pages/benchmarks/descriptions/` (mirrored files)

**Fix:** Add language identifier to code blocks (```text instead of ```)

### 2. Bare URL without markdown link (MD034)
`benchmarks/descriptions/stream_triad.md` line 197

**Fix:** Convert to markdown link:
```diff
-- https://www.cs.virginia.edu/stream/
+- [STREAM Benchmark Homepage](https://www.cs.virginia.edu/stream/)
```

## Priority
Low - cosmetic issue, doesn't affect functionality

## References
- CodeRabbit review: PR #69

mathiasbourgoin:
## The bug (one token)

`emit_cast`'s `TFloat64` arm emitted **`cvt.rn.f64.f32`** for the f32 → f64 widening. That conversion is **exact**, and PTX forbids a rounding modifier on it, so `ptxas` rejects the whole module:

```
line 55; error   : Illegal rounding modifier for instruction 'cvt'
ptxas fatal   : Ptx assembly aborted due to errors
```

Six **public** DSL intrinsics were therefore unusable on the CUDA/PTX backend:

`Sarek_stdlib.Float32.{asin, acos, atan, atan2, expm1, log1p}`

They have no native f32 PTX op, so they reach the widen-to-f64 arm of `emit_intrinsic_transcendental`, generate PTX fine, and then die at `cuModuleLoadData`.

Fix: `cvt.f64.f32`. The narrowing sibling `cvt.rn.f32.f64` and every int → float `cvt` are **inexact**, where the modifier is *required* — confirmed against ptxas, which rejects `cvt.f64.s32` for a *missing* modifier. f32 → f64 is the only exact widening the PTX emitters produce, so it is the only affected site (all `cvt.rzi.*`, `cvt.rmi/rpi`, `cvt.u64.u32`, `cvt.s64.s32`, `cvt.u32.u64` sites re-checked).

**Provenance:** this pre-dates #279. It came in with 80317b39 ("feat(ptx): inverse trig, expm1 and log1p"), which added the six intrinsics; `emit_cast` is not in #279's diff.

## Why it hid

- `test_f32_asin_via_f64` was a **green test asserting the invalid opcode**: `assert_contains ptx "cvt.rn.f64.f32"`. It now asserts `cvt.f64.f32` and additionally asserts the `.rn` form is absent.
- The ptxas gate — whose entire purpose is catching exactly this class — covered five hand-picked kernels (`vector_add`, `int_div_rem_signed`, `int64_compare`, `fmod_intrinsic`, `binop_operand_order`) with no inverse-trig case, i.e. it was bypassed for the whole intrinsic surface. 37 of the 80 intrinsic names reached the PTX emitter in **no** test at all.

## The new gate

`sarek/tests/unit/test_ptx_intrinsic_sweep.ml` assembles **one kernel per intrinsic NAME** — 114 kernels across the 80 dispatched names, ~1.3s of ptxas, so it stays on the default `runtest` alias (no separate alias needed).

- Enumerated from the emitter's own dispatch tables, now exposed as `Sarek_ir_ptx_expr.{index,transcendental,float_ops,bitcast,convert,atomic}_intrinsic_names` — it cannot drift: a new intrinsic is either assembled automatically or fails the "every dispatched intrinsic has a sweep kernel" recipe check.
- Per-name pass/fail reporting, so a failure names the culprit intrinsic.
- Self-skips with a printed reason when `ptxas` is absent (mirrors the existing gate), so the suite stays green off-CUDA.

**Red-on-mutation** (fix reverted) names exactly the six:

```
FAIL asin_f32   ... Illegal rounding modifier for instruction 'cvt'
FAIL acos_f32   ...
FAIL atan_f32   ...
FAIL atan2_f32  ... (two illegal cvts: lines 55 and 60 — the two-arg case)
FAIL expm1_f32  ...
FAIL log1p_f32  ...
ptxas rejected 6 kernel(s) — broken intrinsic(s): acos, asin, atan, atan2, expm1, log1p
```

## Second gap: the dispatcher's unenforced invariant

`emit_intrinsic_native` tried six emitters and took the first `Some`, which is equivalent to the original single `match` **only** while no name is claimed by two emitters — and nothing checked that. Adding e.g. `"rsqrt"` to `emit_intrinsic_float_ops` (where floor/ceil/min/max live), unaware it is already in `emit_intrinsic_transcendental`, would be silently won by list order: no warning, no test failure, wrong lowering. The original `match` would at least have warned about a redundant arm.

Dispatch is now a **name → emitter table lookup** (a double claim raises an explicit internal error) plus a unit test asserting the six name sets are pairwise disjoint. The invariant did hold before this change: 80 distinct names, zero duplicates.

## Gates

- `@sarek/tests/unit/runtest` — green (includes the new sweep)
- `@sarek/tests/codegen_golden/runtest` — green; no golden file contained the old string, so **no snapshot churn**
- `@fmt` — clean on touched files (pre-existing drift in `Device.ml`, `Sarek_float32.ml`, `test_ir_layout.ml` left alone)

Closes #87.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected f32-to-f64 conversion in PTX output to use a valid widening instruction without an unsupported rounding modifier.
  - Improved intrinsic dispatch so duplicate intrinsic ownership is detected and reported clearly.

- **Tests**
  - Added comprehensive PTX intrinsic coverage, including assembly validation when `ptxas` is available.
  - Added checks for complete intrinsic coverage and non-overlapping dispatch categories.
  - Updated snapshots to verify the corrected conversion instruction.

### DIFF
--- a/sarek/codegen/Sarek_ir_ptx_expr.ml
+++ b/sarek/codegen/Sarek_ir_ptx_expr.ml
@@ -419,6 +419,123 @@ let intr_no_f64 intr =
     (intr ^ ": no native f64 " ^ intr
    ^ " in PTX; compute in float32 or on the CPU")
 
+(** {1 Intrinsic name tables}
+
+    The name set OWNED by each per-category intrinsic emitter below. These are
+    the dispatch table consulted by {!emit_intrinsic_native} — not
+    documentation: a name absent from every list is [unsupported], and a name
+    present in two lists is an internal error (the emitters must partition the
+    intrinsic surface, see {!intrinsic_emitter_table}). They also give the ptxas
+    sweep gate a drift-proof enumeration of the whole PTX intrinsic surface
+    (sarek/tests/unit/test_ptx_intrinsic_sweep.ml): an intrinsic added here
+    without an assembling kernel recipe fails that test. *)
+
+let index_intrinsic_names =
+  [
+    "thread_id_x";
+    "thread_idx_x";
+    "thread_id_y";
+    "thread_idx_y";
+    "thread_id_z";
+    "thread_idx_z";
+    "block_id_x";
+    "block_idx_x";
+    "block_id_y";
+    "block_idx_y";
+    "block_id_z";
+    "block_idx_z";
+    "block_dim_x";
+    "block_dim_y";
+    "block_dim_z";
+    "grid_dim_x";
+    "grid_dim_y";
+    "grid_dim_z";
+    "global_thread_id";
+    "global_idx";
+    "global_idx_x";
+    "global_idx_y";
+    "global_size";
+    "block_barrier";
+  ]
+
+let transcendental_intrinsic_names =
+  [
+    "sin";
+    "cos";
+    "tan";
+    "sqrt";
+    "exp";
+    "log";
+    "log10";
+    "pow";
+    "sinh";
+    "cosh";
+    "tanh";
+    "asin";
+    "acos";
+    "atan";
+    "atan2";
+    "expm1";
+    "log1p";
+    "rsqrt";
+  ]
+
+let float_ops_intrinsic_names =
+  [
+    "fabs";
+    "abs_float";
+    "copysign";
+    "fmod";
+    "hypot";
+    "fma";
+    "min";
+    "max";
+    "floor";
+    "ceil";
+  ]
+
+let bitcast_intrinsic_names = ["f64_bits"; "bits_f64"]
+
+let convert_intrinsic_names =
+  [
+    "float";
+    "float_of_int";
+    "float64";
+    "float64_of_int";
+    "int_of_float";
+    "int_of_float64";
+    "of_int";
+    "to_int";
+  ]
+
+let atomic_intrinsic_names =
+  [
+    "atomic_add_int32";
+    "atomic_add_global_int32";
+    "atomic_sub_int32";
+    "atomic_min_int32";
+    "atomic_max_int32";
+    "atomic_and_int32";
+    "atomic_or_int32";
+    "atomic_xor_int32";
+    "atomic_exch_int32";
+    "atomic_exch_int64";
+    "atomic_add_int64";
+    "atomic_add_float32";
+    "atomic_add_float64";
+    "atomic_cas_int32";
+    "atomic_cas_int64";
+    "atomic_inc_int32";
+    "atomic_inc_global_int32";
+    "atomic_dec_int32";
+  ]
+
+(** Every intrinsic name the PTX backend lowers natively, in dispatch order. *)
+let intrinsic_names =
+  index_intrinsic_names @ transcendental_intrinsic_names
+  @ float_ops_intrinsic_names @ bitcast_intrinsic_names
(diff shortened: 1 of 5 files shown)


mathiasbourgoin:
Addressed both cross-runtime review findings in 0c4f94b1.

**1. MUST FIX — generation no longer skipped with ptxas (fixed).** The sweep now ALWAYS enumerates and generates all 114 kernels and fails on any codegen exception; only `assemble_ok` is skipped when ptxas is absent, and the printed note states what is still being checked. Proof that CPU-only coverage is now real: renaming one emitter arm (`"bits_f64"` → `"bits_f64_typo"`) while leaving the name registered fails a CPU-only run with `FAIL exp_f64  codegen raised … unsupported construct: intrinsic: bits_f64` (×4 kernels) — previously green.

I also added a static illegal-opcode scan over the generated PTX (one entry: `cvt.rn.f64.f32`), so the class this gate exists for goes red on CPU-only machines too — no assembler can report it there. Re-confirmed red-on-mutation of the cvt fix on **both** paths:
- with ptxas: `6 intrinsic kernel(s) rejected — broken intrinsic(s): acos, asin, atan, atan2, expm1, log1p` (ptxas: Illegal rounding modifier)
- CPU-only (7ms): same six, `emits illegal cvt.rn.f64.f32 — a rounding modifier on the EXACT f32->f64 widening`

**2. Dual-source registry — chose the cheap check, not the refactor.** Converting the six emitters into name→handler registries means restructuring 28 match arms, several with bodies shared across up to six names (`| "asin" | "acos" | "atan" | "atan2" | "expm1" | "log1p" -> …`), ~600 lines — too large and too risky to fold into a one-token bug fix. Instead: a test that scans the emitter source (declared as a dune dep) and requires arms and tables to agree **in both directions**. Verified red both ways:
- arm not registered: `emit_intrinsic_float_ops has match arm(s) [ceil] that are NOT in float_ops_intrinsic_names — the arm is unreachable…`
- registered with no arm: `bitcast_intrinsic_names declares [f32_bits] with no matching arm in emit_intrinsic_bitcast…`

The registry refactor remains the right end state and is worth tracking separately.

Gates re-run: `@sarek/tests/unit/runtest` and `@sarek/tests/codegen_golden/runtest` green; `@fmt` clean on touched files.

Issue #86, opened 2026-01-11
Add CLI argument parsing to remaining benchmarks


mathiasbourgoin:
## Context
Currently only `bench_matrix_mul_tiled.ml`, `bench_stream_triad.ml`, and `bench_vector_copy.ml` have comprehensive CLI argument parsing with options like:
- `--sizes`: Comma-separated problem sizes
- `--iterations`: Number of timing iterations
- `--warmup`: Number of warmup iterations
- `--output`: Output directory
- `--all-backends`: Run on all available backends

## Issue
The following benchmarks lack CLI parsing and use hardcoded defaults:
- `bench_dot_product.ml` 
- `bench_reduction_max.ml`

## Proposed Solution
Add CLI argument parsing similar to `bench_stream_triad.ml`:
1. Create `parse_args()` function
2. Define `Arg.spec` list with standard options
3. Parse command-line arguments into config
4. Update entrypoint to use parsed config

## References
- CodeRabbit review comment: PR #69
- Good example: `benchmarks/bench_stream_triad.ml` lines 306-339

## Priority
Low - benchmarks work correctly with defaults, this just adds configurability for power users.

PR #293, opened 2026-07-25
fix(sarek/core): key the Runtime kernel cache by device and invalidate it on backend teardown (#86)

mathiasbourgoin:
Fixes six findings in the kernel-compilation cache layers (#86). Every High/Medium below was reproduced by execution on this machine (two OpenCL devices are enumerated — dGPU id 0, iGPU id 1 — which is what makes F1/F2 reproducible), and each fix has a red-on-mutation regression test.

## F1 — High, silent wrong results: the Runtime cache key omitted device identity

`Runtime.compile_kernel` keyed its memo on `(device.framework, name, Hashtbl.hash source)`. `Device.t.framework` is the **backend name** (`Device.ml`: `framework = B.name`), so every device of one backend collided on it, while `Kernel.compile_cached` closes the *specific* backend kernel into the `Kernel.t` it returns — an outer cache hit handed device B a kernel built for device A, with no exception raised. Each backend's own cache already keyed correctly via `Compile_cache.make_key ~device:…`; this outer layer threw that away.

Reproduced: `compile_kernel` for device 1 after device 0 returned a kernel bound to device 0; end-to-end, a kernel writing `7.0` read back `7 7 7 7` on a solo device-1 run and `0 0 0 0` when device 0 warmed the cache first, on the identical device. Fixed at `sarek/core/Runtime.ml:88-106` by building the key with `Compile_cache.make_key` and the device folded in, which also retires the 30-bit `Hashtbl.hash source` (a second, lower-probability wrong-kernel channel) in favour of the MD5-digest convention every other key in the tree uses. Red→green evidence: `test_runtime_cache_device_key` reports `requested device 1, got a kernel bound to device 0` and `16/16 elements were not written` when the device component is removed; all three checks pass with it.

## F2 — High, use-after-free through the public API: the outer memo was never invalidated by teardown

The comment on the cache claimed `destroy` could be a no-op because "clearing this outer layer only drops the memoization". That was backwards. `Kernel.clear_cache device` → `B.Kernel.clear_cache ()` → `Guarded_cache.clear` → `destroy` → `Opencl_api.Kernel.release` / `cuModuleUnload`, while `Runtime.kernel_cache` still held `Kernel.t` values closing over the released handles and served them to the next `Runtime.run`. Same hazard via `device_destroy_hooks` → `Kernel.evict_device`.

Reproduced: `run` → 7; `Kernel.clear_cache d` → OK; `run` again → **SIGSEGV rc=139**.

Fixed by making the outer layer participate in teardown. The backends sit *below* `Runtime` and cannot reference it, so this needs a back-channel: new `spoc/framework/Cache_hooks.{ml,mli}` carries `on_clear_all` / `on_device_destroy` listeners. `Runtime.ml:70-86` registers `clear` and an `evict_device` mapper; `Kernel.ml:157-166` fires `notify_clear_all` **before** the backend releases anything; `Cuda_api.ml:270-277` and `Hip_api.ml:213-220` fire `notify_device_destroy` before their local hooks unload modules. The false comment is replaced with one that states the actual invariant. The device-destroy notification carries a *backend-local* index while the Runtime cache groups by global `Device.id`, so the listener widens rather than narrows — over-eviction of a pure memo is free, under-eviction serves a released handle. Red→green: removing the `notify_clear_all` call makes `test_runtime_cache_teardown` exit 139; with it, exit 0 over two clear/run cycles.

## F3 — Medium: a build survived eviction of the device it was built for

`Guarded_cache.find_or_build` released the lock for the build, so a device destroy landing between the miss and the re-acquire evicted nothing (the entry was not in the table yet); the builder then inserted and recorded the value against a now-dead device id — never unloaded (leak), and served to later lookups, including after the id was recreated (a launch on a module from a destroyed context). `test_concurrent_evict_device`'s conservation assertion is *satisfied* by that interleaving, which is why the suite was green.

Fixed with a per-device eviction epoch, read under the lock at miss time and re-checked on re-acquire; on a change the freshly built value is `destroy`ed instead of installed and `Device_destroyed_during_build` is raised (returning it would be the same use-after-free class as F2 — no valid value exists for a device that is gone). The reviewer's deterministic probe is promoted into `test_guarded_cache.ml`, plus a companion asserting that `clear` — which does *not* invalidate the device — still installs an in-flight build, so the epoch check is not over-applied.

## F4 — Medium: values escape the lock with no ownership — chose (a), documented

`find_or_build` returns `v` after unlocking, and `clear`/`evict_device` run `destroy` outside the lock, so domain B's `clear_cache` can `clReleaseKernel` / `cuModuleUnload` a handle domain A is mid-`Kernel.launch` with. The `.mli` presented the module as making "the cache" concurrency-safe; it makes the **table** safe, not the **values**, and `test_concurrent_clear` only asserts that nothing raised, so it cannot detect the difference.

Chose **(a): document the caller obligation precisely, no behaviour change.** Reasoning: (b) is not clean here. Refcounting would have to reach every `find_or_build` caller across five backends plus `Runtime` (each would need an explicit release around `launch`, and `Kernel.t` currently exposes no such scope), and deferred destroy needs a quiescence signal the library does not have — a device teardown must release *now*, before `cuCtxDestroy`, so "defer until no reader" can only be implemented as a wait, which would let a stuck reader block teardown. That is a design change deserving its own review, not a passenger on a keying fix. Meanwhile the obligation is in fact satisfied by every in-tree caller (teardown is a program-wide quiescent operation), so (a) states the truth rather than papering over a live bug. The `.mli` now spells out the exact bad interleaving, states the obligation, notes that the tests cannot detect a violation of it, and names refcounting/deferred destroy as a tracked follow-up instead of an implied guarantee.

## F5 — Low: raw lock/unlock in `find_or_build`

An exception between `Mutex.lock` and `Mutex.unlock` (a `Hashtbl.replace` resize under `Out_of_memory`, or `Stack_overflow`) would leave the mutex held forever and deadlock every later cache operation on every domain. Both critical sections are straight-line, so both now use `Mutex.protect`, matching the rest of the module.

## F6 — Low: a raising `destroy` stranded the other victims

OpenCL/Metal `destroy` callbacks *can* raise (`Opencl_api.release` goes through `check`; CUDA/HIP wrap in `ignore` and cannot). `List.iter t.destroy victims` abandoned every remaining victim on the first raise — and they are already unlinked from the table, so that leaked them outright, a behaviour regression from the pre-#275 `clear_cache`, which released inline *before* clearing so a raise left entries still cached and reclaimable. Each `destroy` is now isolated and the first exception re-raised once every victim has been attempted, so the failure is still reported without being fatal to the rest. Separately, on the lost-race path a failing `destroy` of the discarded loser no longer propagates in place of the perfectly good `winner`, removing an intermittent, multi-domain-only compile failure. Both have regression tests that go red when the isolation is removed.

## Does F1 invalidate the benchmark sweep?

**No, and not narrowly — the sweep never reaches the buggy code path.** `docs/benchmarks/hip-vs-opencl-vulkan-2026-07-24.md` runs `bench_*`, which goes `Execute.run_vectors` → `B.run_source` → `B.Kernel.compile_cached`, i.e. straight into the **per-backend** cache whose key already included the device id. `Runtime.compile_kernel` — the F1 site — has no in-tree caller outside tests.

Verified rather than assumed, by re-running the `Verified ✗` rows after the fix:

| benchmark | size | backend | dev 0 (dGPU) | dev 1 (iGPU) |
|---|--:|---|:--:|:--:|
| vector_add | 1 M | HIP | verified ✓ | verified ✓ |
| vector_add | 100 M | HIP | verified ✗ | verified ✗ |
| vector_add | 1 M | OpenCL | verified ✓ | verified ✓ |
| vector_add | 100 M | OpenCL | verified ✗ | verified ✗ |
| nbody | 1024 | HIP | verified ✓ | verified ✓ |
| nbody | 4096 | HIP | verified ✗ | verified ✗ |

The ✗ persists after the fix, is **size-dependent** (✓ at the small size, ✗ at the large one), is identical across backends, and appears on **both** devices — including device 0, which is enumerated and compiled first and therefore could never have aliased. That is exactly the float32-vs-float64 epsilon signature footnote 2 describes. F1 would instead have produced an unwritten buffer at *every* size and only on the *second* device. The `HIP wins 13/21` verdict and the whole dGPU comparison stand; no doc change needed.

## Gates

- `@spoc/framework/test/runtest` — green (Guarded_cache: 8/8, including the 4 new cases).
- `@sarek/tests/unit/runtest` — green.
- `@sarek/tests/codegen_golden/runtest` — green, goldens byte-identical (this is a caching/keying change; no generated code moves).
- New e2e tests `test_runtime_cache_device_key` and `test_runtime_cache_teardown` — both exit 0, both wired into the default `runtest` alias, both skip with a printed reason (exit 0) when the devices they need are not enumerated.
- `@sarek/tests/e2e/runtest` — one failing rule, `test_static_tag_erasure`, **pre-existing**: it fails identically on the unmodified base (exit 1, intermittently 139) on the Vulkan/RADV dGPU leg. Unrelated to this change.
- `@fmt` — clean on every touched file. Pre-existing drift left alone in `sarek/core/Device.ml`, `sarek/interp/Sarek_float32.ml`, `sarek/tests/unit/test_ir_layout.ml`, and also `sarek/codegen/Sarek_ir_{glsl,softmath,ptx_expr}.ml` and `sarek-cuda/test/test_ptx_stride_spike.ml`.
- A whole-repo `dune build` still fails pre-existing on `sarek/transpile/test/test_transpile_proof.bc` (bytecode link, `caml_unix_sigwait` missing in this switch); the aliases above were built individually.

## Deliberately deferred

- **F4 (b)** — refcounting / deferred destroy in `Guarded_cache`, per the reasoning above. The obligation is now documented instead of silently assumed.
- **Vulkan device teardown** — `Vulkan_api_device.destroy` calls `vkDestroyDevice` with no eviction hook at all, so `Vulkan_api_kernel.cache` keeps pipelines from a destroyed device. That is a pre-existing gap in the *backend* layer, wider than #86, and adding only the outer notification would have hidden it behind a half-fix.

### DIFF
--- a/sarek-cuda/Cuda_api.ml
+++ b/sarek-cuda/Cuda_api.ml
@@ -277,11 +277,24 @@ module Device = struct
        recreated. *)
     Hashtbl.remove device_cache dev.id ;
     (* Notify layers above this backend BEFORE the local hooks unload modules:
-       they memoize values that close over the handles those hooks release. *)
-    Spoc_framework.Cache_hooks.notify_device_destroy dev.id ;
+       they memoize values that close over the handles those hooks release.
+       [notify_device_destroy] re-raises the first failing listener, and
+       [device_cache] has already been emptied, so letting it escape here would
+       leave the device unreachable with its context still alive and its modules
+       still loaded. Capture it, finish the teardown, re-raise at the end. A
+       failure in the teardown itself (cuCtxDestroy) legitimately wins over a
+       memoization-drop failure. *)
+    let listener_exn =
+      match
+        Spoc_framework.Cache_hooks.notify_device_destroy ~backend:"CUDA" dev.id
+      with
+      | () -> None
+      | exception e -> Some e
+    in
     List.iter (fun hook -> hook dev.id) !device_destroy_hooks ;
     retire_device dev.id ;
-    check "cuCtxDestroy" (cuCtxDestroy dev.context)
+    check "cuCtxDestroy" (cuCtxDestroy dev.context) ;
+    Option.iter raise listener_exn
 end
 
 (** {1 Memory Management} *)
--- a/sarek-cuda/Cuda_api.ml
+++ b/sarek-cuda/Cuda_api.ml
@@ -276,6 +276,9 @@ module Device = struct
        can't be returned by [Kernel.compile_cached] after the context is
        recreated. *)
     Hashtbl.remove device_cache dev.id ;
+    (* Notify layers above this backend BEFORE the local hooks unload modules:
+       they memoize values that close over the handles those hooks release. *)
+    Spoc_framework.Cache_hooks.notify_device_destroy dev.id ;
     List.iter (fun hook -> hook dev.id) !device_destroy_hooks ;
     retire_device dev.id ;
     check "cuCtxDestroy" (cuCtxDestroy dev.context)
--- a/sarek-hip/Hip_api.ml
+++ b/sarek-hip/Hip_api.ml
@@ -216,10 +216,21 @@ module Device = struct
        destroy under the runtime model, so there is no hipCtxDestroy analog. *)
     Hashtbl.remove device_cache dev.id ;
     (* Notify layers above this backend BEFORE the local hooks unload modules:
-       they memoize values that close over the handles those hooks release. *)
-    Spoc_framework.Cache_hooks.notify_device_destroy dev.id ;
+       they memoize values that close over the handles those hooks release.
+       [notify_device_destroy] re-raises the first failing listener; since
+       [device_cache] has already been emptied, letting it escape here would
+       leave the device unreachable with its modules still loaded. Capture it,
+       finish the teardown, re-raise at the end (mirrors the CUDA backend). *)
+    let listener_exn =
+      match
+        Spoc_framework.Cache_hooks.notify_device_destroy ~backend:"HIP" dev.id
+      with
+      | () -> None
+      | exception e -> Some e
+    in
     List.iter (fun hook -> hook dev.id) !device_destroy_hooks ;
-    retire_device dev.id
+    retire_device dev.id ;
+    Option.iter raise listener_exn
 end
 
 (** {1 Memory Management} *)
--- a/sarek-hip/Hip_api.ml
+++ b/sarek-hip/Hip_api.ml
@@ -215,6 +215,9 @@ module Device = struct
        selected, mirroring the CUDA backend. HIP has no explicit context to
        destroy under the runtime model, so there is no hipCtxDestroy analog. *)
     Hashtbl.remove device_cache dev.id ;
+    (* Notify layers above this backend BEFORE the local hooks unload modules:
+       they memoize values that close over the handles those hooks release. *)
+    Spoc_framework.Cache_hooks.notify_device_destroy dev.id ;
     List.iter (fun hook -> hook dev.id) !device_destroy_hooks ;
     retire_device dev.id
 end
--- a/sarek/core/Device.ml
+++ b/sarek/core/Device.ml
@@ -144,8 +144,26 @@ let best () =
       let opencl = by_framework "OpenCL" in
       if Array.length opencl > 0 then Some opencl.(0) else first ()
 
-(** Reset initialization state (for testing) *)
+(** Reset initialization state (for testing).
+
+    Retires the global id space, so every cache keyed on it must be dropped in
+    the same breath. [init] restarts [global_id] at 0 over whichever frameworks
+    it is handed, so the same id routinely denotes a DIFFERENT physical device
+    after a reset: with [~frameworks:["OpenCL"]] then
+    [~frameworks:["Native"; "OpenCL"]], the single Native device shifts OpenCL
+    down one and id 1 moves from the second OpenCL device to the first. A
+    surviving entry keyed on id 1 is then served to the wrong device, silently —
+    a compiled kernel carries no device check — which is exactly the aliasing
+    class this cache was keyed by device to prevent.
+
+    Only [Sarek.Runtime]'s outer memo is affected: it keys on the global
+    [Device.t.id], while every per-backend cache keys on its own backend-local
+    device index, which a reset does not perturb. So the notification here is
+    enough, and it releases nothing — the listeners drop memoization only (see
+    {!Spoc_framework.Cache_hooks}), so unlike [Kernel.clear_cache] this cannot
+    invalidate a handle a caller still holds. *)
 let reset () =
+  Spoc_framework.Cache_hooks.notify_clear_all () ;
   devices := [||] ;
   initialized := false
 
--- a/sarek/core/Kernel.ml
+++ b/sarek/core/Kernel.ml
@@ -160,12 +160,39 @@ let launch (Kernel (module K)) ~(args : args) ~(grid : Framework_sig.dims)
     the backend handles (clReleaseKernel / cuModuleUnload / ...), and any outer
     memo — notably [Runtime.kernel_cache] — holds [Kernel.t] closures over
     exactly those handles. Leaving them in place means the next lookup hits the
-    outer memo and launches through a released handle (segfault). *)
+    outer memo and launches through a released handle (segfault).
+
+    The notification is fired on BOTH sides of the backend clear, and that is
+    load-bearing rather than belt-and-braces. The pre-notification drops what is
+    already memoized; it cannot cover a build that starts after it. An outer
+    build that misses after the first notification snapshots the {e new}
+    generation, compiles a handle that [B.Kernel.clear_cache] then releases, and
+    on re-acquire finds the generation unchanged — so it installs a value over a
+    dead handle and poisons the memo for the rest of the process, which is the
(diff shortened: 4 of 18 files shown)


mathiasbourgoin:
## Review follow-ups pushed — `2dd9f1f7`

**C1 (Critical) — accepted; it was a scoping error in my reasoning, not a slip.** My comment on `clear` justified skipping the epoch with *"the device is still alive, so the value being built is still valid"*. That holds for a cache that **owns** its handles and fails for one whose values **borrow** another cache's — which is precisely `Runtime.kernel_cache`, the cache I had just registered as the `on_clear_all` listener, and for which `Kernel.clear_cache` is the announcement that the borrowed handles are about to die. I generalised an owning-cache invariant onto a borrowing cache in the same commit that made it a borrowing cache.

Fixed as designed: a cache-wide generation bumped by `clear`, re-checked on re-acquire, gated by a create-time `?invalidated_by_clear` flag. `Runtime.kernel_cache` sets it; the five backend caches do not. Raises a new `Cache_cleared_during_build` (distinct from `Device_destroyed_during_build` — no device died) and `destroy`s the value.

Red→green, both directions, which is what convinced me the flag is load-bearing rather than decorative:
- check disabled → `clear during an in-flight build is rejected on a borrowing cache` **FAILS** (the value is installed and a later lookup is served the closure over the released handle).
- check forced on unconditionally → `clear during an in-flight build still installs on an owning cache` **and** the pre-existing `concurrent clear vs find_or_build` stress test both **FAIL**.
- as shipped → all 9 Guarded_cache cases green.

**C2 (Important) — accepted, and you are right that my disclosed trade-off was the actual defect.** "Over-eviction of a pure memo is free" was true *before* this PR and stopped being true *inside* it, the moment `evict_device` gained the epoch bump. I carried the old justification forward without re-deriving it against the new semantics.

Took **(a)**: `notify_device_destroy` now carries `~backend` alongside the index, and the listener matches on both. One wrinkle worth recording — the label has to be the **family** name, not `B.name`: `Cuda_api` backs two registered backends (`CUDA/PTX`, `CUDA/C`) and cannot name a single one, so the backends pass `"CUDA"` / `"HIP"` and the listener matches `d.framework = backend || d.framework` starting with `backend ^ "/"`, mirroring `Device.resolve_framework`. This is documented in `Cache_hooks.mli`. I did not take (b) because a memo-drop-only `drop_device` would leave the wrong default in place for the next listener.

Regression needs no race — it is cache identity, deterministic: a `~backend:"HIP"` notification with a colliding index must leave the OpenCL entry *physically identical* (`==`) in the cache, and the OpenCL notification must evict it. Reverting to index-only matching gives `FAIL: a HIP device-0 teardown evicted the OpenCL device-0 entry`.

**C3 (Important) — accepted.** Both `Cuda_api.Device.destroy` and `Hip_api.Device.destroy` now capture the listener exception, complete the *whole* teardown (local hooks → `retire_device` → `cuCtxDestroy`), and re-raise at the end; a failure in the teardown itself legitimately wins over a memoization-drop failure. I chose re-raise-after rather than swallow so an extension-point bug stays visible. `Cache_hooks.mli` gains a **"Notifying from a teardown path"** section stating the obligation and naming those two call sites as the shape to copy. New `test_cache_hooks.ml` covers the contract: the backend family + index reach the listener, and one failing listener does not skip the others.

**Smaller items — both taken.** `Runtime.compile_kernel` and `Runtime.run` now carry `@raise` notes for both exceptions. The listener's silent no-op for a device absent from `!Device.devices` is documented in place as a known residual corner that the C2 fix does not close.

**Vulkan — not expanded, per your instruction**, and thank you for re-scoping it: I had disclosed only the missing notification, and you are right that `Vulkan_api_kernel` not passing `~device_id` means per-device eviction is not even expressible there. Two layers, both pre-existing, both untouched here.

### Gates (re-run in full on `2dd9f1f7`)
- `@spoc/framework/test/runtest` — green. Guarded_cache **9/9** (was 8, +1 borrowing case), Cache_hooks **2/2** (new).
- `@sarek/tests/unit/runtest` — green. `@sarek/tests/codegen_golden/runtest` — green, goldens byte-identical (no golden file appears in `git status`).
- `test_runtime_cache_device_key` (now 4 checks incl. the C2 scoping pair) and `test_runtime_cache_teardown` — both exit 0.
- `@sarek/tests/e2e/runtest` — still exactly one failing rule, `test_static_tag_erasure`, confirmed pre-existing on the unmodified base earlier in this PR.
- `@fmt` — clean on every touched file; the same pre-existing drift left alone.

### Red-on-mutation re-verified after the changes
| mutation | result |
|---|---|
| device component dropped from the key (F1) | `requested device 1, got a kernel bound to device 0` + `16/16 elements were not written` |
| `notify_clear_all` removed (F2) | **rc=139**, core dumped |
| generation check disabled (C1) | `[FAIL] clear during an in-flight build is rejected on a borrowing cache` |
| generation check forced on for all caches (C1, inverse) | `[FAIL]` on the owning-cache case **and** on `concurrent clear vs find_or_build` |
| listener reverted to index-only matching (C2) | `FAIL: a HIP device-0 teardown evicted the OpenCL device-0 entry` |

Not merged.

PR #302, opened 2026-07-25
fix(spoc,sarek): close the three cache-invalidation holes #293 left open (#119, #90)

mathiasbourgoin:
Closes #119. Also closes #90.

PR #293 fixed the device-blind `Runtime.compile_kernel` cache and its missing teardown invalidation, but left three holes open. This closes all three.

## The three holes, as they were

**H1 — a direct backend clear bypasses the notification entirely.**
Every backend's `Kernel.clear_cache` was a bare `Guarded_cache.clear`:
`sarek-opencl/Opencl_plugin_base.ml:255`, `sarek-metal/Metal_plugin_base.ml:262`, `sarek-vulkan/Vulkan_api_kernel.ml:414`, `sarek-cuda/Cuda_api.ml:663`, `sarek-hip/Hip_api.ml:477`. `Cache_hooks.notify_clear_all` was fired only by the *wrapper* at `sarek/core/Kernel.ml:184`. `Framework_registry.find_backend` is public, so `B.Kernel.clear_cache ()` released every `cl_kernel`/`cl_program`/`VkPipeline`/`CUmodule` with `Runtime.kernel_cache` still holding `Kernel.t` closures over them — the same use-after-free #293 fixed, through a sibling entry point. `spoc/framework/Cache_hooks.mli:33-40` documented it as a known hole.

**H2 — the eviction scope was mixed.**
`~device_id` was passed to `Guarded_cache.find_or_build` by CUDA (`sarek-cuda/Cuda_api.ml:649`) and HIP (`sarek-hip/Hip_api.ml:471`) only — not by OpenCL (`sarek-opencl/Opencl_plugin_base.ml:252`), Metal (`sarek-metal/Metal_plugin_base.ml:259`) or Vulkan (`sarek-vulkan/Vulkan_api_kernel.ml:409`). Since `keys_by_device` is populated *only* from that argument, `Guarded_cache.evict_device` could not reach an OpenCL, Metal or Vulkan entry at all: those three caches were per-device in their *keys* but global in their *eviction*. Recorded at `Cache_hooks.mli:41-46`.

**H3 — both e2e tests were vacuous on CI.**
`sarek/tests/e2e/test_runtime_cache_device_key.ml:100-107` and `test_runtime_cache_reset_ids.ml:80-87` / `:126-141` (and `test_runtime_cache_teardown.ml:61-66`) printed `SKIP:` and `exit 0`. Under `(rule (alias runtest) (action (run ...)))` at `sarek/tests/e2e/dune:1196-1209` that is byte-for-byte indistinguishable from a run that verified everything. CI has no GPU (and pocl was dropped in 859c8601), so all three skipped and the alias went green having checked nothing.

## The scope decision — per-device, everywhere

Justified from how the backends actually key their compiled artifacts: every one of them folds the backend-local device index into `Compile_cache.make_key` (`Cuda_api.ml:641`, `Hip_api.ml:462`, `Opencl_plugin_base.ml:245`, `Metal_plugin_base.ml:253`, `Vulkan_api_kernel.ml:404`). A compiled artifact is per-device *by construction*, so eviction granularity should match keying granularity rather than being coarser. Going global instead would mean a single device teardown throws away every other device's compiled kernels — and, since `evict_device` aborts in-flight builds, would make one device's teardown raise inside another device's healthy compile.

So all five backends now pass `~device_id`, and all five register the **same** `Cache_hooks.on_device_destroy` listener matching on their own family name. CUDA's and HIP's private `device_destroy_hooks` lists are deleted in favour of it — one mechanism, not two. OpenCL and Metal expose no device-destroy entry point at all, so nothing fires for them today; their listeners are wired but inert, and the `.mli` says so rather than leaving it implicit.

## #90 is closed, not deferred

The scope decision made it tractable. With `~device_id` passed, per-device eviction is now expressible in `Vulkan_api_kernel`; and `Vulkan_api_device.destroy` now fires `notify_device_destroy` *before* releasing anything, with the capture-finish-re-raise shape `Cache_hooks.mli` requires. Ordering is the whole property: every object the Vulkan cache's `destroy` releases (pipeline, pipeline layout, descriptor pool, descriptor set layout, shader module) belongs to the VkDevice being torn down. Both layers are covered by executed tests.

## Red-proof

Device: **AMD RX 7900 XTX** — Mesa `radeonsi` OpenCL (2 devices: the dGPU and the 7950X iGPU), **RADV** Vulkan, DRM 3.64, kernel 7.1.2-3-cachyos. **No NVIDIA GPU**, so CUDA is compile-only here.

Each mechanism was mutated and the covering test required to fail:

| mutation | test | result |
|---|---|---|
| `Runtime`'s `on_clear_all` listener unregistered | teardown + direct-clear | FAIL — *"Kernel.clear_cache released the backend handles but the outer memo still holds the Kernel.t that closes over them… notify_clear_all did not reach Runtime's on_clear_all listener"* |
| outer memo key drops the device | device_key | FAIL — *"requested device 1, got a kernel bound to device 0 (outer cache key omits device identity)"* |
| `Device.reset` stops notifying | reset_ids | FAIL — *"16/16 elements were not written - the outer memo served id 1's pre-reset entry, which is bound to a different physical device"* |
| `around_clear` stops notifying | teardown + direct-clear | FAIL — *"a direct B.Kernel.clear_cache () released the backend handles without invalidating the outer memo - the notification is in the Sarek.Kernel.clear_cache wrapper only…"* |
| `Guarded_cache.clear` evicts nothing | teardown | FAIL (same named assertion) |
| OpenCL drops `~device_id` | backend_cache_scope | FAIL — *"OpenCL: the device-0 teardown did NOT evict the backend cache entry. Guarded_cache.evict_device can only reach entries that were installed with ~device_id…"* |
| Vulkan drops `~device_id` | vulkan_cache_device_scope | FAIL — *"…so the pipeline, layouts, descriptor pool and shader module for a destroyed VkDevice survive it (#90)"* |
| `Vulkan_api_device.destroy` stops notifying | vulkan_cache_device_scope | FAIL — *"Vulkan_api_device.destroy left the kernel cache entry in place: the recreated device 0 was handed a pipeline… belonging to the VkDevice that was just destroyed"* |

The two crash-shaped cases previously had SIGSEGV as their *only* signal, which names nothing. Both now assert on the physical identity of what `Runtime.compile_kernel` returns *before* the crash-prone launch, so they report the problem in words and still keep the launch (identity alone cannot prove the handles were really released).

**Skip is now visible.** Simulating a device-less CI box (`OCL_ICD_VENDORS` at an empty directory, Vulkan ICD hidden):

```
Testing `Runtime cache device key'.
  [SKIP]        outer memo          0   is keyed by device identity and dropp...
Test Successful in 0.000s. 0 test run.
```

— against the previous `SKIP: …` + `exit 0`, which dune reported as a plain green.

## Not verified on this hardware

- **CUDA and HIP.** Compile-only (no NVIDIA GPU; the HIP arm was not exercised at runtime). Their `~device_id` wiring predates this PR; what is new for them is that eviction moved from a private hook list onto `Cache_hooks`. Reviewed, compiled, not executed.
- **Metal.** Compiles on Linux; no Apple hardware here.
- The CUDA/HIP/Metal listener registrations are code-review-only.

## Pre-existing issue found, deliberately not fixed here

`sarek/tests/e2e/test_static_tag_erasure.exe` **SIGSEGVs on origin/main** (cf58801b, unmodified, standalone) inside `libvulkan_radeon` during `vkCreateComputePipelines`, and passes with `OCL_ICD_VENDORS` pointed at an empty directory. It is a Mesa-level conflict: creating a Vulkan compute pipeline in a process that already enumerated radeonsi OpenCL crashes. It is the only failing rule in `@sarek/tests/e2e/runtest` both before and after this PR. It is why the Vulkan scope test is its own executable that links `sarek_vulkan` directly and never initialises OpenCL.

Separately, `Vulkan_types.ml:910` declares `pName` with ctypes' `string` view; `setf` stores a pointer to a ctypes-managed buffer that becomes unreachable as soon as `setf` returns, so `vkCreateComputePipelines` reads the entry-point name out of memory whose finaliser may already have run. I could not produce a red/green for it — the Mesa crash above masks it — so I have not shipped a speculative fix. Worth its own issue.

### DIFF
--- a/sarek-cuda/Cuda_api.ml
+++ b/sarek-cuda/Cuda_api.ml
@@ -45,12 +45,6 @@ let check (ctx : string) (result : cu_result) : unit =
     ctx
     result
 
-(** Hooks invoked with a device id right before its context is destroyed.
-    [Kernel] registers an eviction hook here (after [Kernel.cache] is defined
-    below) so that [Device.destroy] can retire per-device compiled kernels
-    without a circular module dependency. *)
-let device_destroy_hooks : (int -> unit) list ref = ref []
-
 (** Host-side kernel-argument buffers (the [CArray] of argument pointers plus
     the per-argument value cells) for launches that may still be in flight,
     keyed by device id.
@@ -270,14 +264,14 @@ module Device = struct
   let destroy dev =
     (* Evict from device_cache first: leaving a stale entry means a later
        [get idx] returns a handle whose context has already been destroyed
-       (mirrors the Vulkan fix in Vulkan_api_device.ml). Also run the
-       registered destroy hooks (Kernel.cache eviction) while the context
-       is still current, so stale module/function handles for this device
-       can't be returned by [Kernel.compile_cached] after the context is
-       recreated. *)
+       (mirrors Vulkan_api_device.destroy). *)
     Hashtbl.remove device_cache dev.id ;
-    (* Notify layers above this backend BEFORE the local hooks unload modules:
-       they memoize values that close over the handles those hooks release.
+    (* Notify BEFORE anything is released, while the context is still current:
+       the listeners are the cache layers — this backend's own [Kernel.cache]
+       (which unloads this device's modules) and, above it, memos closing over
+       exactly those handles. Doing it here is what stops a stale
+       module/function handle being returned by [Kernel.compile_cached] after
+       the context is recreated under the same index.
        [notify_device_destroy] re-raises the first failing listener, and
        [device_cache] has already been emptied, so letting it escape here would
        leave the device unreachable with its context still alive and its modules
@@ -291,7 +285,6 @@ module Device = struct
       | () -> None
       | exception e -> Some e
     in
-    List.iter (fun hook -> hook dev.id) !device_destroy_hooks ;
     retire_device dev.id ;
     check "cuCtxDestroy" (cuCtxDestroy dev.context) ;
     Option.iter raise listener_exn
@@ -502,13 +495,18 @@ module Kernel = struct
       ~destroy:(fun k -> ignore (cuModuleUnload k.module_))
       ()
 
-  (* Evict every cached kernel compiled for [device_id]. Registered as a
-     [device_destroy_hooks] callback below so [Device.destroy] retires
-     these handles before the underlying CUDA context is destroyed. *)
-  let evict_device device_id =
-    Spoc_framework.Guarded_cache.evict_device cache device_id
-
-  let () = device_destroy_hooks := evict_device :: !device_destroy_hooks
+  (* Evict every cached kernel compiled for [device_id]. Registered on the
+     shared [Cache_hooks] registry — the one mechanism every backend uses (see
+     Cache_hooks.mli) rather than a CUDA-private hook list — so that
+     [Device.destroy], which fires the notification before it destroys
+     anything, retires these module handles while the context is still alive.
+     Match on the family name, never on the index alone: backend-local indices
+     collide across backends, and [evict_device] does not merely drop
+     memoization, it aborts in-flight builds for that index. *)
+  let () =
+    Spoc_framework.Cache_hooks.on_device_destroy (fun ~backend index ->
+        if String.equal backend "CUDA" then
+          Spoc_framework.Guarded_cache.evict_device cache index)
 
   (* Replace the .target directive in a PTX string to match the given SM version.
      This makes a static PTX string portable: PTX written for sm_86 loads fine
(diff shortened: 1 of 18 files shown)


mathiasbourgoin:
@coderabbitai review

mathiasbourgoin:
@coderabbitai full review

mathiasbourgoin:
@coderabbitai review

mathiasbourgoin:
@coderabbitai review

Issue #101, opened 2026-01-12
Fix radix_sort benchmark segmentation fault


mathiasbourgoin:
## Description

The `bench_radix_sort` benchmark is segfaulting during execution, preventing it from completing and generating benchmark data.

## Error

```
▶ Radix Sort...
./benchmarks/run_all_benchmarks.sh : ligne 176 : 1978883 Erreur de segmentation (core dumped)
dune exec benchmarks/bench_radix_sort.exe -- --output "${RUN_DIR}"
make: *** [Makefile:389: bench-all] Error 139
```

## Impact

- Radix sort is one of the 20 benchmarks in the web viewer
- Currently shows "No data available" message
- Users cannot compare radix sort performance across systems
- All other 19 benchmarks work correctly

## Context

- Other sort benchmark (bitonic_sort) works fine
- Radix sort has description file and is configured in web viewer
- Issue discovered while running all benchmarks for PR #100
- Temporary workaround: error handling added to run script to continue past failure

## Steps to Reproduce

```bash
eval $(opam env)
dune exec benchmarks/bench_radix_sort.exe
```

## Expected Behavior

Benchmark should complete successfully and generate JSON output files like other benchmarks.

## Suggested Investigation

1. Check for out-of-bounds array accesses
2. Review shared memory usage and limits
3. Test with smaller array sizes
4. Compare implementation with working bitonic_sort benchmark
5. Run with debugger (gdb) to identify crash location

## Related

- PR #100: Benchmark visualization pages (includes radix_sort in web viewer)
- Part of Phase 2 implementation priority (sorting benchmarks)

PR #305, opened 2026-07-25
fix(harness): track the review-tool bundle, and make four silent harness failures loud (#108/#101/#102/#103)

mathiasbourgoin:
Backlog #108, #101, #102, #103 — the agent/review harness itself.

Every guard below was proven by constructing the bad input and capturing the refusal. A guard nobody has watched trigger is not a guard, so each negative case is paired with a positive control.

## #108 — the review-tool bundle was invisible to git

All 23 files of the review-tool bundle were untracked and hidden behind `.git/info/exclude`, which is machine-local and never shared. `git status` was clean, `.gitignore` said nothing, and **a fresh clone got no review gates at all**. `review-bundle-verify.js` could not catch it: it checks the bundle is complete and unmodified, which an untracked bundle satisfies perfectly on the machine that has it.

**Tracked (25 paths):**

| What | Why |
|---|---|
| `scripts/review-bundle.manifest.json` | the integrity sentinel; useless if it does not ship |
| `scripts/{review-normalize,check-review-convergence,review-bundle-verify,xruntime-review}.js`, `scripts/{check-scope-diff,xruntime-exec}.sh`, `scripts/REVIEW-BUNDLE.md` | the gates themselves |
| `scripts/lib/review/*.js` (8), `scripts/lib/xruntime/*.js` (3+1 new) | their shared libraries |
| `schema/review-{finding,trace}.schema.json` | the schemas the gates validate against; `schema/review-json-schema.md` was already tracked and referenced them |
| `tools/data-schema/fixtures/review-finding/{valid,invalid}/*.jsonl` | the fixtures that prove the schema validator works |
| `scripts/check-review-bundle-tracked.sh` | new guard (below) |

**Left untracked, moved from `.git/info/exclude` into `.gitignore` so the decision is visible and reviewable:** `briefs/` (per-PR scaffolding), `skills-meta/` (per-workstation health reports), `kb/`, `.harness/`, `.agents/` (roster installs regenerated by `/recruit`), and the roster runtime dirs `roster/`, `skills/`, `workflows/`, `dist/`, `.codex/`, plus `specs/main-master-reconcile.md`. These carry machine-local state and are not gates. `.gitignore` now says so explicitly, including why the bundle is the opposite case.

**New guard** `scripts/check-review-bundle-tracked.sh` fails when any bundle path is untracked *or* matched by any ignore source, printing the offending rule and its source file, then delegates content integrity to `review-bundle-verify.js`. Wired into CI, which runs on a fresh checkout — the exact case that was broken.

**What untracked actually cost, concretely.** While this PR was open, `scripts/check-review-convergence.js` was overwritten and there was **no pristine copy anywhere on disk** to restore from. Recovery meant reconstructing the file by reversing eight edits and checking the result byte-for-byte against the manifest sha. Once the bundle is tracked, git *is* the pristine copy and that recovery is `git checkout -- scripts/check-review-convergence.js`.

> **Fired:** run before staging, it reported all 23 paths as `UNTRACKED` + `IGNORED` with the `.git/info/exclude` line number for each, while `review-bundle-verify` reported OK on the same tree.
> **Accepted:** a `--no-local` clone of the branch runs every gate green — bundle verify, Alcotest registration, and all four test suites (83 assertions).

## #101 — worktree isolation landing on the wrong repo

`scripts/agent-worktree-bootstrap.sh`. Five failure modes, five refusals:

1. **Wrong repo.** When the project is an untracked directory inside another repo, the enclosing repo is the parent and the agent gets a worktree containing none of the project's sources. Refused. A *tracked* subdirectory still proceeds, with a warning.
2. **Stale base.** Compares local `origin/<branch>` against `ls-remote`. It deliberately does **not** auto-fetch: a silent fetch mutates a checkout other agents share, and would hide the very condition being checked. `--allow-stale-base` is the explicit override and records itself.
3. **Missing `briefs/`.** Git-ignored, therefore absent from a fresh worktree, yet it holds the ledger, control files and the xruntime journal. Symlinked, not copied, so there is one ledger.
4. **Wrong dune root.** Refuses a worktree location under an ancestor holding `dune-project`/`dune-workspace`, and pins the root in the worktree it creates.
5. **Shared scratchpad.** Each dispatch gets `.agent-scratch/<agent>-<utc>-<pid>`. Two agents wrote to the same generic `pr.md` and one published the other's text as a PR description; a bare filename can no longer alias. An existing worktree path is refused rather than reused, so a re-dispatch cannot land on another agent's live checkout — the incident where uncommitted work was destroyed.

> **Fired:** 17 assertions in `agent-worktree-bootstrap.test.sh`, including a positive control that a freshly fetched base is accepted so the stale refusal is not vacuous. Confirmed against the real layout — `--project /home/mathias/dev/interviews` refuses and names `/home/mathias/dev` as the repo it would otherwise have branched.

## #102 — a malformed output armed the runtime breaker

The breaker treated every degraded outcome as a broken runtime. A non-conforming **output** is not that: the runtime started, ran, and answered — the answer failed a contract the caller never stated. Four parallel cross-runtime passes tripped `opencode` on all four PRs at once, and the breaker read self-inflicted contention as runtime failure, suppressing probes for unrelated work.

- `xruntime-classify.js` labels every non-healthy outcome `runtime` or `caller`. `spawn-error` / `timeout` / `tree-mutation` / `empty-output` stay `runtime`; only `non-conforming-output` is `caller`.
- `shouldRefuseDegraded` ignores caller-fault entries from **both** enforcement inputs. Entries written before this field existed have no `fault` key and are still read as runtime faults, so nothing already on disk is silently un-armed.
- the result carries `fault` plus a `remedy` line — bare "degraded" is what made operators blame the runtime.
- **`--emit-contract`** prints the machine-parsed output contract so callers stop paraphrasing it per prompt. A contract nobody can obtain mechanically is a caller fault nobody can fix.

The manifest **declares** the divergence rather than hiding it: `channel: local-patched`, version `1.6.0+spoc.1`, and a `local_patches[]` entry naming the paths, reason, covering test, and `reapply_on_upgrade`. Verify still passes, so the manifest remains an honest sentinel against *undeclared* drift.

> **Fired:** 13 assertions, each caller-fault case paired with a runtime-fault positive control, including end-to-end through the real CLI — a stub emitting garbage must not suppress the second probe, a stub emitting nothing must. **Mutation-checked:** stubbing `isCallerFault` to return `false` turns 3 of them red, including the E2E case.

## #103 — roster-implement post-hook

`scripts/roster-implement-posthook.sh`:

- **Ledger schema.** Upstream's rules are a jq predicate embedded in an untracked skill file, so a fresh clone has no ledger schema at all. `scripts/lib/ledger-schema.js` is the tracked, jq-free authority; when jq and the skill file are both present the hook runs **both** and fails on disagreement. Two copies of a rule that can drift apart unnoticed are worse than one.
- **Worktree `briefs/` consolidation.** A conflict (both sides changed the same file differently) is reported and neither side overwritten.
- **Alcotest registration.** `scripts/check-alcotest-registration.js` flags any top-level `let test_*` in an Alcotest file that nothing else references. It strips comments and string literals first, so a commented-out registration does not count as one, and it reports how many files and suites it actually scanned so "OK" over zero suites cannot pass for a real result.

CI was running none of the harness's own tests. It now runs all four suites plus the registration guard.

> **Fired:** 23 assertions — 8 malformed ledgers each with their own message, a jq/schema drift pair, a consolidation conflict leaving both copies intact, and 4 Alcotest orphan shapes. On the real tree the registration check scans 541 `.ml` files across 59 suites and finds no orphans; the hook validates a real ledger with jq agreeing.

## Review round (adversarial + CodeRabbit)

`CHANGES_REQUESTED` with 11 actionable comments and 4 nitpicks, plus adversarial findings F1/F2/F3/F10. Disposition for every one is in commit `fc50d154`; the short version:

**F1 was a live regression this PR introduced and is the most important fix here.** `#102` added fault attribution, but `classifyExitCode` corroborates only exit 3 and exit 124 — so exit 1/127/137/139 fell through to output inspection, which never reads the exit code. Since `xruntime-exec.sh` merges stderr into stdout, a *crashed* runtime was classified `fault=caller` and never armed the breaker. A runtime dying on every invocation would have been re-probed forever. The coverage hole that hid it is closed too: every previous assertion used `exitCode: 0` and both E2E stubs exited 0.

Also fixed: the guard could be defeated by editing the manifest it derives its worklist from (F2); reachability of the tools from CI is now measured rather than assumed (F3); the orphan/`#307` boundary is pinned by test (F10); and the journal fail-close poison, where one corrupt line permanently blocked every future lookup — including any new digest, i.e. the runtime-upgrade escape hatch out of an armed breaker.

**`local_patches[]` is now enforced rather than documentation.** Each entry names markers that exist only while the patch is applied; a test demonstrates `review-bundle-verify` going green on a tree where a simulated upgrade silently reverted the patch, with the marker as the only thing that notices.

**Five findings skipped, with reasons** — all pre-existing upstream-owned code that this PR only made visible by tracking it, each adding permanent re-apply debt for an issue #305 did not introduce: `pending_check` vs the finding schema (**verified not live** — validation runs on inputs, before the key is added), `reviewed_sha`/`fix_sha` value validation, cross-runtime trace coverage for `skipped-degraded`/`blocked` (contradicts the documented design), trace-entry validation before append, and pinning an immutable installer revision. These belong upstream in the roster repo.

Suites: 18 + 35 + 21 + 23 + 30 assertions, all green from a fresh `--no-local` clone. The new bundle-guard suite runs in CI.

## Note on how this was built

Developed in a dedicated worktree, never by checking out a branch in the shared `~/dev/SPOC` checkout — which is precisely the #101 defect being fixed here.

## Summary by CodeRabbit

- **New Features**
  - Added automated review normalization, convergence checks, trace tracking, and cross-runtime verification.
  - Added tooling to create isolated agent workspaces with shared review artifacts and safeguards.
  - Added integrity verification for the review-tool bundle and validation for test registration and review scope.
  - Added standardized schemas for review findings, traces, and durable pipeline state.

- **Documentation**
  - Added guidance for bundle maintenance, local patches, trace records, and cross-runtime fault handling.

- **Tests**
  - Added comprehensive coverage for workspace setup, review validation, bundle integrity, and runtime fault classification.


mathiasbourgoin:
Worked through all 15 review items (11 actionable + 4 nitpicks) plus adversarial findings F1/F2/F3/F10. Fixes in `fc50d154`. Nothing dropped silently — the five skips have stated reasons below.

## Fixed

| # | File | Finding | What changed |
|---|---|---|---|
| F1 | `xruntime-classify.js` | *(adversarial)* nonzero exits fell through to output inspection, so a crashed runtime was classified `fault=caller` and never armed the breaker | Exit status checked before output shape; any nonzero exit is a runtime fault. Not promoted to a specific cause, so D-3 survives: a bare exit 3 still may not assert `tree-mutation`. **Coverage hole fixed too** — every prior assertion used `exitCode: 0` and both E2E stubs exited 0; there are now 6 nonzero-exit cases, both uncorroborated-exit cases, and a crashing E2E stub. |
| F2 | `check-review-bundle-tracked.sh` | *(adversarial)* guard derived its worklist from the manifest it polices | Expectations now come from two independent sources: tracked files under wholly bundle-owned directories, and an anchor list in the guard itself. A test asserts `review-bundle-verify` is blind to the same tampering. |
| F3 | `check-review-bundle-tracked.sh` | *(adversarial)* tools survive a fresh clone, callers do not | Decision stated in `REVIEW-BUNDLE.md`: callers stay local. The gap is now *measured* — transitive reachability from CI/Makefile roots, naming every tool nothing tracked invokes. First cut counted `.md` mentions as callers and reported nearly everything reached; fixed, with a test proving the verdict changes when a tool is genuinely wired in. |
| F10 | *(boundary)* | *(adversarial)* five orphaned files | They belong to the #98/#104 mandated-steps task, not this PR — tracking them here would steal #307's content. The guard reports the orphan class instead, and a test pins the boundary so this bundle never claims a sibling PR's `scripts/lib/*.js`. |
| N2 | `xruntime-journal.js` | one old corrupt line fail-closes every later lookup forever | Corruption fails closed only while it could belong to the current cycle — the only window `shouldRefuseDegraded` reads. Unknown cycle keeps the strict old behaviour. This mattered more than it looks: a *new* digest never matches, so the scan always ran past the corruption, permanently disabling the runtime-upgrade escape hatch out of an armed breaker. |
| N1 | `xruntime-classify.js` | `faultFor()` duplicated inline | Derived through `faultFor()` everywhere. |
| N3 | `review-lifecycle.js` | `deriveRoundState` treats any non-GO status as a NO-GO continuation | Taken. Verified against the 30-assertion assemble suite first. The old fallback was not merely a guess — it increments the round and carries `rounds_audit`/`cross_runtime` forward out of a verdict whose validity is unknown, manufacturing attested-looking state from unverifiable input. |
| N4 | `check-alcotest-registration.js` | every `.ml` read and scanned twice | Single pass; `analyze()` returns `{isSuite, orphans}`. |
| 3651148676 | `xruntime-digest.js` | spawn-launch failures hashed as valid output | A missing binary produced a *stable, real-looking* digest of the empty string. Now a placeholder, and reported as `version-probe-spawn-error:<code>` rather than mislabelled a timeout. |
| 3651148652 | `agent-worktree-bootstrap.sh` | `ls-remote` unbounded and fails open | 15s timeout; failure refuses. Unverified is not the same as fresh — that was the same fail-open the check exists to prevent, relocated from a missing fetch to a missing network. |
| 3651148654 | `agent-worktree-bootstrap.sh` | unescaped `eval`-sourceable output | All emitted values `printf '%q'`-quoted; test constructs a path containing `$(touch ...)` and asserts nothing executes. |
| 3651148656 | `check-alcotest-registration.js` | `readdirSync` unguarded | Guarded like the `statSync` above it. |
| 3651148688 | `roster-implement-posthook.sh` | `mkdir`/`cp` failures swallowed | Both report and fail. The swallowed case was the worst available: "consolidated N files", exit 0, and the ledger events the next phase needs never arrived. |
| 3651148691 | `xruntime-review.js` | unguarded `fs.writeFileSync` | Controlled exit 2. An uncaught ENOSPC would kill the process before anything is journaled, and an unjournaled invocation is invisible to both the breaker and the convergence gate's attestation check. |

Separately: **`local_patches[]` was documentation only** — nothing verified a declared patch was still applied. Entries now carry `markers`, strings that exist only while the patch is applied, and the guard fails when one is missing. A test demonstrates `review-bundle-verify` green on a tree where a simulated upgrade reverted the patch. A patch declaring `reapply_on_upgrade` without markers is itself a failure.

## Skipped, with reasons

All five are pre-existing, upstream-owned code that this PR only made visible by tracking it. Each would add a permanent re-apply obligation (every bundle upgrade overwrites it) for an issue #305 did not introduce. They belong upstream in the roster repo.

| # | File | Why skipped |
|---|---|---|
| 3651148663 | `normalize-rules.js` / `review-finding.schema.json` | **Verified not live.** `validateFindings()` runs on input candidates at `review-normalize.js:171`, and `splitReobservations()` adds `pending_check` afterwards at line 260; the result goes into `dispositions`, never back through the validator. Nothing fails closed today. Adding a property changes a bundle file's sha for a speculative future check. |
| 3651148666 | `review-convergence-rules.js` | Real gap — `typeof entry.reviewed_sha === "undefined"` accepts `null` and `""`. Pre-existing, not introduced here. |
| 3651148668 | `review-trace-rules.js` | Not a defect. `schema/review-json-schema.md` documents the exclusion explicitly: "`skipped-degraded` and `blocked` are never corroborated." Changing it is a spec change and upstream's call. |
| 3651148673 | `review-trace.js` | Real gap — entry built and appended without validating against `review-trace.schema.json`. Pre-existing. |
| 3651148681 | `REVIEW-BUNDLE.md` | The doc already instructs replacing `REF` with trusted source coordinates, and the installer is fetched rather than shipped. Upstream doc content. |

## Verification

18 + 35 + 21 + 23 + 30 assertions, all green from a fresh `--no-local` clone, plus the repo-wide Alcotest (541 files / 59 suites) and test-alias guards. F1's fix is mutation-checked: reverting it turns 10 assertions red including the E2E. The new bundle-guard suite runs in CI.

mathiasbourgoin:
Round 3: F5–F9 and F11. All confirmed against the reviewer's reproductions before fixing.

## F5 — false attestation was the worse half, and it was separate from the fail-open

The `ls-remote` fail-open was already fixed in the previous round (bounded timeout, refusal on failure). The second half was not, and it is the more damaging one: `STALE_NOTE` was initialised to `"base freshness verified"` and only ever overwritten inside the `origin/*` arm, so **any non-remote base skipped the check entirely and still reported verified**. A caller could not distinguish "checked and fresh" from "never checked".

Now the note starts at `NOT CHECKED` and only becomes `verified against origin/<branch>` after an actual successful comparison. A non-remote base warns explicitly. Two assertions guard it: one that a local base says NOT CHECKED, and one that the literal string "base freshness verified" is **absent** from that output — plus a positive control that a genuinely fresh origin base still says verified and names the ref.

Real-world, `--base HEAD`:
```
warning — base freshness NOT CHECKED — HEAD is not a remote-tracking ref.
  Freshness can only be checked for an origin/* base. This base may be
  behind whatever it was branched from; nothing here verifies otherwise.
```

## F6 — both halves fixed

`--check-only`, whose documented contract is "run every precondition check", was the one mode that skipped one. The collision refusal lived after the early exit, so it reported `CHECKS PASS` on a directory another agent was live in. Moved before the exit; the test reproduces your `OTHER_AGENT_IS_HERE` fixture.

The partial-tracking downgrade is the one that worried me most, because `~/dev` really is one `git add SPOC/README.md` away from re-opening #101. `ls-files --error-unmatch` succeeds on a single match, which is the wrong question. A worktree materializes only **tracked** files, so the right question is whether the project is tracked *completely*: any non-ignored untracked file is a file the agent will not have. Partial tracking now refuses and names the missing files, with `--allow-partial-tracking` as the explicit override. Your exact fixture (outer repo tracks `proj/STRAY.md` only) is a test case, alongside a positive control that a **fully** tracked subdirectory still proceeds.

I also removed the codified warning path you flagged — the old suite asserted "tracked subdirectory warns but proceeds" on a partially-tracked fixture, which is precisely how a test locks in a bug.

## F7 — fixed, and the class of bug generalised

`dirname .` is `.`, so a relative root never reached `/`. Root is absolutized before anything walks its ancestors, and the walk terminates on a **fixpoint** rather than on reaching `/` — belt and braces, because an unbounded loop inside a dispatch guard hangs the dispatch. `AGENT_WORKTREE_ROOT` goes through the same path and has its own test. `--root`, `--base` and `--branch` are now validated; `--agent` was the only one that was.

## F8 — the check tested *mention*, not *registration*

All seven evasions confirmed and fixed. It now extracts the identifiers actually passed to a registration form and compares against defined cases, so a mention grants nothing; liveness is reachability from registration roots through binding bodies, which keeps genuine helpers out of the report without an exclusion list. The lexer handles nested comments, quoted strings, even/odd backslash runs and char literals, preserves line numbers, and **on an unterminated string or comment restores the raw tail instead of blanking it** — that blanking was the mechanism behind evasions 6 and 7.

Investigating its own output turned up something you did not list: **a second registration form already in use in this repo**. `sarek/tests/unit/*` registers with the raw tuple `("mangle_name", \`Quick, test_mangle_name)` and no `test_case` constructor at all. Cases registered that way were invisible to the check.

Two apparent orphans in the real tree were false positives, both helpers whose names merely start with `test_` (`test_kernels` returns a kernel list; `test_near_zero` takes two arguments). Neither can be passed to `Alcotest.test_case`. Fixed by reachability rather than by an exclusion list.

**I verified this independently of the agent that wrote it**, because a clean run over 59 suites is exactly the shape of result that deserves suspicion. My own fixtures for the laundering, and-chain and module-nested cases are each caught with file:line. Red-on-mutation on copies of real repo files (repo untouched): deleting one raw-tuple registration reports `test_mono.ml:114`; deleting one `Alcotest.test_case` reports `test_codegen_ptx_conformance.ml:599`. Both forms fail correctly, so the green is not vacuous.

I also checked the scope guard, since narrowing what gets checked is its own way to go green: **no file containing `Alcotest.run` is excluded by it.** One residual limitation is documented in the source — a helper file that defines cases using bare `assert` and never mentions Alcotest is out of scope.

## F9 — the core fix, plus one rule I deliberately did **not** ship

The precise check is `--expect-phase`: the **last** ledger event must be the phase the hook exists for. That catches your scenario exactly — `/roster-implement` dies before appending its event, the hook validates the stale pre-implement ledger and hands it to review — and it has no false positives by construction, since the hook only runs after that phase. A ledger with no `implement` event at all is now reported as a STALE ledger by name.

**I did not ship a phase-order rule, and I want to be explicit about why rather than leave it looking overlooked.** I built one and measured it against the 34 real ledgers in `briefs/` first:

```
ledgers=34  forward-jump=7  after-ship=1  below-floor=0

FORWARD-JUMP: deriver-variant-fields  full  research>spec>plan>implement>review>qa>ship
FORWARD-JUMP: fix-critical-semantics  full  intake>plan>implement>review>review>qa>ship
FORWARD-JUMP: glsl-ffi-fallback       full  research>spec>plan>implement>review>qa>ship
   (+4 more)
```

Full-mode runs legitimately skip optional phases — `research > spec > plan` has no `intake`, `intake > plan` has no `spec` — and the ledger records nothing that separates a deliberate skip from an erroneous omission. A rule firing on a fifth of valid input is not a gate; it is something people learn to ignore. `SEQ`-as-membership-only is therefore now a documented deliberate limitation in `ledger-schema.js`, not an oversight. If you want order enforced, the ledger needs to record skips explicitly first — happy to take that as its own task.

The three secondary points are all fixed:
- the Alcotest **scanned-count is consumed**, not just the exit code — `scanned 0 .ml file(s)` exits 0, so a wrong `--project` produced a confident green from a checker that examined nothing
- the jq cross-check's **absence is reported** — needing jq *and* an untracked skill file made "ledger schema OK" mean two different things depending on the machine
- `PROJECT` no longer silently accepts whatever `git rev-parse --show-toplevel` returns; the hook states which project it acted on and refuses an auto-detected project with no `briefs/`

## F11 — fixed

Default is now `$AGENT_WORKTREE_ROOT`, else `$XDG_CACHE_HOME/agent-worktrees`, else `~/.cache/agent-worktrees`. A test greps the script to assert no machine-specific mount is hardcoded anywhere in it — including in comments, which is what caught my first attempt.

## Verification

Bootstrap 35 assertions, post-hook 33, xruntime 35, bundle guard 18, verdict-assemble 30. Every negative paired with a positive control. Real-world checks: `--project ~/dev/interviews` still refuses and names `~/dev`; the post-hook validates the real `f16-dsl-slice1` ledger with `--expect-phase qa`, cross-checks against jq, and scans 3371 `.ml` files across 326 suites.

**One question: there is no F4.** Both messages went F3 → F5. If it exists I have not seen it, and I would rather ask than let it be the finding nobody noticed was missing.

mathiasbourgoin:
@coderabbitai review

mathiasbourgoin:
@coderabbitai full review

mathiasbourgoin:
Round 4: findings 1–6 all fixed, each reproduced on the reviewer's exact scenario first. The "plus, confirmed" list is deferred to a follow-up ticket (text handed to the coordinator) — none of it defeats a headline claim.

## Where the reviewer's mechanism was wrong, and it mattered

**Finding 1's ENOENT variant** was described as a digest mismatch — journal records `<runtime>:version-…`, lookup uses the prompt digest. It isn't. The placeholder digest is stable and *does* match; the real cause is that the version-probe branch ran **before** the breaker was consulted, so it was never asked. Same symptom, different bug — fixing the described cause would have shipped a no-op.

**Finding 5's false negative** was given as `Alcotest.run "S" []` with a `dead_suite` above it reporting OK. That case reproduced as **already working** (exit 1, case named). The live false negative is the neighbouring one, `if false then [...] else []`, where the case is textually inside the suite expression.

Both are recorded because a fix aimed at a misdiagnosis is worse than no fix — it closes the ticket without closing the hole.

## Fixed

| # | What was wrong | Now |
|---|---|---|
| 1 | Breaker suppressed only every *other* probe — the refusal record is itself journaled, and `skipped-degraded` ≠ `degraded`, so it cleared the arm. Measured: `degraded, skipped-degraded, degraded, …` | Journal scan skips non-probe records; audit trail intact. Six probes → one `degraded` then refusals throughout |
| 1b | A runtime whose binary is absent never reached the breaker at all (5/5 probes ran) | Breaker consulted before the version-probe branch |
| 2 | `--exclude-standard` treats ignored files as "not missing", so an ignored project + one `git add -f` passed as tracked | Refuses when the outer repo's ignore rules cover the project; **not overridable** |
| 3 | `markers: []` accepted and vacuous — confirmed exit 0 with the #102 patch genuinely reverted | Empty arrays and empty needle lists fail; counted by needles, not keys |
| 4 | `local_patches[]` policed itself from the file an upgrade regenerates — dropping the key returned "OK — sha-matched" | Guard carries its own `REQUIRED` (file, marker) list, independent of the manifest |
| 5 | Wrapper layout `run "S" suite` red on correct code | Takes the run call's argument region, bounded at the next top-level phrase |
| 5b | `chmod 000` flipped "1 unregistered" → "OK, 0 scanned" | Named and exit 2 |
| 6 | `--expect-phase` trailing or empty turned the stale-ledger refusal into exit 0 | Usage error; explicit `--expect-phase none` is a *recorded* skip that says so |

Old and new checkers give identical results on the same tree (553 files / 68 suites), so the finding-5 fix altered neither count nor verdict.

## Recorded, not fixed

Two finding-5 false negatives need more than lexing: a case registered inside dead code needs OCaml evaluation, and a test file no dune stanza builds needs dune awareness. Both are in a `KNOWN LIMITATIONS` block in the source with reproducers. The dead-code one has a test asserting the *current* behaviour **plus** the presence of that documentation, so closing the gap makes the test fail loudly instead of letting the limitation outlive its comment.

## Three of my own, found while fixing these

- `--allow-partial-tracking` was **rescuing** the #101 defect in my first attempt at finding 2. An escape hatch on a guard whose purpose is to refuse is a hole with a nicer name.
- `check-ignore` consults the index by default, so the force-added file **masks the very rule being checked**; `--no-index` is load-bearing. Now a source comment.
- Four tests I appended to the Alcotest suite landed **after** its summary print — they executed but could never report, able to fail the build without saying why. That is this PR's own defect, in this PR's own test file.

Two fixtures had to change because they put their marker in a comment and were asserting the weaker behaviour. Updated rather than relaxing the check, same call as deleting the F6 warning-path assertion.

Suites: 39 Alcotest, 41 xruntime, 38 bootstrap, 33 post-hook, 30 bundle-guard, 30 verdict-assemble — green from a fresh `--no-local` clone.

Issue #97, opened 2026-01-12
Sprint 3 Benchmarks - Complete Infrastructure Integration


mathiasbourgoin:
## Overview

Complete migration of Sprint 3 (P1) benchmarks to use the shared `Benchmark_runner` infrastructure with full JSON output, descriptions, and benchmark suite integration.

## Sprint 3 Benchmarks

Three scientific computing benchmarks focusing on physics simulation and PDE solvers:

1. **N-Body Simulation** (`bench_nbody`) - Naive all-pairs gravitational forces
2. **2D Convolution** (`bench_conv2d`) - 3×3 box blur image filtering
3. **2D Stencil** (`bench_stencil_2d`) - 5-point Jacobi iteration

## Scope

Following the same infrastructure integration pattern as Sprint 2 (issue #93):

### Phase 3.1-3.2: Benchmark Migration ✅
- [x] Migrate `bench_nbody.ml` to use `Benchmark_runner`
- [x] Migrate `bench_conv2d.ml` to use `Benchmark_runner`
- [x] Migrate `bench_stencil_2d.ml` to use `Benchmark_runner`
- [x] Update function signatures to accept `~device ~size ~config`
- [x] Replace console output with JSON `Output.device_result`
- [x] Use `config.warmup` and `config.iterations`
- [x] Convert timing to milliseconds

### Phase 3.3: Integration ✅
- [x] Add to `run_all_benchmarks.sh` build targets
- [x] Add execution entries with progress indicators

### Phase 3.4: Documentation ✅
- [x] Create `descriptions/nbody.md` (all-pairs O(N²), compute-intensive)
- [x] Create `descriptions/conv2d.md` (9-point stencil, memory-bound)
- [x] Create `descriptions/stencil_2d.md` (5-point stencil, memory-bound)
- [x] Include algorithm overview, Sarek kernel code, performance characteristics

## Changes Summary

- **3 benchmarks migrated**: N-Body, 2D Convolution, 2D Stencil
- **3 description files**: Comprehensive markdown documentation
- **Integration**: Added to benchmark runner script
- **Net change**: +1231 lines across 8 files

## Default Sizes

- **N-Body**: [512, 1024, 2048, 4096] particles
- **Conv2D**: [256, 512, 1024, 2048] pixels per side
- **Stencil 2D**: [256, 512, 1024, 2048] cells per side

## Throughput Metrics

- **N-Body**: G interactions/s (billion particle-particle interactions)
[...]

PR #96, opened 2026-01-11
Phase 3: Full Sprint 2 Infrastructure Integration

mathiasbourgoin:
## Overview

Complete Sprint 2 benchmark infrastructure integration with JSON output, descriptions, backend code generation, and runner integration.

Closes #93 (Phase 3 portion)

### 3.1-3.2: JSON Output & Multiple Sizes
- Migrated all 5 Sprint 2 benchmarks from `run_simple` to `run_benchmark`
- Benchmarks now return `Output.device_result` instead of bool
- Support multiple sizes via `--sizes` flag
- Use shared `config.warmup` and `config.iterations`
- **Benchmarks**: scan, bitonic_sort, histogram, gather/scatter (split), radix_sort
- **Default sizes configured**:
  - scan: [64, 128, 256]
  - bitonic_sort: [1024, 4096, 16384]
  - histogram: [1M, 10M, 50M]
  - gather/scatter: [1M, 10M, 50M]
  - radix_sort: [1M, 10M, 50M]

### 3.3: Integration with run_all_benchmarks.sh
- Added all Sprint 2 benchmarks to build targets
- Added execution entries with progress indicators
- Removed obsolete TODO note about Sprint 2 integration
- All benchmarks now generate JSON output

### 3.4: Benchmark Descriptions
Created comprehensive Markdown descriptions for 6 benchmarks:
- `scan.md`: Prefix sum with Hillis-Steele algorithm
- `bitonic_sort.md`: Data-oblivious parallel sorting
- `histogram.md`: 256-bin histogram with atomic operations
- `gather.md`: Indirect read operations
- `scatter.md`: Indirect write operations
- `radix_sort.md`: Multi-pass sorting with 4-bit radix

Each description includes algorithm overview, Sarek kernel code, performance characteristics, and expected results.

### 3.5: Backend Code Generation
- Added gather and scatter kernels to `generate_backend_code.ml`
- Generated CUDA, OpenCL, Vulkan GLSL, and Metal implementations
- Output saved to `benchmarks/descriptions/generated/`
- Other kernels (scan, bitonic_sort, histogram, radix_sort) use complex features better documented inline

### 3.6: Benchmark Viewer Compatibility
- ✅ `to_web.ml` already compatible with Sprint 2 JSON format (no changes needed)
- Verified Sprint 2 JSON files work correctly with web viewer

## Testing

Tested benchmarks manually:
```bash
# Scan benchmark
dune exec benchmarks/bench_scan.exe -- --sizes 128
[...]

Issue #85, opened 2026-01-11
Generated Metal Code displayed on website is not pretty printed


mathiasbourgoin:
Kirc_kernel.source_for_backend should return the source code for each backend. It works but the Metal backend does not pretty prints source.

All the benchmarks generated code for Metal should also be regenerated once pretty printing is implemented.

mathiasbourgoin:
I think the generated code was actually pretty printed, maybe it's the js code that does not render it correctly. The kernel code is currently rendered as a oneliner for instance, for mandelbrot:

```metal
#include <metal_stdlib>
using namespace metal;
kernel void sarek_kern(device float a [[buffer(0)]], constant int &sarek_a_length [[buffer(1)]], device float b [[buffer(2)]], constant int &sarek_b_length [[buffer(3)]], device float* c [[buffer(4)]], constant int &sarek_c_length [[buffer(5)]], constant int &m [[buffer(6)]], constant int &n [[buffer(7)]], constant int &k [[buffer(8)]], uint3 __metal_gid [[thread_position_in_grid]], uint3 __metal_tid [[thread_position_in_threadgroup]], uint3 __metal_bid [[threadgroup_position_in_grid]], uint3 __metal_tpg [[threads_per_threadgroup]], uint3 __metal_num_groups [[threadgroups_per_grid]]) { int tid = __metal_gid.x; int row = (tid / n); int col = (tid % n); if (((row < m) && (col < n))) { float sum = 0.0f; for (int i = 0; i <= (k - 1); i++) { sum = (sum + (a[((row k) + i)] b[((i * n) + col)])); } c[((row * n) + col)] = sum; } }
```

PR #308, opened 2026-07-25
fix(sarek): close the wrong-width family at the source-type mappers (#85, #96, #97)

mathiasbourgoin:
Closes the "wrong-width / silently-wrong-type" family at the source-type mappers: the three reported members (#85, #96, #97), two more the added validator found, and the rendering defect that made #97 unreadable.

## The validator (the main deliverable)

`sarek/tests/unit/test_type_width_totality.ml` sweeps **every** scalar source type through the width-critical mappers (`elttype_of_typ`, `slot_elttype_of_typ`, the vector-element path, `c_type_of_typ`) and requires, for each:

> either an explicit located rejection, or `byte_width(device element) = byte_width(host element)`

A silent width change is never admissible; a rejection always is. That is the invariant every member of this family has broken.

**It cannot go vacuous.** The source enumeration is unfolded from a total successor chain with no wildcard; the IR width function is a total match on `Sarek_ir_ppx.elttype`; the host width table is a total match on the source type; and the enumeration size is asserted. A new constructor on either side is a *compile error in the test*, not a silently-skipped case.

Run against `main` it fails on `char` (expected 1, received 4) on all three element mappers, and on `float16` through `c_type_of_typ`. That is how #96 and the two new instances below were found.

## Fixes

| | |
|---|---|
| **#96** — `char` read at 4-byte stride (family #5) | `elttype_of_typ` mapped `TReg Char -> Ir.TInt32` ("char represented as int32"). `Vector.char` is a Bigarray of chars, **one** byte. A `char vector` kernel emitted `int* o, int* a` and strode the buffer at 4×, silently. Now rejected at lowering with a diagnostic naming the cause. **Nothing is lost**: `Execute.check_launch_args` already refused a `Char` vector against a `TInt32` parameter on physical width, on both the device and interpreter entry points, so such a kernel could never launch. Only the timing and the message change. |
| **NEW** — `TReg (Custom _)` claimed a 4-byte scalar | Same arm, same shape: an unregistered type name (from `%sarek_intrinsic`'s fallback) got `Ir.TInt32` under the comment "Custom types handled separately". Now rejected, with a message saying how to register it. |
| **NEW** — `float16` declared as C `int` | Both copies of `c_type_of_typ` ended in `\| _ -> "int"`, so a 2-byte f16 became a 4-byte `int` member. Made total in `Sarek_ctype_gen`. |
| **#85** — variant payload computed then discarded (family #4) | `Sarek_ctype_gen` ran `let _ = typ_of_core_type ~loc ct in ... None`, so `Shade of float32` emitted `int col_sarek_Shade_t;` and a builder `build_col_Shade()` taking no argument at all — wrong type **and** no way to set it. The payload type is now threaded through. |
| **#97** — polymorphic `[@sarek.module]` helper at a non-default element type | Two defects, one rendering and one message — see below. |

## #97: the deliverable is the diagnostic, and here is why

A helper whose body calls a float32-only stdlib intrinsic genuinely *cannot* be monomorphised at float64. Making that instantiation work means overloading the stdlib's math intrinsics over element types — a language change, not a bug fix. The rejection was already correct; what was broken was how it reached the user.

**1. The location was garbage.** `Sarek_ast.loc` kept only line+column and rebuilt the `Lexing.position` with `pos_bol = 0; pos_cnum = column`. `pos_cnum` is an **absolute file offset**, and the driver seeks to `pos_bol` to echo the offending line — so it quoted bytes from the top of the file under a diagnostic naming line 10:

```
File "p64.ml", line 10, characters 15-31:
10 | ...............float
   |
10 | let[@sare..................................................
Error: Cannot unify types: float32 and float64
```

`loc` now carries `loc_bol`/`loc_end_bol` so `loc_to_ppxlib` is a true inverse of `loc_of_ppxlib`. **This fixes the rendering of every Sarek diagnostic**, not just this one — the existing negative tests now point at the right source too.

**2. The message named no callee.** A new `Instantiation_mismatch` replaces a bare `Cannot_unify` when the callee is an `EVar`. After both fixes:

```
File "p64.ml", line 10, characters 17-37:
10 |       o.(tid) <- norm a.(tid) a.(tid)]
                      ^^^^^^^^^^^^^^^^^^^^
Error: 'norm' cannot be used at this call site: float32 and float64 do not match.
       A [@sarek.module] helper is monomorphised per call site, so a polymorphic
       helper can only be instantiated at the element types its BODY admits — a
       body using a float32-only operation (the Sarek_stdlib.Std math intrinsics
       are float32) cannot be instantiated at float64. Give the helper a concrete
       type, or use the element-type-specific module (e.g. Sarek_float64).
```

## Removed: the write-only constructor-string emitters in `Sarek_lower_ir`

`c_type_of_typ` / `record_constructor_strings` / `variant_constructor_strings` fed `Kirc_types.register_constructor_string`, which appends to a list **nothing in the tree reads** — the backends build record and variant definitions themselves from `kern_types`/`kern_variants` via `Sarek_ir_codegen.gen_variant_def`.

They are removed rather than repaired **because they could not be repaired**: replacing their `\| _ -> "int"` wildcard with the rejection the class demands immediately broke a real kernel (`e2e/test_static_tag_erasure`), since this path is handed types it has no C name for and depended on the placeholder to keep going. A generator that works only while it is silently wrong, and whose output no one reads, is dead weight with a live hazard attached.

The `test_lower_ir` `c_type_generation` and `constructor_generation` groups go with them; the invariant they were nominally about is now enforced for the **live** mappers, against **host** widths, by the new validator.

## Reported, not changed: `[%%sarek.type]` is not wired up

`Sarek_ctype_gen` is reachable only from `Sarek_ppx.expand_sarek_type`, and the `[%%sarek.type]` **extension** is absent from the driver's rule list (`sarek_type_extension` is bound to `()`), so writing it fails with ppxlib's "Uninterpreted extension" — loudly, so nothing depends on it. The live registration path is the `[@@sarek.type]` **attribute**. Its emitter's naming convention (`build_X_sarek`, `X_sarek_tag`) has also drifted from what the backends emit (`make_<mangled>_<ctor>`, `int tag`, `union {...} data`). I fixed the defect and documented the situation in both module headers rather than leave a corrected-but-unreachable trap or wire up an emitter whose output would not match — reconciling the two conventions is an owner decision.

## Tests

Each was written first and **observed RED on this tree**, with a message naming the real defect, before the corresponding fix.

- `test_type_width_totality` — 5 cases, the class validator
- `test_diagnostic_locations` — 4 cases, the loc round trip
- `test_ctype_gen` — 4 cases, variant payload + the `int` wildcard
- `negative/test_char_vector`, `negative/test_poly_helper_f64` + Makefile stanzas

## Verification

Hardware: AMD RX 7900 XTX + Ryzen 7950X iGPU — radeonsi OpenCL, RADV Vulkan, CPU Native. **No NVIDIA GPU present, so CUDA/PTX is compile-only here.** No Metal hardware.

- `dune build` (default target) — clean
- `dune runtest` over `unit`, `codegen_golden`, `new_runtime`, `sarek/core/test`, `spoc`, `sarek/ppx/test` — all pass
- `make test_negative` — every stanza PASS, including the two new ones
- e2e on hardware: `test_static_tag_erasure`, `test_ktype_record`, `test_klet_variant`, `test_helper_tuple_ret`, `test_module_poly`, `test_transpose` — all PASSED

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,6 +134,10 @@ test_negative:
 	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_shared_tuple.cma > "$$out" 2>&1; if grep -q "Tuple-typed shared-memory arrays are not supported" "$$out"; then echo "  PASS: tuple shared array rejected"; else cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
 	@echo "Testing recursive-call vector-swap rejection (tail-recursion + vector)..."
 	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_tailrec_vec_swap.cma > "$$out" 2>&1; if grep -q "must pass its own vector parameter" "$$out"; then echo "  PASS: recursive vector-swap rejected"; else cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
+	@echo "Testing char element-type rejection (1-byte host vs 4-byte device stride)..."
+	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_char_vector.cma > "$$out" 2>&1; if grep -q "is not a supported Sarek element type" "$$out"; then echo "  PASS: char element type rejected"; else cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
+	@echo "Testing polymorphic-helper instantiation diagnostic (names the callee)..."
+	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_poly_helper_f64.cma > "$$out" 2>&1; if grep -q "'norm' cannot be used at this call site" "$$out"; then echo "  PASS: polymorphic helper instantiation named the callee"; else cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
 	@echo "All negative tests checked (see KNOWN-ISSUE line above for the one non-blocking gap, if any)"
 
 
--- a/sarek/ppx/README.md
+++ b/sarek/ppx/README.md
@@ -349,7 +349,7 @@ type 'a local   = Local of 'a        (* Thread-local / register *)
 
 ### IR Generation & Code Generation
 - **[Sarek_lower_ir](Sarek_lower_ir.ml)** - Generate Sarek IR
-- **[Sarek_ctype_gen](Sarek_ctype_gen.ml)** - C struct/union/builder strings for `[%%sarek.type]`
+- **[Sarek_ctype_gen](Sarek_ctype_gen.ml)** - C struct/union/builder strings for the `[%%sarek.type]` extension (currently unwired — see the module header; the live path is the `[@@sarek.type]` attribute, whose device definitions come from `Sarek_ir_codegen`)
 - **[Sarek_quote](Sarek_quote.ml)** - Quote kernels as OCaml expressions
 - **[Sarek_quote_ir](Sarek_quote_ir.ml)** - Quote IR constructors
 - **[Sarek_native_gen](Sarek_native_gen.ml)** - Native OCaml code generation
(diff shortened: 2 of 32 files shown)


mathiasbourgoin:
@coderabbitai review

PR #110, opened 2026-01-12
Add benchmark results from drangleic (2026-01-12)

picdc:
## Release Notes

* **Chores**
  * Added comprehensive benchmark results for compute kernels (bitonic sort, convolution, dot product, gather, histogram, Mandelbrot, matrix operations, n-body, radix sort, reduction, scan, scatter, stencil, stream triad, transpose, and vector operations) across CUDA, OpenCL, and Vulkan frameworks at various problem sizes for performance tracking and regression testing.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #315, opened 2026-07-26
fix(backends): choose the FP build options instead of inheriting or discarding them (#136, #125)

mathiasbourgoin:
Backlog **#136** and **#125** (backlog numbering — see the CodeRabbit note below; these are *not* the GitHub PRs of those numbers).

---

## The headline: the measurement killed the fix

An Apple M4 was made available mid-review, so **#125 no longer ships unverified — and running it showed the fix was the wrong lever.**

The original change set `fastMathEnabled = NO` and presented it as a contraction defence. On `o = a*b + c`, over the **8773 of 65536** elements where the device's *own* `fma` differs from the separately-rounded value:

| build | contracted |
|---|---|
| default options | 8773 / 8773 |
| `mathMode = MTLMathModeSafe` | **8773 / 8773** |
| `mathMode=Safe` + `fpFunctions=Precise` | **8773 / 8773** |
| `fastMathEnabled = NO` | **8773 / 8773** |
| **`#pragma METAL fp contract(off)`** | **0 / 8773** |

**It would have defeated zero contractions.** `a*b+c` is bit-identical under every compile-option setting; only the source pragma moves the number.

This is the **third** barrier in this repository that protected nothing — after the CUDA `"+f"` inline asm (§4) and the GLSL `precise`/`NoContraction` decoration on the RADV f16 narrowing (§6). §1 corollary 2 is now the most load-bearing sentence in the policy doc: *a flag that names the hazard is not a mechanism that prevents it.*

`Sarek_ir_metal` now emits the pragma in every kernel. The compile options are still set — but for math-function accuracy, and they are no longer described as anything else.

### Checked, not assumed: the options *are* honoured

A successful compile proves plumbing, not semantics, so this needed its own liveness check — the same discipline that made the OpenCL control valuable. On `sqrt(a) + 1/a`, 65536 inputs, vs default:

| setting | results changed |
|---|---|
| `mathMode = Safe` | 16017 / 65536 |
| `mathMode=Safe` + `fpFunctions=Precise` | **22135 / 65536** |
| `fastMathEnabled = NO` | 22135 / 65536 |
| `mathMode = Fast` / `fastMathEnabled = YES` | 0 (confirms the default) |

So Metal is **not** in rusticl's class: it honours these options, they are simply orthogonal to contraction. And it is **two knobs, not one** — a fresh `MTLCompileOptions` reports `mathMode=2` (Fast) *and* `mathFloatingPointFunctions=0` (Fast); setting `mathMode` alone leaves math functions on `metal::fast`, which is the 16017-vs-22135 gap. I would have missed the second knob without reading the SDK headers.

---

## #136 — OpenCL: `clBuildProgram` with no options

`Opencl_api.Program.build` passed an empty option string and every caller in the tree passed nothing. An empty option string is not "no policy" — it is whatever each vendor decided, and OpenCL's default permits a 3-ulp `sqrt` and a 2.5-ulp divide against §1's requirement that both are correctly rounded.

`Opencl_fp.build_options` now assembles the string at `Opencl_api.Program.build` — the single point an option string reaches `clBuildProgram`, so the guard covers flags a *future maintainer of this module* adds, not merely a caller's. Same placement argument as `Cuda_nvrtc.compile_with_string_opts` (§5).

### The measurement, and its limits

**Prove red before trusting green — and the red is not here.**

- The red is **quoted, not reproduced**: 1.81e-14 FAIL → 8.87e-15 PASS on a GTX 1070 Max-Q (sm_61, CUDA 12.9, driver 580.119.02), measured during #298. That device is not on this machine.
- **The red does not reproduce locally.** `test_real64` on rusticl/radeonsi 26.1.4-arch3.1, RX 7900 XTX and Raphael iGPU: OpenCL `sqrt` **9.68e-15 — already a PASS**.

**Measured OpenCL cost, as asked: not measurable on this box, and here is why rather than a number.** `tools/probes/opencl_build_options_probe.c`, 2^20 inputs/device, carries two controls:

| variant | sqrt vs correctly-rounded | bit-differs from baseline |
|---|---|---|
| baseline (empty string) | ≤1 ulp | — |
| `-cl-fp32-correctly-rounded-divide-sqrt` | ≤1 ulp | **0 / 1048576** |
| `-cl-fast-relaxed-math` (**FP liveness control**) | ≤1 ulp | **0 / 1048576** |
| `-DSAREK_PROBE_SCALE=…` (**plumbing control**) | 3 ulp | 1048576 / 1048576 |

The plumbing control **passes** — the option string does reach rusticl's compiler. The FP liveness control **fails** — even `-cl-fast-relaxed-math` changes nothing. rusticl delivers FP build options and ignores them, so *"the flag changed nothing"* is indistinguishable from *"the flag was discarded"*, and no accuracy or cost claim may be founded on these devices. The probe prints −3.0% / +0.2%; **those are run-to-run noise around an unchanged kernel and are labelled as such in the probe, the module and the policy.** The CUDA ~12% on `bench_nbody` is a different backend on a different device and does not transfer.

### The flag is capability-gated, and that is the load-bearing decision

The spec permits `-cl-fp32-correctly-rounded-divide-sqrt` only when `CL_DEVICE_SINGLE_FP_CONFIG` carries `CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT`; otherwise `clBuildProgram` returns `CL_INVALID_BUILD_OPTIONS`. Unconditional, it fails **every kernel build** on every device without the capability — total, not numerical.

**This machine would not have caught that.** Measured: both local devices report `fp_config = 0x6` (`INF_NAN | ROUND_TO_NEAREST`, no `CR_DIV_SQRT`, no `DENORM`) and **both accept the flag with `CL_SUCCESS` anyway** — rusticl permissive where the spec says refuse. Control: `-cl-this-option-does-not-exist` *is* refused, so "ACCEPTED" isn't the probe's answer to everything.

### The full audit — every default now chosen

| option | decision | why |
|---|---|---|
| `-cl-fp32-correctly-rounded-divide-sqrt` | **set**, gated | default off ⇒ 3-ulp sqrt / 2.5-ulp div; `Sarek_df64`'s Newton/Karp step squares its seed error |
| `-cl-denorms-are-zero` | never passed; **refused** | flushes binary32 subnormals, forbidden by §1 |
| `-cl-fast-relaxed-math` | never passed; **refused** | implies unsafe+finite math, permits `native_*` |
| `-cl-unsafe-math-optimizations` | never passed; **refused** | reassociation — destroys an EFT outright |
| `-cl-finite-math-only` | never passed; **refused** | assumes no NaN/Inf; the oracle assumes no such thing |
| `-cl-no-signed-zeros` | never passed; **refused** | `Sarek_df64` renormalisation relies on signed zero |
| `-cl-mad-enable` | never passed; **refused** | the contraction hazard by name |
| `-cl-single-precision-constant` | never passed; **refused** | silently demotes double literals |
| `-cl-opt-disable` | **allowed** | conservative; cannot relax FP semantics |
| `FP_CONTRACT` | **not addressable** | on by default, no build option turns it off; the source pragma was already measured not to work here. Still defeated by construction via `mul_rn` |

**Refused, not countered**: unlike HIP's append-last trick, OpenCL has no option that undoes `-cl-fast-relaxed-math`. Same reasoning as §5.

**Subnormals are flushed here regardless** — neither local device advertises `CL_FP_DENORM`. A device property Sarek cannot correct; recorded, not fixed.

### The guard has been seen to fire

Seven mutations, each red with a distinct message. The two that matter most are anti-vacuity:

- **assembled-string** — the gate could "pass" while assembling a string nobody uses.
- **fp-config-query-is-live** — the gate is *off* on every device here, so a silently broken `CL_DEVICE_SINGLE_FP_CONFIG` query would disable it everywhere and make every "correctly gated off" case pass while checking nothing.

---

## #125 — Metal: measured on an M4, and the measurement contradicted the fix

**Apple hardware became available, so this is no longer unverified — and running it changed the answer.** Apple M4, macOS 15.6.1 (24G90), Apple clang 17.0.0, Metal.framework from the Command Line Tools SDK. No Xcode needed: `newLibraryWithSource:options:error:` compiles through the driver at runtime, which is exactly the call under test.

### The previous revision of this PR was wrong

It set `fastMathEnabled = NO` and presented it as a contraction defence. On `o = a*b + c`, over the **8773 of 65536** elements where the *device's own* `fma` differs from the separately-rounded value:

| build | contracted |
|---|---|
| default options | 8773 / 8773 |
| `mathMode = MTLMathModeSafe` | **8773 / 8773** |
| `mathMode=Safe` + `fpFunctions=Precise` | **8773 / 8773** |
| `fastMathEnabled = NO` | **8773 / 8773** |
| **`#pragma METAL fp contract(off)`** | **0 / 8773** |

`a*b+c` is bit-identical under every compile-option setting. **This is §1 corollary 2 for the third time in this repo** — a flag that names the hazard is not a mechanism that prevents it. `Sarek_ir_metal` now emits `#pragma METAL fp contract(off)` in every kernel, and the compile options are no longer described as a contraction defence.

Method: the separately-rounded reference is computed in **two kernel passes** with the product round-tripped through device memory so it cannot itself be fused, and the `fma` is **read from the device** rather than modelled — the RADV lesson in §6.

### The options are still worth setting, and they *are* honoured

A compile that succeeds proves plumbing, not semantics, so this needed its own liveness check. On `sqrt(a) + 1/a`, 65536 inputs, vs default:

| setting | results changed |
|---|---|
| `mathMode = Safe` | 16017 / 65536 |
| `mathMode=Safe` + `fpFunctions=Precise` | **22135 / 65536** |
| `fastMathEnabled = NO` | 22135 / 65536 |
| `mathMode = Fast` / `fastMathEnabled = YES` | 0 (confirms the default) |

**Metal is not in rusticl's class** — it does not accept these and discard them.

### Two knobs, not one

A fresh `MTLCompileOptions` reports `mathMode = 2` (Fast) **and** `mathFloatingPointFunctions = 0` (Fast). Setting `mathMode` alone leaves math functions on `metal::fast` — that is the 16017-vs-22135 gap. Both are now set. "Metal defaults to fast math" was quoted from Apple's docs; it is now measured, and it is two defaults.

### `mathMode` vs `fastMathEnabled` — settled by measurement, as asked

The enum values are **read from the SDK** (`MTLLibrary.h:241-246`, `258-262`), which retires the previous revision's stated reason for avoiding `setMathMode:`. `fastMathEnabled` is deprecated since macOS 15.0 but `mathMode` doesn't exist before it, so **both** are kept, selected by `respondsToSelector:`. They are **equivalent, measured**: `fastMathEnabled=NO` and `mathMode=Safe + fpFunctions=Precise` are **bit-identical over 65536 elements** of `sqrt+recip+sin+log+exp`. The pre-15 fallback is exact, not degraded.

Also verified on the M4: all four selector strings the binding registers exist, and the raw `long`-taking `objc_msgSend` signature the ctypes code uses lands both values. All nine byte-exact generated goldens compile on the device with the pragma in place.

### Interpreter agreement: measured on the M4, and it holds

`test_df64` and `test_real64` now run on Apple M4 / macOS 15.6.1 / Apple clang 17. **Metal reproduces the interpreter's figures exactly, on every op:**

| op | Metal | Interpreter | tol | |
|---|---|---|---|---|
| `mul` | 9.07e-15 | 9.07e-15 | 1.42e-14 | PASS |
| `add` | 5.33e-15 | 5.33e-15 | 7.11e-15 | PASS |
| `sub` | 6.51e-15 | 6.51e-15 | 7.11e-15 | PASS |
| `div` | 5.08e-15 | 5.08e-15 | 1.42e-14 | PASS |
| `sqrt` | 8.53e-15 | 8.53e-15 | 1.42e-14 | PASS |

Sampled maxima over each test's input set, one device, agreement between summary statistics — not bounds, not element-wise identity.

### The pragma does *not* move df64 — and that is the correct answer

Rebuilt with the pragma emptied (binary confirmed rebuilt: new mtime, 80 bytes smaller), re-run: **every figure identical to the digit.**

That is `Sarek_df64` working as designed, not a failure of the pragma. The library defeats contraction **by construction** — `mul_rn` routes products through `fma`, so no fusable multiply survives — which is exactly why §2 already records "no flag" as the CUDA and OpenCL mechanism. **df64 is immune to contraction by design and therefore cannot detect it.** Expecting it to move the way `sqrt.rn.f32` did on NVIDIA conflates two mechanisms: that was an approximate *seed inside the algorithm*, which `mul_rn` doesn't protect against; this is contraction, which it does.

So the pragma protects exactly what df64 cannot: **caller-written kernels** — the §3 hazard that "lives in caller code, and no gate in this repository can see". `a*b+c` is contracted 8773/8773 without it and 0/8773 with it. The synthetic probe is the right instrument, and the only one.

### Can Metal leave "may NOT rely on"? Partially — and here is the line

**Yes for f32 arithmetic.** Moved off the list.

**No for:** f16 (never probed on Metal), subnormals (never probed), element-wise identity, any other device or OS, and **kernels using records or variants — which do not compile on Metal at all.**

### Two separate defects the M4 uncovered

1. **Two of nine generated Metal goldens do not compile at all.** `record_kernel`/`variant_kernel` emit `constant T* &` buffer params; Metal rejects the pointee address space. **Pre-existing** — control with the pragma stripped fails identically — and not fixed here, since `device` vs `constant` is a design decision. Needs its own issue.

2. **Six `opencl_validation_sweep` float64 cases fail on macOS**, because Apple Silicon's OpenCL has no `cl_khr_fp64`. Environmental, and impossible for this branch to have caused (it touches no OpenCL codegen, no f64 path). But worth flagging: cases 7-8 of that same sweep **SKIP** on f64 while 16-20 **FAIL** — the sweep's capability gating disagrees with itself, and is a good candidate for the "a skip is not a pass" treatment.

Everything else on the M4 is green, including all six `metal_contraction_pragma` cases.

**ladon was left exactly as found:** throwaway `/tmp` clone, nothing installed, no `opam switch` changed, and Mathias's dirty `apple_m4_benchmarks` checkout never touched.

## CodeRabbit

**Round 1 — three findings, all valid, all addressed** (commits `0e8a5b5`..`520359c5`):

- **`Metal_bindings.ml:334` — leak on fallback.** Real: after `init` succeeded, the selector-probing path returned `None` without releasing the object. Fixed.
- **`opencl_build_options_probe.c` — unclamped loops.** Real: `clGetPlatformIDs`/`clGetDeviceIDs` report the *true* count even when `num_entries` caps what was written, so >8 platforms/devices would read uninitialised stack. Both clamped, with a note printed on truncation.
- **Stale `#125` / `#136`.** Real ambiguity — but *removing* them, as suggested, would be wrong: they are **backlog** numbers, and this repo's numbering collides with GitHub's (`main` already says "tracked as #136"; the pre-existing `#57`/`#110`/`#111`/`#116` likewise don't match those GitHub PRs). Spelled **`backlog #NNN`**: keeps the identifier, drops the false link. Pre-existing bare numbers in untouched prose left alone — renaming the project's citation convention is a maintainer decision.

**Round 2 — the `CHANGES_REQUESTED` is stale, not new.** The three comments carry the same IDs and the same `2026-07-26T17:39:37Z` timestamp as round 1, and the check run reports `CodeRabbit — Review rate limited`, so it never re-reviewed the new commits. One of its own comments already reads *"✅ Addressed in commits 0e8a5b5 to 520359c"*. I re-verified all three against current code rather than trusting that: the clamps are at `opencl_build_options_probe.c:243,252`, the release is on the post-`init` `None` path, and every Metal reference reads `backlog #NNN`. Nothing further to fix; the review needs re-running or dismissing.

---

## HIP — does it share the omission? No. It had a neighbouring one.

Measured here, ROCm 7.2.4 / AMD clang 22.0.0git, gfx1100, `out = sqrtf(a) + a/b`:

| build | divide / sqrt | `denorm_mode_32` |
|---|---|---|
| what Sarek passed | refined: `v_div_scale`/`v_div_fmas`/`v_div_fixup_f32` + Newton sqrt (13 instrs) | **3** |
| `-fno-hip-fp32-correctly-rounded-divide-sqrt` (**control**) | bare `v_rcp`/`v_sqrt` + frexp/ldexp, no `v_div_fixup_f32` (10) | 3 |
| `-fgpu-flush-denormals-to-zero` (**control**) | refined | **0** |
| **with the two flags set explicitly** | **identical**, all 13 | 3 |

So HIP's inherited defaults were already conformant. The two flags are set anyway — **measured ISA-identical, zero cost** — because an inherited default is a choice nobody made. Since `base_options` is appended last, a caller passing either negated form is now neutralised (verified: denorm returns to 3, `v_div_fixup_f32` returns), and a future clang default change becomes a no-op instead of a silent regression.

**The limit of append-last, measured**: `-ffast-math` removes `v_div_fixup_f32` and is **not** rescued by appending the conformance flag, and nothing was found that undoes the `-cl-*` spellings.

**HIP's real gap was silence.** These passed the warning list unwarned, each measured to change gfx1100 codegen — now covered: `-Ofast`, `-cl-fast-relaxed-math`, `-cl-unsafe-math-optimizations`, `-fapprox-func`, `-fgpu-flush-denormals-to-zero`, `-fno-hip-fp32-correctly-rounded-divide-sqrt`, `-munsafe-fp-atomics`.

**Deliberately not done** — the first four are not neutralisable by anything, which is exactly when §5 says *reject*, not warn. That is a caller-visible behaviour change and belongs in its own issue. Until then a caller can still relax HIP float semantics and only read a warning.

Also drops the disputed **"373"** from `Hip_rtc.ml` and `test_hip_rtc_options.ml` — §2 records it as one of two mutually inconsistent in-tree uses. No count is asserted in its place, because the right one has not been established.

---

## Verification

- `dune build @all` — exit 0. `dune build @fmt` — exit 0.
- `dune runtest` — exit 0, zero `[FAIL]`.
- `test_real64` — **byte-identical** to the pre-change baseline. `test_df64` PASSED.
- `check-test-alias-coverage.sh` — OK, all 81 e2e executables wired. New tests are in `(tests (names …))` blocks, so dune wires them to `runtest`.
- **Every new check proved red by mutation** — 7 OpenCL, 4 HIP, 1 Metal fail-soft, 1 Metal codegen pragma (emitted an empty pragma → 5 goldens red) — then restored green.
- `ci/assert-toolchain.sh` fails **pre-existing** on this box: 3/10 checks, missing CUDA headers on an AMD-only machine. The identical script from `origin/main` fails the same way, and this PR touches nothing under `ci/`. Note it reports **10** checks, not the 12 expected — worth a look, separately.

**Hardware.** AMD box: RX 7900 XTX (radeonsi/rusticl OpenCL, RADV, HIP gfx1100) + Raphael iGPU — no NVIDIA. Apple M4 / macOS 15.6.1 / Apple clang 17.0.0 over SSH for all Metal figures. Nothing was installed or changed on the Mac; probes were built with the Command Line Tools already present and run from `/tmp`. Every figure is from one of those two machines unless explicitly marked as quoted from sm_61.

## Summary by CodeRabbit

- **New Features**
  - OpenCL floating-point conformance is now enforced per device capability, using computed effective build options.
  - HIP compilation uses stricter divide/sqrt rounding and denormal/flush handling defaults.
  - Metal code generation disables floating-point contraction via emitted kernel pragma, improving accuracy consistency.
- **Bug Fixes**
  - Metal now correctly applies conformant compile options (previously ineffective), restoring intended accuracy behavior.
  - Unsafe “relaxing” OpenCL FP-related build options are rejected.
- **Tests**
  - Added/expanded OpenCL, HIP, and Metal option-handling and contraction-regression suites; updated golden shader expectations.
- **Documentation**
  - Refreshed cross-backend floating-point contraction policy and mechanism/guard mappings with new evidence.

### DIFF
--- a/benchmarks/descriptions/generated/dot_product_generated.md
+++ b/benchmarks/descriptions/generated/dot_product_generated.md
@@ -245,6 +245,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device float* a [[buffer(0)]], constant int &sarek_a_length [[buffer(1)]], device float* b [[buffer(2)]], constant int &sarek_b_length [[buffer(3)]], device float* output [[buffer(4)]], constant int &sarek_output_length [[buffer(5)]], constant int &n [[buffer(6)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/gather_generated.md
+++ b/benchmarks/descriptions/generated/gather_generated.md
@@ -72,6 +72,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device int* input [[buffer(0)]], constant int &sarek_input_length [[buffer(1)]], device int* indices [[buffer(2)]], constant int &sarek_indices_length [[buffer(3)]], device int* output [[buffer(4)]], constant int &sarek_output_length [[buffer(5)]], constant int &n [[buffer(6)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/mandelbrot_generated.md
+++ b/benchmarks/descriptions/generated/mandelbrot_generated.md
@@ -99,6 +99,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device int* output [[buffer(0)]], constant int &sarek_output_length [[buffer(1)]], constant int &width [[buffer(2)]], constant int &height [[buffer(3)]], constant int &max_iter [[buffer(4)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/matrix_mul_generated.md
+++ b/benchmarks/descriptions/generated/matrix_mul_generated.md
@@ -93,6 +93,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device float* a [[buffer(0)]], constant int &sarek_a_length [[buffer(1)]], device float* b [[buffer(2)]], constant int &sarek_b_length [[buffer(3)]], device float* c [[buffer(4)]], constant int &sarek_c_length [[buffer(5)]], constant int &m [[buffer(6)]], constant int &n [[buffer(7)]], constant int &k [[buffer(8)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/matrix_mul_tiled_generated.md
+++ b/benchmarks/descriptions/generated/matrix_mul_tiled_generated.md
@@ -177,6 +177,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device float* a [[buffer(0)]], constant int &sarek_a_length [[buffer(1)]], device float* b [[buffer(2)]], constant int &sarek_b_length [[buffer(3)]], device float* c [[buffer(4)]], constant int &sarek_c_length [[buffer(5)]], constant int &m [[buffer(6)]], constant int &n [[buffer(7)]], constant int &k [[buffer(8)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/radix_sort_histogram_generated.md
+++ b/benchmarks/descriptions/generated/radix_sort_histogram_generated.md
@@ -130,6 +130,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device int* input [[buffer(0)]], constant int &sarek_input_length [[buffer(1)]], device atomic_int* histogram [[buffer(2)]], constant int &sarek_histogram_length [[buffer(3)]], constant int &n [[buffer(4)]], constant int &shift [[buffer(5)]], constant int &mask [[buffer(6)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/radix_sort_scatter_generated.md
+++ b/benchmarks/descriptions/generated/radix_sort_scatter_generated.md
@@ -82,6 +82,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device int* input [[buffer(0)]], constant int &sarek_input_length [[buffer(1)]], device int* output [[buffer(2)]], constant int &sarek_output_length [[buffer(3)]], device atomic_int* counters [[buffer(4)]], constant int &sarek_counters_length [[buffer(5)]], constant int &n [[buffer(6)]], constant int &shift [[buffer(7)]], constant int &mask [[buffer(8)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/reduction_generated.md
+++ b/benchmarks/descriptions/generated/reduction_generated.md
@@ -249,6 +249,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device float* input [[buffer(0)]], constant int &sarek_input_length [[buffer(1)]], device float* output [[buffer(2)]], constant int &sarek_output_length [[buffer(3)]], constant int &n [[buffer(4)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/reduction_max_generated.md
+++ b/benchmarks/descriptions/generated/reduction_max_generated.md
@@ -345,6 +345,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device float* input [[buffer(0)]], constant int &sarek_input_length [[buffer(1)]], device float* output [[buffer(2)]], constant int &sarek_output_length [[buffer(3)]], constant int &n [[buffer(4)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/scatter_generated.md
+++ b/benchmarks/descriptions/generated/scatter_generated.md
@@ -72,6 +72,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device int* input [[buffer(0)]], constant int &sarek_input_length [[buffer(1)]], device int* indices [[buffer(2)]], constant int &sarek_indices_length [[buffer(3)]], device int* output [[buffer(4)]], constant int &sarek_output_length [[buffer(5)]], constant int &n [[buffer(6)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/stream_triad_generated.md
+++ b/benchmarks/descriptions/generated/stream_triad_generated.md
@@ -71,6 +71,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device float* a [[buffer(0)]], constant int &sarek_a_length [[buffer(1)]], device float* b [[buffer(2)]], constant int &sarek_b_length [[buffer(3)]], device float* c [[buffer(4)]], constant int &sarek_c_length [[buffer(5)]], constant float &scalar [[buffer(6)]], constant int &n [[buffer(7)]],
 uint3 __metal_gid [[thread_position_in_grid]],
--- a/benchmarks/descriptions/generated/transpose_naive_generated.md
+++ b/benchmarks/descriptions/generated/transpose_naive_generated.md
@@ -83,6 +83,7 @@ void main() {
 ```metal
 #include <metal_stdlib>
 using namespace metal;
+#pragma METAL fp contract(off)
 
 kernel void sarek_kern(device float* input [[buffer(0)]], constant int &sarek_input_length [[buffer(1)]], device float* output [[buffer(2)]], constant int &sarek_output_length [[buffer(3)]], constant int &width [[buffer(4)]], constant int &height [[buffer(5)]],
 uint3 __metal_gid [[thread_position_in_grid]],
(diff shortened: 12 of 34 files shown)


mathiasbourgoin:
@coderabbitai review

PR #317, opened 2026-07-26
fix(sarek): make the declared intrinsic surface actually callable

mathiasbourgoin:
## What this fixes

Thirteen stdlib intrinsics were user-callable and unusable. I swept the whole
intrinsic surface rather than fixing the reported names, and the same class
turned up in **three** independent tables — one of which fails silently rather
than loudly.

A `let%sarek_intrinsic` is declared **once**, but three tables have to honour it,
and all three had drifted from the declaration.

### 1. `Sarek_pure_registry` — bare name list, five names missing

`float64_list` was a bare **name** list with no name→device-symbol mapping, and
five names declared by `Sarek_float64.Float64` were absent from it entirely:
`abs_float`, `copysign`, `expm1`, `log1p`, `hypot` (the report named two).

Observed, on hand-built IR before the fix:

```
[Float64] abs_float | CUDA=RAISE:Unknown intrinsic: Float64.abs_float
                    | OPENCL=RAISE:Unknown intrinsic: Float64.abs_float
                    | METAL=RAISE:Unknown intrinsic: Float64.abs_float
                    | GLSL=b[idx] = abs(a[idx]);
```

GLSL survives only because its `pre_hook` polyfills four of them and its `arm`
renames `abs_float`→`abs`, exactly as reported.

**The obvious one-line fix is the miscompile the table's own header warns
about**: adding `"abs_float"` to a list of *names* emits a call to
`abs_float(...)`, which no backend defines. `float64_list` is now a
`(sarek_name, device_symbol)` list, so `Float64.abs_float` emits `fabs` — and
`abs` on GLSL, since the rename table is keyed on the *Sarek* name and has to
survive the split. That last point is pinned by a test.

### 2. The native (`cpu_kern`) backend — no target symbol

The native backend picks a target module and then copies the intrinsic **name**
onto it verbatim. `Float32` pointed at `Sarek_cpu_runtime.Float32`, which had no
`abs_float` / `expm1` / `log1p` / `hypot` / `copysign` / `fmod` / `minus`;
`Float64` pointed at OCaml's `Stdlib.Float`, which is missing a third of the
Float64 surface (`abs_float`, `rsqrt`, `fmod`, `copysign`, `of_int32`,
`to_int32`, `of_float32`, `to_float32`, the four `*_float64` names, the eight
operator forms).

Both are compile errors inside PPX-generated code. This kernel could not be
written:

```
Error: Unbound value Sarek.Sarek_cpu_runtime.Float32.copysign
```

`Sarek_float32` gains the seven missing names; the new `Sarek_float64_native`
mirrors the Float64 stdlib surface name-for-name, each body copied from that
intrinsic's own `ocaml = ...` field (which is its host-side specification).

### 3. The FFI fallback discarded the caller's backend — silent, and on the hot path

`Sarek_registry.fun_device_template` asked for the framework `"generic"`, which
`cuda_or_opencl` resolves to the **CUDA** branch. Every backend reaching this
fallback got the CUDA spelling.

This is not a corner: a real PPX kernel records the *declaration-site* path
`["Sarek_stdlib";"Float32"]`, which the pure registry does not register, so
ordinary `Float32.*` calls land here. An OpenCL kernel calling
`Float32.abs_float` was emitted as `fabsf(...)`:

```
/tmp/sarek_gate_ocl_4a5c94.cl:3:12: error: use of undeclared identifier 'fabsf'
```

The stdlib already declared both spellings (`dev "fabsf" "fabs"`); only this
lookup was throwing the framework away. Metal joins the OpenCL branch: across
the whole stdlib the two branches differ *only* in the CUDA `f` suffix, and MSL
has no `fabsf` either.

## Closing the class, not the names

`sarek/tests/unit/test_intrinsic_surface.ml` derives its expectation from
`Sarek_registry`'s link-time record of every `let%sarek_intrinsic` — written by
the PPX from the declaration itself, so it is ground truth — and reconciles it
against the other two tables. A new stdlib intrinsic cannot ship unroutable
without someone remembering a list.

The native half uses a **witness table**: each entry references the real runtime
function, so a missing runtime function is a *build* error, and a completeness
check against the registry makes a missing witness entry a *test* failure.

## Every gate observed red before being trusted green

| Gate | Red output |
|---|---|
| pure-registry completeness | `["abs_float"; "copysign"; "expm1"; "hypot"; "log1p"]` |
| native target module | got `["Float"]`, expected `Sarek_cpu_runtime.Float64` |
| witness completeness | `["copysign"]` when one entry is dropped |
| OpenCL spelling + clang gate | `use of undeclared identifier 'fabsf'` |
| e2e native execution | value mismatches when `minus` is wired to `add` |

The clang check carries its **own negative control**: it re-injects the pre-fix
spelling into the emitted kernel and fails if clang accepts it, so a green
result cannot mean "the compiler wasn't looking".

`scripts/check-test-alias-coverage.sh`: `OK: all 82 e2e test executables are
wired to an alias`. `dune build @fmt` exits 0. `dune build @runtest` exits 0.
`benchmarks/descriptions/generated/` regenerated — **no diff** (the benchmarks
don't use the affected intrinsics).

## Hardware and toolchain

Verified on AMD RX 7900 XTX + Ryzen 9 7950X iGPU: rusticl/radeonsi OpenCL, RADV
Vulkan, plus the Native and Interpreter devices; `clang` for the OpenCL syntax
gate. **No NVIDIA, Metal or WebGPU device was involved** — the CUDA/Metal/WGSL
claims here are codegen-level, and toolchain-checked only where a checker exists
locally.

## Deliberately left (evidenced, not overlooked)

- **The pure registry does not register the paths the PPX actually emits.**
  `["Sarek_stdlib";"Float32"]` and `["Sarek_float64";"Float64"]` are what a real
  kernel carries; the registry holds `["Float32"]`, `["Math";"Float32"]`,
  `["Sarek_stdlib_meta";"Float32"]`, `["Float64"]`. So the framework-aware pure
  registry is bypassed on the common path and the FFI fallback does the work.
  Registering them is a two-line change, but it would make the pure registry
  win over WGSL's and Metal's polyfill arms for names those backends have no
  builtin for (`expm1`, `hypot`, `copysign`, `fabs` are not WGSL), converting a
  loud `Unknown intrinsic` into silently-invalid shader source. Doing it safely
  needs per-framework builtin coverage in the registry — and a WGSL/Metal
  toolchain I don't have locally to verify against.
- **WGSL already emits four invalid builtins** on the `["Float32"]` path today
  (`fabs`, `expm1`, `hypot`, `copysign`). Pre-existing, unchanged by this PR,
  and the same follow-up covers it.

## Summary by CodeRabbit

* **New Features**
  * Added native CPU support for Float64 intrinsics (math, conversions, comparisons, and operators).
  * Expanded Float32 intrinsic surface with additional stdlib-compatible math and arithmetic helpers.
* **Bug Fixes**
  * Fixed intrinsic registry device-template spelling by explicitly preserving the active framework across CUDA/Metal/OpenCL.
  * Improved Float64 pure-registry handling when OCaml/stdlib names differ from emitted device symbols.
* **Tests**
  * Added new native e2e + intrinsic surface unit tests, plus additional Float32 overflow regression coverage.

---

### Major: `expm1` skipped `check_overflow`

This lands in the **interpreter**, which is this project's cross-backend oracle,
so a mechanical `|> check_overflow` was not good enough. If `expm1` overflowed at
a different point here than on a device, every agreement test using it would be
comparing against a reference that is itself wrong at the extremes — and the
disagreement would get attributed to the device.

**Settled by exhaustive sweep, not inspection.** A C program replicating
`float32_of_float_impl` exactly — its clamp and its flush-to-zero included — run
over **all 2^32 binary32 inputs** against C's `expm1f` / `log1pf`:

| | spurious overflow | missed overflow | first `+x` → `inf` | 1-ulp diffs |
|---|---|---|---|---|
| `expm1` | **0** | **0** | `88.7228394` — *identical to `expm1f`* | 2 of 4.28e9 |
| `log1p` | **0** | **0** | `+inf` only — no finite input overflows | 11 of 4.28e9 |
| `exp` *(control, already wired)* | 0 | 0 | `88.7228394` | 314586 |
| `sqrt` *(control, cannot overflow)* | 0 | 0 | — | 0 disagreements at all |

**So the value was already correct, at a bit-identical threshold — only the
reporting was missing.** The defect is real but narrower than it looked: an
observability gap in `Warn`/`Exception`, not an oracle numerical defect. The
remaining ~1.6e7 disagreements per function are entirely flush-subnormals-to-zero,
which is `float32_of_float_impl`'s deliberate module-wide behaviour, shared by the
already-wired `exp` control, and untouched here.

`check_overflow` is **value-neutral in the default `Silent` mode** — it returns its
argument untouched — so this cannot perturb a single number the oracle produces.
The tests assert that property separately from the reporting one.

### Neighbours, checked rather than assumed

- **`log1p`** is wired even though no finite input overflows, because `log` and
  `log10` already route their pole through `check_overflow` — house convention,
  not a dead gate. **NaN is unaffected**: `check_overflow` tests only
  `= infinity` / `= neg_infinity`, and NaN compares false to both, so
  `log1p x < -1.` still returns NaN exactly as C's `log1pf` does. Pinned by a
  test, because propagating `log`'s own NaN→−inf collapse into `log1p` would have
  been a **new** oracle divergence — the opposite of the point.
- **`hypot`** already had it; a 1.33e8-pair boundary sweep confirms it is
  genuinely reachable (first at `hypot(6.11e37, 3.35e38)`), 0 spurious, 0 missed.
- **`copysign`, `fmod`, `abs_float`** cannot overflow (magnitude-preserving,
  `|r| < |divisor|`, `|x| ≤ max`); **`minus`** is `sub`, already wired.

### Minor: `expm1`/`log1p` missing from `fallback_names`

Correct, and not cosmetic — they are declared `dev "expm1f" "expm1"` /
`dev "log1pf" "log1p"`, so the discarded framework was observable on them too.
Verified they **extend** the gate rather than pad it: under the pre-fix
`"generic"` framework, clang rejects the emitted kernel with
`use of undeclared identifier 'expm1f'`. Easy to overlook because Metal polyfills
both in its `pre_hook`; OpenCL has none and took the CUDA spelling straight
through.

### Red before green

```
[FAIL] Exception: expm1 100.0 raises Float32_overflow
[FAIL] Exception: log1p (-1.0) raises Float32_overflow
Tests: 60, Passed: 58, Failed: 2
```

The **value** assertions stayed green throughout — which is what proves the change
is reporting-only. Every new assertion has a negative control (`expm1 1.0` must
*not* raise, `log1p -2.0` must *not* raise, `hypot 1.0 1.0` must *not* raise), so
none can pass by raising unconditionally.

Sweep toolchain: clang 21 / glibc on x86-64 (Ryzen 9 7950X). It models the OCaml
implementation against the host libm; **it is not evidence about any GPU's
`expm1`**, and no GPU was involved.

---

## Rebase note — after a rebase onto a base that touched your gate's dependencies, green is not evidence

Rebased onto `43653248` (which carries #316 and #319). The only conflict was
`sarek/tests/unit/dune`, purely additive on both sides with an empty common
ancestor — #319 appended its `test_sync_stmt_emission` stanza where this branch
appends `test_intrinsic_surface`. Both kept.

Worth stating because it is a failure mode that arrives through **someone else's
merge rather than your own edit**: #316 rewrote
`sarek/tests/codegen_golden/opencl_gate/opencl_clang.ml`, which is the exact file
this PR's clang gate calls into (`available ()` / `run_clang`). Had that changed
behaviour, the gate would have degraded to `Alcotest.skip ()` — which still
prints `[OK]` in the Alcotest summary. A passing run would then have meant
nothing, and no edit of mine would have caused it.

So the gate was re-proven by mutation on the rebased base rather than trusted:
reverting the OpenCL backend to the pre-fix `"generic"` framework still yields

```
FAIL emitted OpenCL for Float32.abs_float does not compile:
/tmp/sarek_gate_ocl_9bb078.cl:3:12: error: use of undeclared identifier 'fabsf'
```

with exit 1 and no `SKIP` in the output. Tree restored and verified clean
afterwards.

Post-rebase verification was by **counting and locating**, not by exit status:
2 commits, 21 files, `merge-base HEAD origin/main == origin/main`, `HEAD` on a
branch, no `rebase-merge` directory left behind, and one distinctive marker from
each commit located **by line number** (`Sarek_pure_registry.ml:220`,
`Sarek_registry.ml:178` and `:243`, `Sarek_native_intrinsics.ml:118`;
`Sarek_float32.ml:248`/`:250`, `test_intrinsic_surface.ml:368`,
`test_float32.ml:251`). Line numbers rather than `grep -c`, because a count
cannot distinguish "the fix is present" from "a comment mentions it".

`dune build`, `@runtest`, `@fmt` (exit 0, zero diffs) and
`check-test-alias-coverage.sh` (82 executables wired) all pass on the rebased
base. `benchmarks/descriptions/generated/` regenerated — empty diff, inspected.

### Review freshness after the rebase — verified, not asserted

The `CodeRabbit` check on the rebased HEAD now reads **`success — Review completed`**
rather than the previous `Review rate limited`, and it is attached to `0257cb3a`.
But it answered a `@coderabbitai review` request in 5 seconds and its own reply
notes that it "does not re-review already reviewed commits" — so a green check
alone cannot distinguish "re-reviewed and found nothing" from "recognised the
content and skipped".

Rather than rely on that, the question was made mechanical: **is the contribution
this PR adds still byte-identical to the one that was reviewed?**

```
git diff a04e7bef...231f97e3   # the reviewed tree's contribution   -> 1573 lines
git diff 43653248...0257cb3a   # the rebased tree's contribution    -> 1573 lines
```

Comparing only the added/removed lines of the two patches, with `+++`/`---`
headers excluded:

```
reviewed +/- lines: 1305
current  +/- lines: 1305
PM_DIFF_EXIT=0   (every added/removed line is identical)
file lists identical
```

The two patches differ **only** in blob index hashes, hunk line offsets
(`@@ -334` → `@@ -330`, `@@ -130` → `@@ -147`) and one hunk's *context* lines in
`sarek/tests/unit/dune`, where the stanza above mine is now #319's
`test_sync_stmt_emission` instead of `test_opencl_gate`. Not one added or removed
line changed.

So the carried-over review covers the current diff exactly. The rebase moved this
contribution, it did not alter it.

### DIFF
--- a/sarek/codegen/Sarek_ir_cuda.ml
+++ b/sarek/codegen/Sarek_ir_cuda.ml
@@ -284,7 +284,10 @@ and cuda_backend =
     pre_hook = (fun _ ~full_name:_ _ _ _ -> false);
     post_hook =
       (fun buf path name args ->
-        Dispatch.emit_registry_template ~gen_expr buf path name args);
+        (* Same framework tag the pure-registry lookup uses: without it this
+           fallback emitted the CUDA spelling on every backend. *)
+        let framework = Option.value ~default:"CUDA" !current_framework in
+        Dispatch.emit_registry_template ~gen_expr ~framework buf path name args);
     on_unknown =
       (fun full ->
         Codegen_error.raise_error (Codegen_error.unknown_intrinsic full));
--- a/sarek/codegen/Sarek_ir_intrinsic_dispatch.ml
+++ b/sarek/codegen/Sarek_ir_intrinsic_dispatch.ml
@@ -133,8 +133,10 @@ let count_placeholders s =
 (** Expand a Metal/CUDA/OpenCL FFI-registry ([Sarek_registry]) device template
     for [path.name]. Returns [false] when the registry has no template (caller
     then raises the unknown-intrinsic error); [true] once emitted. *)
-let emit_registry_template ~gen_expr buf path name args =
-  match Sarek_registry.fun_device_template ~module_path:path name with
+let emit_registry_template ~gen_expr ~framework buf path name args =
+  match
+    Sarek_registry.fun_device_template ~module_path:path ~framework name
+  with
   | None -> false
   | Some template ->
       let arg_strs =
--- a/sarek/codegen/Sarek_ir_metal.ml
+++ b/sarek/codegen/Sarek_ir_metal.ml
@@ -340,7 +340,10 @@ and metal_backend =
         else false);
     post_hook =
       (fun buf path name args ->
-        Dispatch.emit_registry_template ~gen_expr buf path name args);
+        (* Same framework tag the pure-registry lookup uses: without it this
+           fallback emitted the CUDA spelling on every backend. *)
+        let framework = Option.value ~default:"Metal" !current_framework in
+        Dispatch.emit_registry_template ~gen_expr ~framework buf path name args);
     on_unknown =
       (fun full ->
         Codegen_error.raise_error (Codegen_error.unknown_intrinsic full));
--- a/sarek/codegen/Sarek_ir_opencl.ml
+++ b/sarek/codegen/Sarek_ir_opencl.ml
@@ -330,7 +330,10 @@ and opencl_backend =
     pre_hook = (fun _ ~full_name:_ _ _ _ -> false);
     post_hook =
       (fun buf path name args ->
-        Dispatch.emit_registry_template ~gen_expr buf path name args);
+        (* Same framework tag the pure-registry lookup uses: without it this
+           fallback emitted the CUDA spelling on every backend. *)
+        let framework = Option.value ~default:"OpenCL" !current_framework in
+        Dispatch.emit_registry_template ~gen_expr ~framework buf path name args);
     on_unknown =
       (fun full ->
         Codegen_error.raise_error (Codegen_error.unknown_intrinsic full));
--- a/sarek/interp/Sarek_float32.ml
+++ b/sarek/interp/Sarek_float32.ml
@@ -197,3 +197,49 @@ let to_int x = int_of_float x
 
 (** Pretty print with float32-appropriate precision *)
 let to_string x = Printf.sprintf "%.7g" x
+
+(******************************************************************************
+ * Sarek stdlib surface mirror
+ *
+ * The native (cpu_kern) backend lowers an [IntrinsicRef (["Float32"], name)] to
+ * the OCaml identifier [Sarek.Sarek_cpu_runtime.Float32.<name>] with the name
+ * copied VERBATIM (sarek/ppx/Sarek_native_intrinsics.ml: map_stdlib_path, then
+ * Sarek_native_helpers.evar_qualified). So every name declared by
+ * sarek/Sarek_stdlib/Float32.ml must exist here under exactly that spelling,
+ * or the kernel does not compile — the user sees "Unbound value
+ * Sarek.Sarek_cpu_runtime.Float32.<name>" pointing into PPX-generated code.
+ *
+ * The seven names below were declared in the stdlib but missing here, so
+ * Float32.abs_float / expm1 / log1p / hypot / copysign / fmod / minus were
+ * unusable from any native kernel. sarek/tests/unit/test_intrinsic_surface.ml
+ * reconciles the two surfaces so this cannot silently recur.
+ *
+ * The four *_float32 aliases mirror the stdlib's explicit arithmetic
+ * intrinsics. They are reachable only when written as qualified calls
+ * (Float32.add_float32 a b); infix `+.` is lowered structurally by the PPX and
+ * never reaches this module.
+ ******************************************************************************)
+
+let abs_float x = abs x
+
+let minus x y = sub x y
+
+let expm1 x = to_float32 (Stdlib.expm1 x)
+
+let log1p x = to_float32 (Stdlib.log1p x)
+
+let hypot x y = to_float32 (Stdlib.hypot x y) |> check_overflow "hypot"
+
+let copysign x y = to_float32 (Stdlib.copysign x y)
+
+(* C [fmod] semantics: sign of the dividend, magnitude < |divisor|.
+   [Float.rem] is OCaml's [fmod]; the stdlib intrinsic uses the same. *)
+let fmod x y = to_float32 (Float.rem x y)
+
+let add_float32 x y = add x y
+
+let sub_float32 x y = sub x y
+
+let mul_float32 x y = mul x y
+
+let div_float32 x y = div x y
--- a/sarek/interp/Sarek_float32.ml
+++ b/sarek/interp/Sarek_float32.ml
@@ -224,9 +224,30 @@ let abs_float x = abs x
 
 let minus x y = sub x y
 
-let expm1 x = to_float32 (Stdlib.expm1 x)
+(* check_overflow is value-neutral in the default [Silent] mode -- it returns
+   its argument untouched -- so these three report an infinite result in [Warn]
+   and [Exception] mode without changing a single number the interpreter
+   produces as the cross-backend oracle.
 
-let log1p x = to_float32 (Stdlib.log1p x)
+   Reachability was settled by an exhaustive sweep of all 2^32 binary32 inputs
+   against C's expm1f/log1pf (clang 21, glibc, x86-64), replicating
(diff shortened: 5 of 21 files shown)


mathiasbourgoin:
@coderabbitai review

PR #340, opened 2026-07-27
chore(ci): measure whether pocl works on a bare runner (backlog-99)

mathiasbourgoin:
Every OpenCL claim this project makes rests on one runtime: the Intel oneAPI CPU ICD inside `ci/Dockerfile`. Nobody has measured what a **bare** GitHub runner does, so there is no second, independent OpenCL execution environment — and no way to tell an oneAPI-specific result from a real one.

backlog-79 tried pocl once, in the image, and pinned e2e to it. pocl 1.8 + LLVM 11 on jammy compiled no kernel at all (`unknown target CPU 'generic'`), the float32 math tests SKIPPED, and CI went green having validated nothing. `ci/assert-toolchain.sh` records that outcome and says the retry belongs in:

> a separate experimental PR that installs it unpinned and reports whether it can compile on the real GitHub runner. When that lands, the checks belong here.

This is that PR. The measured failure was a property of *jammy's* pocl; `ubuntu-latest` ships one several major versions newer, and that is the unmeasured part.

## The point is information, not a gate

The new `pocl-runner-probe` job is `continue-on-error`, installs pocl **unpinned** (whatever the runner ships is the thing being measured), and the probe exits 0 on its own regardless of verdict. **No test is redirected to pocl and nothing consults the result.**

If pocl cannot compile a kernel there, that is a perfectly good result: it means the project's OpenCL coverage still rests entirely on the container image, and a bare runner cannot serve as a second execution environment. Promoting this to a gate is a separate decision that needs a green here first.

## Five stages, reported separately

Lumping them together is how a probe says "OpenCL is broken" when the truth was "no ICD file was installed".

| | stage | |
| --- | --- | --- |
| A | `icd` | an ICD manifest exists for pocl |
| B | `enumerate` | a pocl device is returned |
| C | `compile` | `clBuildProgram` accepts a trivial kernel — **where backlog-79 died** |
| D | `execute` | the kernel RUNS and the arithmetic is checked element-wise |
| E | `reject` | an INVALID kernel is refused |

**D and E are what make a green trustworthy.** A build that succeeded is not a kernel that ran — the same reason this repo distrusts a reported SKIP. And a compiler that accepts nonsense reports success without compiling: E is the positive control for C. The probe matches the pocl platform *by name*, so another ICD present on the runner cannot report success on pocl's behalf.

Verdict is one of `POCL-WORKS` / `POCL-UNTRUSTWORTHY` / `POCL-ENUMERATES-ONLY` / `POCL-ABSENT`, written to the job summary. An `unknown` stage is bucketed with failure, never with pass.

## Verified locally before shipping

No pocl is installed on the dev machine, so the probe's own machinery was validated against the host's real ICD:

- **against rusticl (RX 7900 XTX) with the platform match relaxed**: all five stages pass, kernel returns **1024/1024 correct elements** — so C, D and E are not inert.
- **mutating the kernel to compute the wrong value**: stage D goes red (1023/1024 wrong) — the arithmetic check is real, not decorative.
- **unmodified, with only `rusticl` and `nvidia` ICDs present**: reports `POCL-ABSENT` rather than crediting a non-pocl device.

## Proving it can fail

`ci/pocl-runner-probe.test.sh` — 12 cases, wired into the harness self-test block. It covers the verdict classifier hermetically via a stub compiler, so it runs anywhere and exercises outcomes this host cannot produce, notably a pocl that compiles invalid source.

| mutation | result |
| --- | --- |
| drop `execute` from the verdict | 2 FAIL |
| treat an unbuilt probe as passed | 2 FAIL |

The probe is informational and exits 0 by design, which is the shape that rots unnoticed — nothing reads its exit code, so a probe that stopped measuring would keep printing a verdict indistinguishable from a real one. The covering test is what stops that.

## The result

```
distro:   Ubuntu 24.04.4 LTS
package:  pocl-opencl-icd 5.0-2.1build3
device:   cpu-haswell-AMD EPYC 9V74 80-Core Processor
version:  OpenCL 3.0 PoCL HSTR: cpu-x86_64-pc-linux-gnu-haswell

icd / enumerate / compile / execute / reject : all pass
POCL_PROBE_VERDICT=POCL-WORKS
```

**Yes** — a bare runner compiles and runs an OpenCL kernel, 1024/1024 elements correct, and rejects invalid source.

### This reframes backlog-79, it does not contradict it

Worth stating plainly, because the old conclusion is cited elsewhere in the tree (`ci/Dockerfile`, `ci/assert-toolchain.sh:345-354`) and reads as a fact about pocl:

| | backlog-79 | here |
| --- | --- | --- |
| where | inside the CI image | bare runner |
| distro | Ubuntu 22.04 (jammy) | Ubuntu 24.04 (noble) |
| pocl | 1.8 | 5.0 |
| LLVM | 11 | current |
| result | compiles nothing (`unknown target CPU 'generic'`) | compiles, runs, rejects |

**"pocl cannot compile a kernel" was never a fact about pocl** — it was a fact about pocl 1.8 against LLVM 11 on jammy. Both measurements are correct; they are measurements of different things. This is exactly why it was worth measuring somewhere else rather than inferring from the first result.

Recorded in the probe header and the `ci.yml` job comment (second commit) so the next reader gets both halves.

## Note: the CI gate that could not see this PR (now fixed upstream)

An earlier revision of this PR carried a note claiming the marker `review-gate-hardening` was present in the pristine upstream file. **That was wrong**, and the correction matters. All four declared markers occur **zero** times in the genuinely pristine file. The check had become *unsatisfiable*: it resolved "pristine upstream" as `git merge-base HEAD origin/main`, which stopped meaning pristine the moment the patch landed on `main`, so it compared the file to itself — in direct contradiction with its neighbour at `:606`, which requires the same bytes to *contain* the markers. `review-gate-hardening` was named only because it is first in the array. No marker string could have fixed it.

Separately, `actions/checkout@v4` defaults to `fetch-depth: 1`, and a `pull_request` checkout is a detached HEAD on `refs/pull/N/merge` with no `refs/remotes/origin/main`, so `merge-base` exited 128, `base` became null, the loop iterated nothing, and the check printed its specificity note and exited 0 — **50 passed / 0 failed on PRs versus 49 / 1 on push.**

Both are fixed and merged in PR #341; the resolver now anchors on the manifest and fails closed on a shallow checkout. This PR is rebased onto `4f5335e0`.

**Why the rebase was necessary rather than cosmetic:** the harness self-test block runs under `set -e` and aborted at position 3, so every check after it — including this PR's covering test — was never reached in CI. The green this PR showed before the rebase did not include its own test. It does now.

## Merge is blocked by token scope, not by this PR

`gh pr merge` refuses: *refusing to allow an OAuth App to create or update workflow `.github/workflows/ci.yml` without `workflow` scope*. The available token carries `gist, read:org, repo` only. This affects the **merge step alone** — the branch, its content and its CI are unaffected (pushes go over SSH, not the API). Any PR touching `.github/workflows/` needs either a widened token or a merge by hand; escalated to Mathias.

Rebased onto `939783ad`. `#339`'s `ledger-schema.test.js` line is now in `main`; it and this PR's `ci/pocl-runner-probe.test.sh` line merged cleanly without a conflict — they sit at opposite ends of the harness block. Both verified present and ordered after the rebase.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -370,6 +370,14 @@ jobs:
   # is a good result -- "a bare runner cannot run an OpenCL kernel either" is
   # information the project does not have today. Promoting this to a gate is a
   # separate decision that needs a green here first.
+  #
+  # MEASURED 2026-07-27 (run 30282252291): VERDICT=POCL-WORKS. Ubuntu 24.04.4,
+  # pocl 5.0, cpu-haswell device; compiled, ran 1024/1024 elements correctly,
+  # and rejected invalid source. backlog-79's "pocl compiles nothing" was pocl
+  # 1.8 + LLVM 11 on jammy inside the image, not a fact about pocl. See the
+  # header of ci/pocl-runner-probe.sh for what this does and does not license.
+  # Still not promoted to a gate: one green is not a stability record, and this
+  # installs an unpinned apt package.
   pocl-runner-probe:
     runs-on: ubuntu-latest
     timeout-minutes: 15
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,6 +144,15 @@ jobs:
           # is the positive control without which "went red" and "is always
           # red" would be the same observation.
           ./scripts/check-license-headers.test.sh
+          # backlog-99. The pocl probe is informational and exits 0 by design,
+          # which is the shape that rots unnoticed: nothing reads its exit
+          # code, so a probe that stopped measuring would keep printing a
+          # verdict indistinguishable from a real one. This asserts the verdict
+          # tracks the stages, and that no combination of failures -- including
+          # a probe that never built -- can report POCL-WORKS. It is hermetic
+          # (a stub compiler supplies the stage results) and needs no pocl, so
+          # it runs here rather than in the probe job.
+          ./ci/pocl-runner-probe.test.sh
 
       - name: Formal proofs admit gate
         run: |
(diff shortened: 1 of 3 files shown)


mathiasbourgoin:
Result is in, from run [30282252291](https://github.com/mathiasbourgoin/Sarek/actions/runs/30282252291) on a real `ubuntu-latest` runner:

```
distro:   Ubuntu 24.04.4 LTS
package:  pocl-opencl-icd 5.0-2.1build3
ICD:      pocl.icd -> libpocl.so.2.12.0
device:   cpu-haswell-AMD EPYC 9V74 80-Core Processor
version:  OpenCL 3.0 PoCL HSTR: cpu-x86_64-pc-linux-gnu-haswell

icd        pass
enumerate  pass
compile    pass
execute    pass
reject     pass

POCL_PROBE_VERDICT=POCL-WORKS
```

**Yes** — a bare runner compiles and runs an OpenCL kernel, 1024/1024 elements correct, and rejects invalid source (so the `compile` pass is not a compiler saying yes to everything).

This does not contradict backlog-79, it **localises** it. That failure was pocl 1.8 + LLVM 11 on jammy, inside the image; the runner is noble with pocl 5.0, four major versions on. "pocl cannot compile" was never a fact about pocl, only about that pairing.

**What it implies for the OpenCL gates that exist today:** a second, independent OpenCL execution environment is now known to be available, so the project's OpenCL claims no longer *have* to rest solely on the image's Intel oneAPI ICD. Two follow-ups, neither done here:

- `ci/assert-toolchain.sh`'s "NOT asserted here: the OpenCL ICD inventory … when that lands, the checks belong here" note can now be acted on.
- A **differential** run — the same kernels under oneAPI *and* pocl — would catch the class of bug a single ICD cannot: generated OpenCL that one runtime accepts and another rejects. That is the real prize, and a separate change with a real cost.

Still **not** promoted to a gate: one green is not a stability record, and this installs an unpinned apt package, which is how a green turns red on someone else's release schedule.

The observation is now recorded in `ci/pocl-runner-probe.sh`'s header and the job comment in `ci.yml`.

PR #344, opened 2026-07-27
feat(vulkan): cooperative-matrix capability query and fragment type (backlog-62 slices 2, 4b)

mathiasbourgoin:
Slice 2 of `docs/design/f16-relaxed-accuracy.md` §7, plus the fragment type that §7 slice 4b makes binding. **No refusal is lifted, and nothing in Sarek emits a cooperative-matrix instruction.** This is the query, the enablement, and the vocabulary.

## What lands

**`Sarek_coopmat`** (`spoc/ir`, beside `Sarek_capability`) — the backend-neutral half: component types, scope, shape, `config`, `device_support`, and the fragment type with its subgroup calling convention.

- **Integer component types are admitted from the start**, per §8 adopted as binding by slice 4b. 12 of the 14 configurations the local RX 7900 XTX advertises are integer, and `SPV_KHR_cooperative_matrix` makes integer accumulation exact at the result precision, so those land under Sarek's **existing strict contract** — no relaxation, no allowlist, no opt-in. `accumulation_is_exact` *computes* that split (from `cfg_c`/`cfg_result`, not the operand types) rather than asserting it in prose.
- **Dimensions are a record, not a closed variant**, so Vulkan's measured 16×16×16 and Metal's measured 8×8×8 both fit.
- `components_per_invocation` encodes the subgroup ABI as a `result`: a fragment's components must divide the subgroup size, and that is a check codegen must do *before* emitting.

**Vulkan plumbing** (`Vulkan_types` / `Vulkan_bindings` / `Vulkan_api_device`) — `VkPhysicalDeviceFeatures2` chain for `shaderFloat16`, `storageBuffer16BitAccess` and `cooperativeMatrix`; `VkPhysicalDeviceProperties2` chain for `driverID`/`driverName`/`driverInfo` and `subgroupSize`; configuration enumeration via `vkGetInstanceProcAddr`. All three features and `VK_KHR_cooperative_matrix` are **requested** at `vkCreateDevice`, following #332: support is not enablement, and the previous code queried `shaderFloat16` nowhere and requested it never.

**`Framework_sig.capabilities`** gains `coopmat : Sarek_coopmat.device_support option`. An *option*, not a list — an empty list and an unprobed device are different facts, and `Sarek_coopmat.verdict` maps `None` to `Unknown`, which does not permit.

## The refusal, observed

Executed on **AMD Radeon RX 7900 XTX (RADV NAVI31)** and **AMD Ryzen 9 7950X iGPU (RADV RAPHAEL_MENDOCINO)**, radv, Mesa 26.1.4-arch3.1, Vulkan 1.4.354.

```
device 0: AMD Radeon RX 7900 XTX (RADV NAVI31)  [driverID=3 radv / Mesa 26.1.4-arch3.1]
          subgroupSize=64 shaderFloat16=true storageBuffer16=true coopmat_configs=14 robust=true
device 1: AMD Ryzen 9 7950X (RADV RAPHAEL_MENDOCINO) [driverID=3 radv / Mesa 26.1.4-arch3.1]
          subgroupSize=64 shaderFloat16=true storageBuffer16=true coopmat_configs=0  robust=false

REFUSED: AMD Ryzen 9 7950X 16-Core Processor (RADV RAPHAEL_MENDOCINO): cooperative-matrix
unavailable — the device advertises no cooperative-matrix configuration matching
16x16x16 f16*f16+f32->f32 subgroup [device-optional; measured: ...]

gate observed: 1 device(s) permitted, 1 refused
device 0: 14 configurations, 12 accumulate exactly (strict contract)
```

That reproduces §4's table through the Sarek path, on the same box, with one device permitting and one refusing under the same driver.

## Three things the measurement corrected

1. **`subgroupSize` is 64 on both local devices.** `Vulkan_plugin_base` hard-coded `warp_size = 32`, which was wrong for *both*. This is ABI, not a statistic: a 16×16 fragment over 64 invocations is 4 components per invocation, not 8, so codegen written against 32 would have been wrong at the calling convention. `warp_size` is now the probed value.

2. **`vkGetPhysicalDeviceCooperativeMatrixPropertiesKHR` answers for devices that do not support the extension.** It returns `VK_SUCCESS` and **all fourteen configurations for the iGPU** — the device the design document offers as the free negative device, which does not advertise `VK_KHR_cooperative_matrix` and whose `cooperativeMatrix` feature reads false. Calling an extension entry point on such a device is UB, and RADV's UB here is a plausible, well-formed, entirely wrong answer. §4's claim is correct as written (it was measured through the extension list), but a probe that enumerated configurations first would have contradicted it and left the gate unable to refuse anything. The probe gates on extension **and** feature before querying.

3. **A wrong `VkComponentTypeKHR` enumerant table looked like a working probe.** It decoded the 14 advertised configurations into 6 plausible-looking wrong ones (`f16*s8->s32`, `u8*u8->u8`) and silently dropped the other 8. Every number looked reasonable and every test passed. Enumerants are now pinned against `vulkan_core.h` by value, and `device_support` carries `ds_advertised_count` so a dropped configuration is a failure rather than a shorter list.

## Proving the gates can fail

The device gate test alone is **tautological** — it compares the verdict against the same list the verdict reads — and this was found by mutation, not by inspection:

| mutant | `verdict tracks support` | `configurations imply enabled feature` | `no configuration dropped` | `enumerants match vulkan_core.h` |
|---|---|---|---|---|
| extension check removed (`if false && …`) | green (iGPU reports 14, both permit) | **RED** | green | green |
| one enumerant wrong (`f16 = 3`) | green | green | **RED** | **RED** |

Both mutants are caught, each by at least one check that does not read the polluted data. `coopmat_enabled` is recorded on the device rather than derived from the configuration list precisely so the first implication is expressible.

## `shaderFloat16` enablement does not move the f16 measurement

Controlled A/B, one build, feature request toggled, `test_vulkan_f16_tripwire` run on each arm:

| arm | RX 7900 XTX | Raphael iGPU |
|---|---|---|
| `shaderFloat16` **not** requested (pre-slice-2 path) | 2912/63488 | 2912/63488 |
| `shaderFloat16` requested | 2912/63488 | 2912/63488 |

Same first divergence on every arm, same `precise` figures, all three calibration controls green. So `fp-contraction-policy.md` §7(b)'s caveat named a real gap in the plumbing but **not** a confound in the measurement — ACO's absorption of the f32→f16 narrowing does not depend on whether the feature was requested. §7(b) is closed with the table; the tripwire's header comment is updated in place.

## What this does NOT do

- **Slice 3** — `Sarek_ir_glsl.reject_float16_kernel` still refuses f16 unconditionally. Untouched.
- **Slice 4a** — the coopmat numeric contract of §5 is still **entirely unmeasured on every implementation**. No `OpCooperativeMatrixMulAddKHR` was executed; neither of §5.4's two mandatory positive controls exists; §5.3's binary64 exactness invariant is not asserted anywhere, because no reference exists to assert it in.
- **Slice 4c** — no codegen. `Sarek_coopmat` values are queryable and none is reachable from the DSL.
- `Execute.check_device_capabilities` is **not** widened. Its gated list stays `[Float64; Int64]`, and cooperative matrices are not added because no kernel can require them yet — a gate on a requirement nothing can express would be a refusal with nothing to refuse.
- `bfloat16` is deliberately absent from `component_type`: Metal advertises an 8×8 `bfloat` `simdgroup_matrix`, but nothing has been swept about what it computes.

Vulkan's backend now reports `Float16` in `device_features` (probed and enabled, so honest); the GLSL policy refusal is unaffected, and no other backend's `device_features` changes.

## Verification

- `dune build @fmt` clean; full build clean.
- Suite: **1633 alcotest / 108 suites + 32 qcheck / 6, 0 FAIL, 23 SKIP** (baseline 1612/106 + 32/6; +21 cases, +2 suites = exactly the two new files). Log confirmed non-empty (4733 lines) before counting, per backlog-150.
- The 4 zero-case suites are pre-existing CUDA/PTX hardware gates (`ptx_stride_spike`, `ptx_atomics_probe`, `ptx_mma_probe`, `cuda_f16_sass`), none introduced here.
- Rebased on current `origin/main` (`1f174523`), rebuilt and re-tested after the rebase.

## Summary by CodeRabbit

* **New Features**
  * Added cooperative-matrix capability modeling and device-level availability checks.
  * Vulkan now reports supported cooperative-matrix configurations, subgroup size, driver details, and relevant numeric feature support.
  * Capability checks now distinguish supported, unavailable, and unprobed hardware.
* **Bug Fixes**
  * Corrected Vulkan configuration decoding and detect dropped advertised configurations.
* **Tests**
  * Added coverage for cooperative-matrix support, configuration matching, subgroup compatibility, and Vulkan capability reporting.
* **Documentation**
  * Updated accuracy and feature-plumbing documentation with latest measurements and clarifications.

### DIFF
--- a/docs/design/f16-relaxed-accuracy.md
+++ b/docs/design/f16-relaxed-accuracy.md
@@ -876,7 +876,7 @@ refusal before slice 3.
 |---|---|---|---|---|---|
 | **0** | the contract, as a testable classifier | the gate can tell `S_fuse_mul_into_narrowing` from a dropped narrowing | none (host-only) | no | open — and now also carries §1.3's per-narrowing ceiling |
 | **1** | element-wise model characterisation of ACO scalar f16 | whether the contract of §1.2 is deliverable at all | RX 7900 XTX + Raphael iGPU (local) | no | **DONE — all 20 shapes measured (2 in slice 1, the other 18 in backlog-151). Deliverable, and §1.2's set is generated rather than enumerated. 18 shapes measured but NOT admitted.** |
-| **2** | `shaderFloat16` + driver-identity + capability plumbing | the gate can be keyed on a driver, not a device name | local, both devices | no | open |
+| **2** | `shaderFloat16` + driver-identity + capability plumbing | the gate can be keyed on a driver, not a device name | local, both devices | no | **DONE 2026-07-27** — plus the coopmat capability query and the fragment type of slice 4b |
 | **3** | GLSL scalar f16, allowlisted | Sarek-*generated* f16 shaders meet the contract | RX 7900 XTX (local) | **yes**, on an allowlist | open |
 | **4** | backlog-62 Vulkan coopmat, f16×f16→f32 | the tensor-core path, end to end | RX 7900 XTX (local) | yes (new capability) | open |
 | **5** | backlog-63 Metal — **scalar f16 first** | the Metal row of §2 | ladon (M4) | no | **DONE 2026-07-27 — strict, 0/63488** |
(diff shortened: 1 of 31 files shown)


mathiasbourgoin:
All four findings addressed in `bb4c63b6`.

**1. Promoted extensions no longer require the string** (`Vulkan_api_device.ml:674-691`, `:753`). `VK_KHR_16bit_storage` is core in 1.1 and `VK_KHR_shader_float16_int8` in 1.2, so the `advertised && supported` conjunction stopped *requesting* features a core-version device supports — #332's lesson arriving from the other side. Each promoted feature is now `feature_bit && (core_version || string)`, against the **effective** API version (`min` of the device's and the 1.2 the instance requests, so lowering the instance version cannot silently start requesting features the app is not entitled to). Only `VK_KHR_cooperative_matrix` — a real extension — keeps an unconditional string requirement. The string still decides `ppEnabledExtensionNames` membership, since requesting an unadvertised extension fails `vkCreateDevice`.

Both local devices advertise both strings, so the hardware hides this. Measured by **simulation** instead — the two promoted strings hidden from `has_ext`, i.e. an otherwise identical core-only device:

| | RX 7900 XTX | Raphael iGPU |
|---|---|---|
| old conjunction | `shaderFloat16=false storageBuffer16=false` | `shaderFloat16=false storageBuffer16=false` |
| new form | `shaderFloat16=true storageBuffer16=true` | `shaderFloat16=true storageBuffer16=true` |

`vkCreateDevice` succeeds in both, and the coopmat arm reports 14/0 unchanged across all four runs — the control, since its string requirement is correct. The derivation is in the comment at `:648-673`.

**2. Contradictory enumerant prose deleted, not corrected** (`Vulkan_types.ml:1365-1374`). It restated f16=3, f32=4, s8=7, s32=9, u8=0, u32=1 directly above constants reading 0/1/3/5/7/9. You are right that this is the file's own trap one layer up: the test pins the constants, nothing pins the prose. So there is no second copy of the numbers now — the comment points at the constants and records why it does, since a corrected restatement would just reset the same clock.

**3. `subgroup_size` clamped at the source** (`Vulkan_api_device.ml:350` `fallback_subgroup_size`, `:390-393`, `:60-73`; `Vulkan_plugin_base.ml:91-102`). `zero_struct` zeroes the properties struct, so a driver ignoring the `sType` left `subgroupSize = 0` in `warp_size` — worse than the wrong-but-usable 32, since a consumer dividing by it faults. Now `fallback_subgroup_size = 32` (the historical value, kept for continuity and nothing else) with `subgroup_size_probed` saying which of the two a caller holds. The comment names what was got on which device: **64 on both** the RX 7900 XTX (RADV NAVI31) and the Ryzen 9 7950X iGPU (RADV RAPHAEL_MENDOCINO), radv / Mesa 26.1.4-arch3.1 — so 32 is wrong for both and must never be the silent normal. `test_vulkan_coopmat_capability.ml:240-245` now asserts every local device is *probed*, not merely positive: the fallback going live is a test failure. (Positive alone would have passed while measuring nothing — the same shape as the tautological gate test.)

**4. Stale remediation step dropped** (`test_vulkan_f16_tripwire.ml:428-438`). Old step (4) told the reader to build the `shaderFloat16` plumbing this PR delivers. Removed, steps renumbered 1-5, and the message now says the plumbing exists so nobody redoes it.

Suite unchanged: **1633 alcotest / 108 suites + 32 qcheck / 6, 0 FAIL, 23 SKIP**. `dune build @fmt` clean. The A/B closing `fp-contraction-policy.md` §7(b) is already in the PR body under its own heading.

No slice-3 work started; `Execute.check_device_capabilities` still ungated for coopmat; `bfloat16` still omitted.

PR #351, opened 2026-07-28
fix(counter): refuse a count for a run that did not complete (backlog-157)

mathiasbourgoin:
## The defect, measured on this tree

`scripts/test-suite-counts.sh` printed a plausible total, `0 FAIL` and exit 0 for a `dune test` that exited 1 on a build error.

The reported instance said 839 cases / 72 suites, "roughly half the real total". The measurement I took is worse than that, and it is what decided the design:

| | cases | suites | dune exit | counter exit |
|---|---|---|---|---|
| clean `dune test --force` | 1697 | 117 | 0 | 0 |
| same tree, one unbound identifier appended to `sarek/tests/unit/test_soa.ml` | **1692** | **116** | **1** | **0** |

99.7% of the true case count and 99.1% of the true suite count, `0 FAIL`, from a build that never happened. Every gate the script already had is satisfied: the log is not empty (backlog-150, `#338`), every epilogue in it parses, and 116 suites clears any floor a real run of this repository could set.

## What I did, and what I rejected

**Rejected: an expected-suite count from the caller, or a floor derived from the true total.** The measurement above kills it. Catching 116-out-of-117 needs a tolerance tighter than one suite in a hundred, and a tolerance that tight rejects every legitimately scoped run — `dune test spoc/ir` is 6 suites, which is exactly what `--min-suites 0` exists to permit. One instrument cannot be both.

**Rejected: making `--runner-exit` mandatory.** Not on call-site cost — there are *zero* scripted call sites. `git grep -l test-suite-counts` returns six files: the script, its covering test, `prove-red.sh`, `prove-red.test.sh`, `kb/properties.md` (which records `"invocation": "manual"`), and `ci.yml`, where the only invocation is of the covering test. CI never runs the tool. The real call sites are operators and agents typing it, and they are unbounded and unreachable by a code change.

It was rejected for a structural reason instead: **the pipe form this tool's own header recommends first cannot supply it.**

```
dune test --force 2>&1 | scripts/test-suite-counts.sh
```

`PIPESTATUS` exists only *after* the pipeline the flag would have to sit inside. A flag the advertised invocation cannot pass is not a flag that can be required, and making it required would convert every existing manual invocation into a usage error with no mechanism anywhere to fix them.

**Done: both of the other two, because neither covers the other's blind spot.**

- `--runner-exit N` — the caller states the runner's status. It always knows it and previously threw it away. Non-zero is exit 4, no counts.
- An **unconditional** scan for the runner's own failure report at column 0: `Error: `, `Command exited with code N.`, `Command got signal`, `Had N errors`. Exit 4, no counts.

The scan misses a runner killed by a signal, a timeout or the OOM killer, which print nothing. `--runner-exit` misses every caller who has not been told it exists, which on the day this lands is all of them. The scan is what protects the invocations that already exist and will not be updated; the flag is what protects the case the log cannot show.

**Exit 4, not 2 or 3.** Its whole purpose is to be a *different* answer from 0 on a log gates 1–3 are content with, and the remedy differs: fix the build, do not re-invoke the counter.

False-positive risk was measured, not asserted: on the 4793-line log of the clean 1697-case run, those four patterns match **zero** lines between them. Alcotest indents its case lines and dune captures test stdout, so column 0 is dune's voice. The covering test pins this with two lines lifted verbatim from that run — `RMem (EVar unbound) -> RMemError [ok]` and `[OK] error_handling 0 Error handling.` — and requires them to still be counted.

## A bug found while writing the gate

Reading the log in Python text mode translates qcheck's bare-`\r` progress-bar redraws into newlines. The first draft located the compile error in the 4793-line sample at "line 301"; `grep -n` says 240, the difference being exactly the 61 carriage returns before it. Worse than a wrong pointer into a 4000-line log: translation *manufactures line starts*, so a `^`-anchored pattern could match text dune never wrote at column 0 — which is the entire premise of the scan. Fixed with `newline=""` / `sys.stdin.buffer`, and pinned by a covering-test case that computes the expected number with `grep -n`.

## The fixture is a real log

`scripts/prove-red-fixtures/dune-test-failed-build-log.txt` is the complete unedited output of

```
dune test spoc/ir spoc/registry spoc/framework --force -j 1
```

with one unbound identifier appended to `spoc/registry/test/test_sarek_registry.ml`. Dune exited 1; six suites ran first and their epilogues are intact. The pre-fix script read it as "46 cases across 6 suites, 0 FAIL", exit 0. Nothing about it is hand-authored, and that is the point — a written failure log is a guess at what dune prints, and this gate is a claim about what dune actually prints. Regenerating it is that one command, recorded in the spec block.

## prove-red

`test-suite-counts.sh` was already a subject, so `EXPECTED_SUBJECTS` is unchanged at 4. Coverage grew and did not shrink: **13 mutations → 15**.

```
== scripts/test-suite-counts.sh
   baseline: exit 0, says 'TOTAL    :    24 cases across   4 suites'
   [empty-stdin] red: exit 2, says 'no test-suite output recognised'
   [truncated-epilogue] red: exit 2, says 'the log is truncated or the format has drifted'
   [below-floor] red: exit 3, says 'below the plausibility floor'
   [failed-build-log] red: exit 4, says 'a log of a run that did not complete'
   [runner-exit-nonzero] red: exit 4, says 'the runner exited 1'

OK: 4 subject(s), 15 mutation(s)
```

`runner-exit-nonzero` runs against the *clean* baseline log, which passes every other gate, so nothing but the flag can be producing that red.

## RED observed

Covering test: **45 passed, 0 failed** (was 27). Three mutations of the shipped script, each restored afterwards:

**M1 — marker scan disabled** (`if runner_failure is not None:` → `if False:`), 7 failures:

```
FAIL: genuine failed-build log exits 4 — the run did not complete -- expected '4', got '0'
FAIL: genuine failed-build log prints no counts on stdout -- expected '0', got '3'
FAIL: genuine failed-build log names the runner's own failure -- expected '1', got '0'
FAIL: gate 0 fires ahead of the plausibility floor (4, not 3) -- expected '4', got '3'
FAIL: 'Command exited with code N.' is a runner failure, exit 4 -- expected '4', got '0'
FAIL: line number counts \n only, agreeing with grep -n -- expected 'at line 4', got ''
FAIL: pipe form catches the failed build too (stdin, exit 4) -- expected '4', got '0'
```

`got '0'` on the first line is backlog-157 itself, reproduced through the covering test. `got '3'` on the fourth is the floor giving the wrong answer to the same log — which is why gate 0 runs ahead of it and why a case asserts that ordering.

**M2 — `--runner-exit` gate disabled**, 3 failures, all `expected '4', got '0'`, none of them touching the marker cases.

**M3 — `newline=""` reverted**, 1 failure, exactly the `\r` bug:

```
FAIL: line number counts \n only, agreeing with grep -n -- expected 'at line 4', got 'at line 6'
```

**KB property red** — `runner_failure is not None` → `runner_failure != None`:

```
::error::kb/properties.md: 1 declaration(s) violated.
    - PROP-SUITE-COUNT-REFUSES-FAILED-RUN: scripts/test-suite-counts.sh contains 0
      occurrence(s) of 'runner_failure is not None', expected at least 1.
```

Positive controls, without which "went red" and "is always red" are the same observation: `--runner-exit 0` on a plausible log is still exit 0 with an identical total; the clean 4793-line log is still 1697/117 exit 0; the `error-shaped-output` log is exit 0 and still counted.

## Existing case moved, deliberately

The `notalog` fixture ended `Command exited with code 2.`, which is now a gate-0 marker and would route it to exit 4 while the case asserts exit 2. Both answers are defensible for that text, which is exactly why the case must not straddle them — a gate-1 case that passes because gate 0 fired first has stopped testing gate 1. The line was moved into its own gate-0 case.

## Suite

`dune test --force` on this branch: dune exit **0**, log 4793 lines (non-empty, checked), and via the tool this PR changes, invoked with `--runner-exit 0`:

```
TOTAL    :  1697 cases across 117 suites
FAIL     : 0
SKIP     : 23
zero-case suites: 4
COUNTER_EXIT=0
```

Also green: `scripts/prove-red.sh` (4 subjects / 15 mutations), `scripts/prove-red.test.sh` (19/0), `scripts/check-kb-properties.sh` (33 declarations), `scripts/check-license-headers.sh`, `shellcheck -S warning` on both changed shell files. No `*.opam` churn.

## Not verified

- The scan's marker set is what dune 3.24 prints on this machine for a compile error and a failing action. `Command got signal` and `Had N errors` are from dune's documented vocabulary and are **not** exercised against a real dune run — only against synthesised lines in the covering test. A dune upgrade that reworded its error prologue would silently narrow this gate; the CI comment now points at that risk, but nothing detects it automatically.
- The reported instance (839/72) was not reproduced directly; I reproduced the *class* on this tree and got a far tighter miss (1692/116), which is strictly the harder case for any count-based defence.
- Agents and operators invoking the tool by hand will not pass `--runner-exit` until told to. Until then they are covered by the marker scan alone, which is the weaker of the two halves.

---

# Review pass (fix-first, second commit `d1b0dc7a`)

A red-prover rebuilt this branch, wrote its own mutations and came back FIX-FIRST on four items. All four reproduced before any edit. The gate was right; three of its four patterns were wrong, in both directions, and one of its two evidence sources could be switched off by typing nothing.

Everything below is a measurement taken on this tree with dune **3.24.1**.

## 1. `^Error: ` is not how this tree's builds fail

The dev profile runs warnings as errors, and dune writes that without a colon:

```
File "spoc/registry/test/test_sarek_registry.ml", line 265, characters 4-31:
Error (warning 32 [unused-value-declaration]): unused value unused_backlog157_probe.
```

A real `dune test spoc/ir spoc/registry spoc/framework --force -j 1`, with `let unused_backlog157_probe x = x + 1` appended to `spoc/registry/test/test_sarek_registry.ml`, exits 1 — and the version of the script this PR shipped read that log as:

```
TOTAL    :    46 cases across   6 suites
FAIL     : 0
exit 0
```

That is the backlog-157 defect, surviving the backlog-157 fix, on the shape this repository fails on most often.

**And it is not fixed by `^Error\b`.** That form matches **65 lines** of a real full-suite log taken on a machine with a GPU:

| source | line at column 0 | run status |
|---|---|---|
| pocl / LLVM | `Error executing LLVM compilation action.` | fine |
| `sarek/tests/e2e/Benchmarks.ml:247` | `Error on AMD Radeon RX 7900 XTX (...): [OpenCL Runtime] Compilation failed for:` | fine — the next line **tolerates** it |

The header's premise, "column 0 is dune's own voice", is false on this tree: dune passes an e2e rule's stdout straight through, unindented. So the pattern is `^Error(?::| \()` — `scripts/test-suite-counts.sh:414`. Both directions are pinned by cases, and the measurement sits in the header next to the pattern so nobody re-widens it.

## 2. Three of the four markers matched nothing dune emits

| shipped pattern | what dune 3.24.1 actually writes | where |
|---|---|---|
| `Command exited with code N.` | `Command [17] exited with code 3:` | `--display verbose` only |
| `Command got signal ...` | `Command [4] got signal SEGV:` | `--display verbose` only |
| `Had N errors` | *nothing* — a dune 2.x form | never |

Measured across `default`, `short`, `progress` and `verbose`, and for `Had N errors` also across `dune build`, `dune build -k` and `dune build --display short -k` with two independently-broken modules. The PR's own "Not verified" section flagged these as unexercised; the answer turned out to be that they had been dead since they were written.

The three guesses are gone. The two measured wordings are in, with a real log to prove it.

Also now stated in the header, because it is a real hole rather than an oversight: **in dune's default display a runtest rule that fails at runtime prints no column-0 marker at all** — only a `File "…/dune", lines …:` header and the action's own output. The scan cannot see that case; `--runner-exit` is its only defence. `^File "` is not a marker candidate, since the same shape introduces ordinary non-fatal warnings.

## 3. `--runner-exit 0` now suppresses the scan

The two sources were being treated as equal evidence. They are not: the caller's exit status *is* the fact, and the scan infers the fact from text — item 1 is that inference firing on text dune never wrote. When both are present the fact wins. Otherwise the weaker source can veto the stronger one and the caller's only remedy is to stop passing the flag.

## 4. `--runner-exit ""` silently disabled the gate

Measured on the first commit of this branch:

```
$ scripts/test-suite-counts.sh --runner-exit "" plausible.log
TOTAL    :    60 cases across  12 suites
exit 0
```

The empty string is the parser's sentinel for *"the caller did not say"*, so a caller who said nothing landed on it and the gate vanished with no diagnostic. Same for `--runner-exit=`. This is not a contrived typo — it is what the two-line form this tool documents degrades to when `rc` is unset or misspelled:

```bash
dune test --force >log 2>&1; rc=$?
scripts/test-suite-counts.sh --runner-exit "$RC" log   # wrong name → gate gone
```

Now exit 2, with a message that says which of the two things went wrong. Given-but-empty is never a fallback — the rule this script's own empty-log refusal is built on, broken by the script itself.

## The ci.yml claim, withdrawn

The comment said this step "is also what notices if dune stops reporting its own failures the way the scan expects". It does not and cannot. The fixtures are frozen files; they keep matching the wording dune used the day they were captured, so a dune rewording leaves CI green while the scan quietly stops seeing real failures. Item 2 is that failure mode having already happened, undetected. The comment now says what the fixtures do pin (the script) and what actually catches drift (re-measuring against the installed dune).

## Coverage grows, it does not move

`scripts/prove-red.sh`, subject `test-suite-counts.sh`: **5 → 8** mutations.

```
[empty-stdin]              red: exit 2, says 'no test-suite output recognised'
[truncated-epilogue]       red: exit 2, says 'the log is truncated or the format has drifted'
[below-floor]              red: exit 3, says 'below the plausibility floor'
[failed-build-log]         red: exit 4, says 'a log of a run that did not complete'
[runner-exit-nonzero]      red: exit 4, says 'the runner exited 1'
[warning-as-error-log]     red: exit 4, says 'a log of a run that did not complete'   ← new
[verbose-command-exit-log] red: exit 4, says 'a log of a run that did not complete'   ← new
[runner-exit-empty]        red: exit 2, says 'given an empty value'                   ← new
```

Covering test: **46 → 60** cases. Two new fixtures, both real logs generated the same way as the existing one, both measured **green** against the shipped gate before the fix.

New KB property `PROP-SUITE-COUNT-MARKER-IS-DUNES-VOICE` pins the precise `Error` form against re-widening.

## Each fix shown red by reverting it

| reverted | cases that went red |
|---|---|
| marker narrowed back to `^Error: ` | 4 — `warning-as-error build log exits 4 — expected '4', got '0'` |
| marker widened to `^Error\b` | 2 — `column-0 tolerated-GPU-error output is NOT a runner failure — expected '0', got '4'` |
| scan no longer suppressed by `--runner-exit 0` | 1 — `--runner-exit 0 suppresses the marker scan — expected '0', got '4'` |
| empty value falls back to "not given" | 4 — `--runner-exit "" exits 2 — expected '2', got '0'` |
| guessed `Command exited with code N.` restored | 2 — `verbose runtime-failure log exits 4 — expected '4', got '0'` |
| `^Error(?::| \()` → `^Error\b` (KB check) | `PROP-SUITE-COUNT-MARKER-IS-DUNES-VOICE: contains 0 occurrence(s), expected at least 1` |

Each suppression case carries its own control: `--runner-exit 0 suppresses the scan` is worthless without `the same log without the flag is still exit 4`, and the tolerated-GPU-error case is worthless without `the same log plus a real dune Error exits 4`. Both controls are in the file and both were checked.

## Suite, re-measured after the fix

`dune test --force -j 8` on `d1b0dc7a`: **dune exit 0**, log 4794 lines. Counted with the tool this PR changes:

```
alcotest :  1665 cases across 111 suites
qcheck   :    32 cases across   6 suites
TOTAL    :  1697 cases across 117 suites
FAIL     : 0
SKIP     : 23
zero-case suites: 4
COUNTER_EXIT=0        (--runner-exit 0)
COUNTER_EXIT=0        (no flag — the widened scan finds nothing in 4794 real lines)
```

Also green: `scripts/prove-red.sh` (4 subjects / 18 mutations), `scripts/test-suite-counts.test.sh` (60/0), `scripts/check-kb-properties.sh` (34 declarations), `scripts/check-license-headers.sh`, `dune build @fmt`. No `*.opam` churn.

## Still not verified (as of `d1b0dc7a`)

- The verbose-mode `Command [N] …` markers only fire under `--display verbose`, which nothing in this repository's CI uses. They are correct-by-measurement and currently unreachable in practice; `--runner-exit` is what covers runtime rule failures in the default display.
- ~~The `Command [N] got signal SIG:` form is measured against a five-line probe project, not against this tree.~~ **Closed below** — it is now pinned by a real SPOC log.
- A dune upgrade still silently narrows the scan, and nothing detects it automatically. That is now stated in `ci.yml` rather than claimed away.

---

# Round 3 — close-out (`54b29ad8`)

Bounded on purpose: one blocking item and one free minor, no new hunting.

## The finding

The `Command […]` half of the marker set was guarded by **nothing**. Everything about it was pinned by exactly one fixture line — `Command [17] exited with code 3:` in `scripts/prove-red-fixtures/dune-verbose-command-exit-log.txt` — so every degree of freedom in the regex collapsed to that line's literals invisibly.

Reproduced on `d1b0dc7a` before touching anything. Mutating the regex (`scripts/test-suite-counts.sh:421` on `d1b0dc7a`, `:447` after this commit) and running all three gates:

| mutation of the regex | covering test | `prove-red.sh` | `check-kb-properties.sh` |
|---|---|---|---|
| `\[\d+\]` → `\[17\]` | **green** | **green** | **green** |
| `code \d+` → `code 3` | **green** | **green** | **green** |
| `got signal \S+` → `got signal ZZZNONSENSE` | **green** | **green** | **green** |
| all three at once | **green** | **green** | **green** |

The `got signal` string occurred nowhere in the repository except the script's own header at `:152`, `:164`, `:421` — no fixture, no covering-test case, no KB anchor — while `:164` cited it as measured. Re-measured: the code is right. The pattern was correct and the check was absent, which is the same state the previous *wrong* guess (`Command exited with code N.`) had been in — asserted in a header, exercised by nothing.

Contrast the `Error` half, which already had two behavioural cases, a false-positive case with its positive control, and a KB literal anchor. **The asymmetry was the finding.**

## (a) A real SEGV fixture

`scripts/prove-red-fixtures/dune-verbose-signal-log.txt`, 333 lines, captured on this tree, not hand-written:

```
dune test spoc/ir spoc/registry spoc/framework --force -j 1 --display verbose
```

with the wild read appended to `spoc/registry/test/test_sarek_registry.ml`:

```ocaml
let () =
  let a : int array = Obj.magic "probe" in
  print_int (Array.unsafe_get a 1_000_000_000)
```

Dune exit 1; `Command [17] got signal SEGV:` at column 0, line 313. Its counts are the same plausible-looking **46 cases across 6 suites, 0 FAIL** the other three fixtures carry, so it exercises the defect identically.

One measurement worth recording: the read has to be **observed**. `ignore (Array.unsafe_get …)` is dead code, gets eliminated, and that run exits **0** — the first attempt produced a green log and no marker at all.

Added with it:

- `scripts/test-suite-counts.test.sh:336-371` — exits 4, prints no counts, plus the fixture-integrity check `grep -cE '^Command \[[0-9]+\] got signal [A-Z]+:'` mirroring `:327`, plus the plausible-total positive control.
- `scripts/test-suite-counts.sh:680-684` — prove-red mutation `verbose-signal-log`, `expect-exit: 4`.
- `scripts/test-suite-counts.sh:631` — the `copy:` line for it.

## (b) The two `\d+` fields, asserted behaviourally

A fixture pins exactly one job number and one exit code, so no fixture-driven case can ever see those fields narrow. Synthetic logs whose values are deliberately *not* the fixtures' — `scripts/test-suite-counts.test.sh:373-414`:

- `Command [4] exited with code 127:` — job ≠ 17, code ≠ 3 — must still exit 4
- `Command [259] got signal ABRT:` — job ≠ 17, signal ≠ SEGV — must still exit 4
- and the false-positive control the Command half never had, the counterpart of the tolerated-GPU-error case: an **indented** `Command [7] exited with code 9:` is a test's own output, not dune's, and is still counted (exit 0, `TOTAL : 61 cases across 1 suites`).

## (c) A KB identity anchor

`kb/properties.md:91` — `PROP-SUITE-COUNT-COMMAND-MARKER-IS-DUNES-VOICE`, `grep-present` on the regex line, mirroring `PROP-SUITE-COUNT-MARKER-IS-DUNES-VOICE` at `:90`. Both halves now get the same guard. 34 → 35 declarations.

## Minor (free): the limit of `--runner-exit 0`'s authority

`--runner-exit 0` outranks the marker **scan** and nothing else — the floor asks a different question (does the log cover the tree?) and an exit status has nothing to say about it. No case paired the flag with a below-floor log, so extending its suppression from gate 0 to gate 3 was invisible. `scripts/test-suite-counts.test.sh:524-533` pins exit 3.

## RED observed, per named mutation

Every mutation re-run against the fixed tree. Restores are from a saved copy, not `git checkout`.

| mutation | gate | observed |
|---|---|---|
| `\[\d+\]` → `\[17\]` | covering test, **exit 1** | `FAIL: a job number and exit code other than the fixture's [17]/3 still exit 4 -- expected '4', got '0'`<br>`FAIL: a job number and signal other than the fixture's [17]/SEGV still exit 4 -- expected '4', got '0'` |
| | KB, **exit 1** | `PROP-SUITE-COUNT-COMMAND-MARKER-IS-DUNES-VOICE: scripts/test-suite-counts.sh contains 0 occurrence(s) of 'r"|Command \\[\\d+\\] (?:exited with code \\d+|got signal \\S+):.*"', expected at least 1.` |
| `code \d+` → `code 3` | covering test, **exit 1** | `FAIL: a job number and exit code other than the fixture's [17]/3 still exit 4 -- expected '4', got '0'` |
| | KB, **exit 1** | same violation as above |
| `got signal \S+` → `got signal ZZZNONSENSE` | covering test, **exit 1** | `FAIL: verbose signal-killed log exits 4 -- expected '4', got '0'`<br>`FAIL: verbose signal-killed log prints no counts on stdout -- expected '0', got '3'`<br>`FAIL: a job number and signal other than the fixture's [17]/SEGV still exit 4 -- expected '4', got '0'` |
| | `prove-red.sh`, **exit 1** | `[verbose-signal-log] FAIL -- the subject DID NOT FAIL: it exited 0, exactly as it does unmutated` |
| | KB, **exit 1** | same violation as above |
| all three at once | covering test, **exit 1** | all four FAIL lines above |
| | `prove-red.sh`, **exit 1** | `[verbose-signal-log] FAIL -- the subject DID NOT FAIL` |
| | KB, **exit 1** | same violation as above |
| `if suites < min_suites:` → `… and runner_exit != "0":` | covering test, **exit 1** | `FAIL: --runner-exit 0 does NOT suppress the plausibility floor (3, not 0) -- expected '3', got '0'` |

`prove-red` stays green under the `\[17\]` / `code 3` narrowings — its fixtures *are* 17 and 3. That is exactly why (b) had to be behavioural and in the covering test.

## Suite, re-measured on `54b29ad8`

`dune test --force`, runner exit captured with `rc=$?`, not from a pipe:

```
DUNE_RC=0
log: 4794 lines (non-empty)

alcotest :  1665 cases across 111 suites
qcheck   :    32 cases across   6 suites
TOTAL    :  1697 cases across 117 suites
FAIL     : 0
SKIP     : 23
zero-case suites: 4
COUNTER_EXIT=0     (--runner-exit 0)
```

Also green: `scripts/test-suite-counts.test.sh` **69/0** (was 60/0 at round 2's writing, 63 on `d1b0dc7a`), `scripts/prove-red.sh` **4 subjects / 19 mutations** (was 18), `scripts/check-kb-properties.sh` **35 declarations** (was 34), `scripts/check-license-headers.sh`. No `*.opam` churn.

## Residual, not fixed — filed, not hunted

- The default-display hole is unchanged and unclosed: a runtest rule that fails at runtime in dune 3.24.1's default display prints no column-0 marker at all, so `--runner-exit` remains the only defence for it. Documented at `scripts/test-suite-counts.sh:194-200`.
- The `--display verbose` markers still fire only under a display mode nothing in this repository's CI uses. Both fixtures are now real, but the code path is exercised by fixtures rather than by any live invocation.
- A dune upgrade still silently narrows the whole scan and nothing detects it automatically.

## Summary by CodeRabbit

* **Bug Fixes**
  * Test-suite counts are no longer printed when the `dune test` run fails or doesn’t complete.
  * Improved detection of runner-level failures, including build errors, verbose command-exit failures, and signal termination.
  * Preserved log formatting for more reliable failure analysis.
* **Documentation**
  * Updated help text and guidance for reporting runner exit status, including new behavior for incomplete runs.
* **Tests**
  * Extended coverage with additional real log fixtures and validation for the “no completion” exit path.


mathiasbourgoin:
Closing — content **fully salvaged**, which is the opposite of how this looked when I first triaged it (backlog-166).

This PR was not a stale duplicate of the merged #357. Its four captured `dune` logs exposed **two independent live false greens in merged `main`**, both now fixed:

| fixture | outcome |
|---|---|
| `dune-test-warning-as-error-log.txt` | exposed `Error (warning N [...]):` having no colon after `Error`, so `^Error:` never matched it → **#365**, merged |
| `dune-verbose-command-exit-log.txt` | exposed `Command [N] exited with code M:` carrying no `Error` label at all → **#367**, merged |
| `dune-verbose-signal-log.txt` | same for `Command [N] got signal SEGV:` — the case backlog-53/80 actually hit → **#367**, merged |
| `dune-test-failed-build-log.txt` | **genuinely redundant.** Checked rather than assumed: it contains only the plain `Error: Unbound value …` spelling, and the current counter returns exit 5 on it. Main covers that mode with an inline heredoc of the same real spelling, and the `build-failed` prove-red mutation appends exactly that form. |

Three of the four fixtures are now in `main`; the fourth adds no spelling the gate does not already catch.

The branch itself stays unmergeable (behind `main`, and its `test-suite-counts.sh` predates #357's GATE 0 reordering), so there is nothing further to take from it.

Recorded for the record: I nearly closed this and #352 together as a pair on the strength of `git diff main...branch`, which shows branch-work-since-merge-base regardless of what `main` has independently — it answers nothing. Comparing content instead is what turned up the two defects.

PR #352, opened 2026-07-28
fix(codegen): route every public emission entry point through generate_with_types (backlog-155)

mathiasbourgoin:
## The decision

`Sarek_ir_opencl.generate_with_fp64` (`sarek/codegen/Sarek_ir_opencl.ml:1206`) called `generate` (`:1116`) rather than `generate_with_types` (`:1159`), and is re-exported at `sarek-opencl/Opencl_plugin.ml:304`. The premise holds.

**Chosen: change the signature.** `generate_with_fp64` now takes a required `~types` and delegates to `generate_with_types`. The fp64 pragma is orthogonal to type emission, so the honest composition is pragma ∘ `generate_with_types`.

**Rejected: withdraw the public re-export.** It closes one of two doors. `Sarek_ir_opencl` has no `.mli`, `sarek_codegen` is a public library, and `benchmarks/generate_backend_code.ml:481` already reaches `Sarek_opencl.Sarek_ir_opencl.*` directly — so hiding the plugin alias would leave the broken function reachable under its original name. That is how the delegation asymmetry survived backlog-142.

`~types` is **required**, not `?types` defaulting to `[]`. An optional label would reproduce the exact silent drop at every call site that forgot it.

The OpenCL backend's own `generate_source` (`Opencl_plugin.ml:190`) was a second hand-kept copy of "typedefs + fp64 pragma"; it now calls `generate_with_fp64`, so the runtime path and the public re-export are one function.

## The gap is broader, and the reported instance is not the worst one

`generate_with_fp64` is the only *third* entry point in `sarek/codegen/` — every other backend has exactly `generate` + `generate_with_types`. So the three-entry-point **shape** is OpenCL-specific. The underlying **gap** is not: *a public surface that accepts a kernel carrying `kern_types` and calls the emitter that ignores them*. Two more instances:

**1. `Sarek_transpile.of_source` — the whole public transpiler API, all five backends.** No plugin, no device. `Sarek_ir_conv.conv_kernel` faithfully carries `kern_types`/`kern_variants` across (`sarek/transpile/Sarek_ir_conv.ml:197`) and `emit_backend` (`Sarek_transpile.ml:234-241`) threw them away. Measured before the fix on a record kernel, CUDA came out as

```c
__global__ void sarek_kern(float* __restrict__ out, int sarek_out_length) {
  int i = (threadIdx.x + blockIdx.x * blockDim.x);
  point p = (point){1.0f, 2.0f};
  out[i] = (p.x + p.y);
}
```

with no `typedef struct {…} point;` above it; OpenCL identically. `of_source_with_abi` (WGSL, `:340`) had the same call.

**2. `Metal_plugin.ml:281`** was `let generate_source = Sarek_ir_metal.generate` — a top-level alias of the typedef-less emitter under the *same name* as `Backend.generate_source` in the module above it, which does use `generate_with_types`.

Both now take `~types` and route through `generate_with_types`.

## The RED I observed

New suite `sarek/tests/unit/test_typedef_entry_points.ml` (4 cases). Every "typedef IS emitted" assertion is paired with a control asserting the typedef-less emitter does NOT emit it, so the substring check cannot pass for an unrelated reason. The transpiler cases check the *use* marker before the *declaration* marker: if the body did not mention the type, declaring it would prove nothing.

Reverted one fix at a time and re-ran:

| mutation | observed |
|---|---|
| `generate_with_fp64` back to `generate` | `FAIL opencl_generate_with_fp64 0` — *"generate_with_fp64 must declare the record it uses"*, `Expected: true / Received: false`; case 2 (pragma + typedef) also red; **control case 1 stayed green**. `2 failures!`, exit **1** |
| `emit_backend` CUDA arm back to `generate` | `FAIL transpiler 0` — *"CUDA: transpiled source must declare `point`"*, with the typedef-less CUDA source printed in the message. `1 failure!`, exit **1** |

Green after restore: `Test Successful … 4 tests run`, exit **0**.

## Suite

`scripts/test-suite-counts.sh`, log checked non-empty **and** `dune test`'s own exit code checked (backlog-157):

| | dune test exit | log lines | cases / suites | FAIL | SKIP | zero-case |
|---|---|---|---|---|---|---|
| `origin/main` (stashed) | 0 | 4793 | 1697 / 117 | 0 | 23 | 4 |
| this branch | 0 | 4803 | 1701 / 118 | 0 | 23 | 4 |

Delta is exactly the new suite. No golden regressed — expected, since no committed golden covers a record kernel through the transpiler. `dune build @fmt` clean; no `*.opam` churn.

## Not verified

No OpenCL or Metal device was used. The claim that the pre-fix source is invalid rests on it naming an undeclared struct type, not on a vendor compiler having rejected it on this machine. `Metal_plugin.ml` *is* compiled on this Linux host (`sarek_metal__Metal_plugin.cmo` exists), so the signature change is type-checked, but no Metal toolchain ran. `Metal_plugin.generate_source` and `Opencl_plugin.generate_with_fp64` have no in-repo callers, so both signature changes are breaking for out-of-repo consumers.

---

# Follow-up: the gate itself was tested, and it failed (41418fd)

An independent red-prover mutated the *fixed* code three ways and the suite above stayed green for all three. Every item below was reproduced here first, then fixed, then re-measured red. **Baseline reproduced exactly**: `dune test` exit **0**, log **4803** lines, **1701 / 118**, 0 FAIL, 23 SKIP, 4 zero-case — every digit matching the table above.

## 1. Both markers were unordered, so the gate could not see a use-before-declaration

Moving the typedef block of `Sarek_ir_opencl.generate_with_types` (`sarek/codegen/Sarek_ir_opencl.ml:1168-1177`) from the prologue to the epilogue — the same bytes — emits

```c
__kernel void k(__global float* restrict out, int sarek_out_length) {
  point p;
  out[0] = (p.x + p.y);
}
typedef struct {
  float x;
  float y;
} point;
```

That is a use above its declaration, rejected by every OpenCL C compiler. **The gate stayed GREEN, exit 0, 4 cases OK** — both checks were unordered `contains`.

Fixed at `sarek/tests/unit/test_typedef_entry_points.ml:85` (`check_declared_before_use`): markers are compared by OFFSET, and the check fails unless the declaration's is strictly lower. Three legs, three distinct messages — no use marker (the case would be vacuous), no declaration marker (the original defect), declaration below the use.

**RED, exit 1, 3 failures:**

```
FAIL generate_with_fp64, record: declaration "} point;" is at offset 146 but the use
"point p" is at offset 72 — the declaration must come FIRST, this source is rejected
by any compiler.
FAIL generate_with_fp64, variant: declaration "} shape;" is at offset 210 but the use
"= make_shape_Circle(" is at offset 80 — …
FAIL transpiler/OpenCL: declaration "} point;" is at offset 215 but the use "point p"
is at offset 109 — …
```

**One correction to the prover.** It reported the six full-suite failures under this mutation as *all* device-backed OpenCL suites, hence uncovered on a runner with no OpenCL device. Measured here on a host with rusticl: two of the six are `codegen_golden` `opencl/record_kernel` and `opencl/variant_kernel` — string comparisons against committed goldens, needing no device. So ordering was not uncovered repo-wide. It was uncovered in the gate that claimed to cover it.

## 2. `of_source_with_abi`'s `~types` was read by nothing

Starving it — `sarek/transpile/Sarek_transpile.ml:349`, `~types:[]` in place of `~types:ir_kernel.kern_types` — left the **full suite identical in every digit**: exit 0, 4803 lines, 1701 / 118, 0 FAIL, 23 SKIP, 4 zero-case. Making the label required bought nothing, because nothing called it.

`transpile_cases` (`test_typedef_entry_points.ml:290`) now carries an emitter closure per row instead of a backend tag, and has a sixth row for `of_source_with_abi` — a second emission path, not a wrapper: it re-runs the pipeline and calls `Sarek_ir_wgsl` itself.

**RED, exit 1:**

```
FAIL transpiler/WGSL/of_source_with_abi: emitted source uses the type but never
declares it (missing marker "struct point").
```

`Metal_plugin.generate_source` (`sarek-metal/Metal_plugin.ml:287`) was worse: **zero callers outside its own file**, so the signature change was not type-checked against a single use. New suite `sarek-metal/test/test_metal_plugin_entry_points.ml` — that directory and not `sarek/tests/unit` because `sarek_metal` is `(optional)` and `sarek-metal/test` is the only place in the tree that links it. Reverted to `Sarek_ir_metal.generate`, **RED, exit 1:**

```
FAIL Metal_plugin.generate_source: emitted source uses the type but never declares it
(missing marker "} point;").
```

## 3. The "and no variant definitions" half was not falsifiable here

Every fixture had `kern_variants = []`, so the file's prose was wider than its assertions. Added a variant-carrying fixture + control (`test_typedef_entry_points.ml:152`). Deleting the `List.iter (gen_variant_def buf) k.kern_variants` line — **RED, exit 1:**

```
FAIL generate_with_fp64, variant: emitted source uses the type but never declares it
(missing marker "} shape;").
```

The use marker is `= make_shape_Circle(`, not the bare constructor name: `gen_variant_def` emits the constructor's *definition* inside the declaration block, so a bare name would be found within the declaration and the ordering check would hold vacuously.

**Two corrections to the prover.** Its mutation C was `List.iter (…)` → `ignore (List.iter (…))`, which still evaluates the iteration — a no-op, so its greenness was not evidence. And *"covered by nothing"* is too strong: a mutation that genuinely deletes the emission takes the full suite to **exit 1, FAIL: 6** — `codegen_golden opencl/variant_kernel`, `opencl_validation_sweep 2` and `22`, and a `whole-module-search-is-vacuous OpenCL` case. What holds is that *this* gate, whose docstring names variant definitions, could not tell.

## Suite after

`scripts/test-suite-counts.sh`, log checked non-empty and `dune test`'s own exit code captured (backlog-157):

| | dune test exit | log lines | cases / suites | FAIL | SKIP | zero-case | counter exit |
|---|---|---|---|---|---|---|---|
| branch before (166fb05e) | 0 | 4803 | 1701 / 118 | 0 | 23 | 4 | 0 |
| this commit (41418fd) | 0 | 4813 | 1705 / 119 | 0 | 23 | 4 | 0 |

Delta is exactly +2 cases in `typedef_entry_points` (variant case + control) and the new 2-case Metal suite. `dune build @fmt` clean; no `*.opam` churn. **No production code changed in this commit** — the four reds were produced by mutating it and reverting.

Also re-checked the original defect against the reworked gate: `generate_with_fp64` back to `generate` → 3 failures, exit 1, controls green.

## Round 3 (04320a6b) — bounded close-out

Three more mutations were measured green against the version above. All three are now red. **No production code changed in this commit either**; the reds below were produced by mutating it and reverting.

### 1. The variant half covered one entry point out of seven

The file's opening sentence claims every public emission entry point declares the record types **and variant types** it uses. The variant fixture added in round 2 covers only `Ocl.generate_with_fp64`. All six `transpile_cases` rows share `record_kernel_src`, which declares `type point = {x; y}` and no variant — so for `of_source` on CUDA/OpenCL/Metal/GLSL/WGSL and for `of_source_with_abi`, the variant half was exactly as unfalsifiable as before, **inside `emit_backend`, the function 166fb05e rewrote**.

`variant_transpile_cases` (`test_typedef_entry_points.ml:439-459`) is `transpile_cases` with a variant source and the variant markers, reusing `check_declared_before_use` unchanged. Per-row keying as before: `} shape;` for the C family, `struct shape` for GLSL/WGSL, `= make_shape_Circle(` for the use everywhere — the bare constructor name matches inside its own definition, which is part of the declaration block.

### 2. The fp64 pragma had no negative control

Lines 51-54 state the file's own discipline: each positive assertion is paired with a control asserting the typedef-less emitter does *not* emit the declaration. Every typedef assertion honoured it; `test_opencl_fp64_still_emits_the_pragma` did not. `generate_with_fp64` is what this PR made the OpenCL runtime path, so the blast radius of an unconditional pragma — a compile error on a device without `cl_khr_fp64`, not a no-op — grew in the same commit that left the direction untested.

### 3. "record type**s**" was checked with one record

`Sarek_ir_codegen.gen_record_typedefs` is shared by CUDA/OpenCL/Metal; both fixtures carried a single record, so a head-only walk was green repo-wide. `point_types` now carries `box` behind `point`, asserted separately in both files. That file is untouched by this PR, so this one is pre-existing — but the gate's prose claimed it.

### Reds observed

| # | mutation | site | result |
|---|---|---|---|
| M1 | `let k_types = {k_types with kern_variants = []}` | `Sarek_transpile.ml:242` (`emit_backend`) | exit 1 — `FAIL transpiler/CUDA, variant: emitted source uses the type but never declares it (missing marker "} shape;")` |
| M1b | same, on the ABI path | `Sarek_transpile.ml:347` | exit 1 — `FAIL transpiler/WGSL/of_source_with_abi, variant: … (missing marker "struct shape")` |
| M2 | `if true then` in place of `if Sarek_ir_analysis.kernel_uses_float64 k` | `Sarek_ir_opencl.ml:1223` | exit 1 — `FAIL control: a float32-only kernel must NOT get the cl_khr_fp64 pragma, so the positive assertion above is not passing for any reason` |
| M3 | `let types = match types with [] -> [] \| t :: _ -> [t]` | `Sarek_ir_codegen.ml:693` (`gen_record_typedefs`) | exit 1 in both files — `FAIL generate_with_fp64, second record: … (missing marker "} box;")` and `FAIL Metal_plugin.generate_source, second record: … (missing marker "} box;")` |

M1 alone left the suite at **exit 0** before this commit, as did M2 and M3.

### Suite

`dune test --force` (forced — the shared dune cache on this host can return exit 0 on a zero-line log): **exit 0**, 4816-line log, 1675 `[OK]` lines. `typedef_entry_points` 8 cases, `metal_plugin_entry_points` 2 cases, both green. `dune build @sarek/tests/unit/fmt @sarek-metal/test/fmt` clean; no `*.opam` churn.

## Still not verified

No OpenCL or Metal device compiler is invoked by either of these two suites; they assert on emitted text. The rusticl-backed suites that *do* compile OpenCL are what ran the ordering claim end to end, and they are not portable — which is the reason these exist.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed generated GPU code for kernels using custom record and variant types by ensuring required type declarations are emitted (or included) across supported OpenCL, Metal, and transpilation paths.
  - Centralized OpenCL FP64 pragma handling so the FP64 extension line is included only when needed.
  - Corrected ABI-based WGSL output to include required type declarations.

- **API Changes**
  - OpenCL and Metal code-generation entry points now require an explicit `~types` type-definition collection.

- **Tests**
  - Added coverage to validate typedef/variant declarations across public code-generation and transpilation entry points.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -106,8 +106,13 @@
   `Metal_plugin.generate_source` now take a required `~types`; the OpenCL
   backend's own `generate_source` was two hand-kept copies of "typedefs +
   pragma" and now calls `generate_with_fp64` so the two cannot diverge again.
-  Guarded by `test_typedef_entry_points`, each assertion paired with a control
-  against the typedef-less emitter so it cannot pass vacuously (backlog-155).
+  Guarded by `test_typedef_entry_points` (OpenCL + transpiler) and
+  `test_metal_plugin_entry_points` (the Metal entry point, which is only
+  callable from `sarek-metal/test` because `sarek_metal` is an optional
+  library). Every marker pair is checked for ORDER, not presence — a typedef
+  emitted BELOW the code that uses it satisfies a substring check and no
+  compiler — and each assertion is paired with a control against the
+  typedef-less emitter so it cannot pass vacuously (backlog-155).
 - `Sarek_df64` silently ran at plain float32 precision on real NVIDIA
   hardware (CUDA/PTX and NVIDIA OpenCL): `ptxas` contracted the multiply in
   `two_prod` into the `add`/`sub` of the `quick_two_sum` closing `df64_mul`,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -92,6 +92,22 @@
 
 ### Fixed
 
+- Three public emission entry points routed to a backend's typedef-less
+  `generate` instead of `generate_with_types`, so a kernel carrying a
+  `[@@sarek.type]` record emitted source naming struct types that were never
+  declared. The one with real reach is `Sarek_transpile.of_source` — the whole
+  public transpiler API, all five backends — which emitted
+  `point p = (point){1.0f, 2.0f};` with no `typedef struct {…} point;` anywhere
+  above it, on CUDA, OpenCL, Metal, GLSL and WGSL alike. The other two are
+  `Sarek_ir_opencl.generate_with_fp64` (re-exported as
+  `Opencl_plugin.generate_with_fp64`) and `Metal_plugin.generate_source`, a
+  top-level alias of `Sarek_ir_metal.generate` under the same name as the
+  runtime path that does use `generate_with_types`. `generate_with_fp64` and
+  `Metal_plugin.generate_source` now take a required `~types`; the OpenCL
+  backend's own `generate_source` was two hand-kept copies of "typedefs +
+  pragma" and now calls `generate_with_fp64` so the two cannot diverge again.
+  Guarded by `test_typedef_entry_points`, each assertion paired with a control
+  against the typedef-less emitter so it cannot pass vacuously (backlog-155).
 - `Sarek_df64` silently ran at plain float32 precision on real NVIDIA
   hardware (CUDA/PTX and NVIDIA OpenCL): `ptxas` contracted the multiply in
   `two_prod` into the `add`/`sub` of the `quick_two_sum` closing `df64_mul`,
--- a/sarek-metal/Metal_plugin.ml
+++ b/sarek-metal/Metal_plugin.ml
@@ -277,5 +277,11 @@ let find_intrinsic = Metal_intrinsics.find
 (** Generate Metal source with custom types *)
 let generate_with_types = Sarek_ir_metal.generate_with_types
 
-(** Generate Metal source for a kernel *)
-let generate_source = Sarek_ir_metal.generate
+(** Generate Metal source for a kernel.
+
+    Takes [~types] (normally [ir.kern_types]) and routes through
+    {!Sarek_ir_metal.generate_with_types}, same as the backend's own
+    [Backend.generate_source]. It used to alias {!Sarek_ir_metal.generate},
+    which emits no record typedefs and no variant definitions — under a name
+    that reads exactly like the runtime path it was NOT (backlog-155). *)
+let generate_source ~types ir = Sarek_ir_metal.generate_with_types ~types ir
--- a/sarek-metal/test/dune
+++ b/sarek-metal/test/dune
@@ -14,3 +14,15 @@
  (name test_metal_compile_options)
  (modules test_metal_compile_options)
  (libraries sarek_metal alcotest))
+
+; backlog-155: Metal_plugin.generate_source used to alias the typedef-less
+; Sarek_ir_metal.generate. It has no caller outside its own file, so this is
+; the only place its signature is type-checked against a use and the only
+; place its routing is executed. sarek_metal is (optional) and only linkable
+; from this directory, which is why the Metal case is not in
+; sarek/tests/unit/test_typedef_entry_points.ml with the others.
+
+(test
+ (name test_metal_plugin_entry_points)
+ (modules test_metal_plugin_entry_points)
+ (libraries sarek_metal sarek_ir alcotest))
--- a/sarek-metal/test/test_metal_plugin_entry_points.ml
+++ b/sarek-metal/test/test_metal_plugin_entry_points.ml
@@ -1,0 +1,146 @@
+(******************************************************************************)
+(* SPDX-License-Identifier: CECILL-B                                          *)
+(* SPDX-FileCopyrightText: 2026 Mathias Bourgoin <mathias.bourgoin@gmail.com> *)
+(******************************************************************************)
+
+(** [Metal_plugin.generate_source] must declare the record types it uses
+    (backlog-155).
+
+    This is the Metal half of [sarek/tests/unit/test_typedef_entry_points.ml],
+    and it lives here rather than there for a mechanical reason: [sarek_metal]
+    is an [(optional)] library and the only place in the tree that links it is
+    [sarek-metal/test], so this is the only directory from which the entry point
+    can be CALLED.
+
+    That mattered more than it sounds. [Metal_plugin.ml:287] used to be
+    [let generate_source = Sarek_ir_metal.generate] — a top-level alias of the
+    typedef-less emitter, under the same name as [Backend.generate_source] in
+    the module above it, which does route through [generate_with_types]. When
+    the alias was replaced by a [~types]-taking function, the change compiled
+    with nothing to check it: [generate_source] at that line has ZERO callers
+    outside its own file, so neither the new signature nor the new routing was
+    exercised by anything. A signature change that no call site type-checks
+    against, and a routing change that no test executes, are both free.
+
+    Same shape as the OpenCL/transpiler file: the declaration marker must appear
+    BEFORE the use marker (a typedef emitted below the code that needs it is
+    invalid Metal all the same), and each assertion is paired with a control
+    against [Sarek_ir_metal.generate] so it cannot pass vacuously. *)
+
+open Sarek_metal
+open Sarek_ir_types
+
+let index_of haystack needle =
+  let nl = String.length needle and hl = String.length haystack in
+  let rec go i =
+    if i + nl > hl then None
+    else if String.sub haystack i nl = needle then Some i
+    else go (i + 1)
+  in
+  if nl = 0 then Some 0 else go 0
+
+let contains haystack needle = index_of haystack needle <> None
+
+let check_declared_before_use ~ctx ~decl ~use src =
+  match (index_of src decl, index_of src use) with
+  | _, None ->
+      Alcotest.failf
+        "%s: the body never uses the type (marker %S), so this case proves \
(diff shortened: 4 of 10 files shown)


mathiasbourgoin:
Closing as **superseded** — triaged by content comparison, not by assuming duplication (backlog-166).

Both halves of this PR's unique content are in `main`:

- **fp64 pragma routing** — `sarek/codegen/Sarek_ir_opencl.ml:1186` and `sarek-opencl/Opencl_plugin.ml:196` emit `#pragma OPENCL EXTENSION cl_khr_fp64 : enable`, with a golden at `test_codegen_golden.ml:2590`. Landed via #356 (backlog-155).
- **the OpenCL clang gate incl. the `-cl-ext=-cl_khr_fp64` negative control** — `sarek/tests/codegen_golden/opencl_gate/opencl_clang.ml:47` plus `sarek/tests/unit/test_opencl_gate.ml:475`. Landed via the backlog-128 OpenCL validation gate.

And the branch is now **behind** main in a way that would regress it: it has zero occurrences of `No_device_spelling` where main has three, so merging it would undo #359's registry refusal (GLSL/WGSL falling through to the CUDA spelling). It also predates the dune wiring added in #364.

Nothing to salvage. Note the contrast with #351, whose captured `--verbose` logs turned out to expose two live false greens in merged main (fixed in #365 and #367) — which is why these two were triaged by comparing content rather than closed together as a pair.

PR #354, opened 2026-07-28
fix(f64): literal precision, three Float64 conversions, helper-body param substitution

mathiasbourgoin:
Three defects, found by writing one kernel that combined f64 transcendentals with
a long accumulation — a shape nothing in the suite exercised. It failed on four
of five local devices, each for a different reason.

## What is fixed

**float64 literals lost ~11 bits.** The PPX rebuilt every float constant with
`string_of_float` (= `"%.12g"`), so `3.14159265358979312G` reached the backends
as `3.14159265359`. That is the entire purpose of the `G` suffix defeated
silently, and `test_float64_kernel_arith` — the suffix's own regression guard —
could not see it: its literals are `0.0G`, `2.0G`, `4.0G`, all exact in twelve
digits. The three call sites become one function, which is what makes the
property checkable at all: a grep for `string_of_float` pins a *spelling*
(`Float.to_string` walks past it), a named chokepoint can be pinned
structurally. The shared spelling also restores the decimal-point repair that
all five backend formatters apply and all three PPX sites omitted.

**Three Float64 conversions were declared but unimplemented** in the interpreter
and in GLSL — `of_int32`, `to_int32`, `of_float32`. int32 is the DSL's integer
type, so `of_int32` is the one user code reaches for first. A gap in the
interpreter matters more than a gap in one backend: it is the cross-backend
oracle, and it could not run kernels the frontend accepts.

**Helper function BODIES were never protected from parameter substitution.**
GLSL exposes scalar kernel parameters as `#define <name> pc.<name>`, WGSL as
`params.<name>`, and both ran their shadow-rename pre-pass over `kern_body`
only. GLSL's `#undef`/`#define` dance around helpers is real but is computed
from PARAMETER names alone, so a helper whose *body* declared a colliding local
was covered by neither. GLSL then emitted `int pc.n = ...` and glslang rejected
the shader; WGSL declared the local, never read it, rewrote every reference to
`params.n`, and returned the **wrong number** with no error. Fixed by running
the existing pre-pass over every emitted helper body on both backends, keyed on
the actual macro/field set.

That last one subsumes an earlier version of this branch, which prefixed the
generated softmath locals into a `sarek_sm_` namespace. CodeRabbit was right
that this only *moved* the collision — nothing here reserves `sarek_` — and it
left the two pre-existing cases untouched. The prefix and its 45 lines of golden
churn are gone.

## What was deliberately NOT done

- **`to_int` and `to_float32` are not implemented.** Their device templates
  truncate/round while their `ocaml` field — which `Sarek_float64_native.ml`
  mirrors and executes — is `Stdlib.int_of_float` (63-bit) and the *identity*.
  Three implementations already disagree; adding arms would make that reachable
  on two more backends rather than settle it.
- **The interpreter's vector-helper-parameter fix was reverted.** It makes those
  folds run, but underneath, helper parameter ids and body-local ids come from
  two overlapping numbering spaces and `lookup_var` resolves by id before name,
  so a tail-recursion temporary can clobber a parameter. Measured: a
  single-helper `vsum acc v k n` fold over four `1.0`s returns **0** on the
  interpreter where Vulkan and Native return **4**, silently. Trading a loud
  refusal for a sometimes-wrong number is a regression. The repair lands in the
  immediate follow-up, with the id fix. The test records this as a documented
  gap and goes **red when it closes**, with an instruction to delete the arm.
- **The root cause of the conversion gap is untouched.**
  `emit_registry_template` is wired into CUDA, Metal and OpenCL only — GLSL and
  WGSL have no registry-template fallback, so *every* template-declared
  intrinsic dies on both. This closes the symptom on one of the two.

## Verification

`dune build @fmt` clean, `dune build` clean, `dune test --force` a 4835-line log
with zero `[FAIL]`. Both e2e tests bit-for-bit on 5 devices (2× Vulkan/RADV,
Native, 2× Interpreter). **Every fix reverted individually turns its covering
assertions red** — verified for each.

The new codegen test caught one of my own mistakes, and it is worth naming
because it is the class this whole branch is about. It first asserted the absence
of `int pc.n =` in the emitted GLSL. That string can never appear — macro
expansion happens in the GLSL compiler, not in our output, which literally
contains `int n = (q + 1);` — so the assertion could not fail, and it stayed
green with the fix reverted. Prove-red is how it was found. It now compiles the
shader with `glslangValidator`, and its textual companion requires both halves
(macro present AND declaration absent) so it cannot be satisfied by a shader
that never exposes the parameter.

Reviewed by four specialist passes (reviewer, KB code-quality, architect,
unscoped) plus CodeRabbit; each found something real, and the branch was
**narrowed** in response rather than widened. Cross-runtime degraded on five
attempts, all caller error on my side, and is recorded as a skip rather than
left looking like a clean pass.

## Summary by CodeRabbit

* **New Features**
  * Added Float64 scalar conversion intrinsics support (e.g., `of_int32`, `of_float32`, `to_int32`) with consistent interpreter/GLSL semantics.
* **Bug Fixes**
  * Fixed GLSL/WGSL shader generation issues involving helper-local name collisions to prevent incorrect behavior.
  * Improved Float64 literal rendering for more reliable source round-tripping.
* **Tests**
  * Added e2e and golden regression tests for Float64 conversions, helper shadowing, and tail-recursive helper generation.
* **Documentation**
  * Documented a known interpreter semantic gap affecting helper vector parameter binding.

### DIFF
--- a/sarek/codegen/Sarek_ir_glsl.ml
+++ b/sarek/codegen/Sarek_ir_glsl.ml
@@ -380,6 +380,43 @@ let f64_root_helpers name =
 
 (** {1 Expression Generation} *)
 
+(** The [Float64] scalar conversion intrinsics that every implementation agrees
+    on: [of_int], [of_int32], [of_float32], [to_int32].
+
+    They are declared in Sarek_float64/Float64.ml and type-check in the DSL, but
+    had no GLSL arm at all, so any kernel using one died with "Unknown
+    intrinsic" — notably [Float64.of_int32], which is the conversion user code
+    reaches for first since int32 is the DSL's integer type.
+
+    Two are DELIBERATELY ABSENT. [to_int] and [to_float32] are declared with
+    device templates that round/truncate while their [ocaml] field — which is
+    what Sarek_float64_native.ml mirrors and executes — is [Stdlib.int_of_float]
+    (63-bit) and the IDENTITY respectively. So three implementations already
+    answer differently, and adding GLSL arms would make that disagreement
+    reachable on another backend rather than resolve it. Deciding those two
+    semantics is tracked separately.
+
+    This lives in [pre_hook] rather than [arm] because the module path is needed
+    to scope it, and only [pre_hook] receives it. The scoping is deliberate: the
+    match is on bare NAME, and [pre_hook] runs BEFORE the path-qualified pure
+    registry, so without the [is_f64] guard any module declaring [of_int] or
+    [to_int32] would be silently captured here and lowered as a GLSL constructor
+    cast, shadowing its own device template. Float32 declares both names today.
+*)
+let glsl_conversions = ["of_int"; "of_int32"; "of_float32"; "to_int32"]
+
+let gen_glsl_conversion buf ~gen_expr name args =
+  let target = match name with "to_int32" -> "int" | _ -> "double" in
+  match args with
+  | [e] ->
+      Buffer.add_string buf target ;
+      Buffer.add_char buf '(' ;
+      gen_expr buf e ;
+      Buffer.add_char buf ')'
+  | _ ->
+      Codegen_error.raise_error
+        (Codegen_error.invalid_arg_count name 1 (List.length args))
+
 let rec gen_expr buf = function
   | EConst (CInt32 n) -> Buffer.add_string buf (Int32.to_string n)
   | EConst (CInt64 n) -> Buffer.add_string buf (Int64.to_string n ^ "L")
@@ -746,6 +783,9 @@ and glsl_backend =
           true)
         else if name = "fmod" then (
           Dispatch.emit_call ~gen_expr buf !current_fmod_name args ;
+          true)
+        else if is_f64 && List.mem name glsl_conversions then (
+          gen_glsl_conversion buf ~gen_expr name args ;
           true)
         else false);
     post_hook = (fun _ _ _ _ -> false);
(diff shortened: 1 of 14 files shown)


mathiasbourgoin:
@coderabbitai resume

Reviews were auto-paused, so the standing `CHANGES_REQUESTED` predates the two
commits that answer it. Status of each comment, and one correction:

**Major, helper formal-parameter collisions (Sarek_ir_wgsl.ml)** — confirmed and
fixed. Reproduced first: `twice (n : int32) = n * 3l` under a kernel scalar `n`
emitted `return (params.n * 3i)` — the argument ignored, the uniform read.

Not by renaming, though. The formals are **removed from the substitution set**
for the duration of that helper, which is what lexical shadowing means; renaming
them would break the positional correspondence with the call site. GLSL escapes
this case because its `#undef`/`#define` dance already covers parameter names —
that dance is what WGSL has no equivalent of.

**Minor, assert the USE resolves to the renamed local** — confirmed and fixed,
and this one was the sharpest comment on the PR. `contains glsl "* 2)"` matched
both the correct output and a half-rename, and `return (pc.n * 2);` *compiles* —
so the glslangValidator gate did not cover it either, because only the
declaration side is syntactically invalid. The renamed identifier is now read out
of the emitted source and the use must carry the same one, on both backends, plus
a direct assertion that the helper never reads the push constant. 7 assertions
red with `sarek/codegen/` at `origin/main`.

**Minor, golden case for a vector-length macro collision** — not added, because
it is not expressible. The PPX rejects any identifier beginning with `sarek_`
(`sarek/ppx/Sarek_error.ml`), so a parameter named `sarek_<vec>_length` cannot be
written; I tried, and the frontend refused the test kernel. `len_names` is now in
the collision set anyway so the guard does not depend on that check holding, and
the test header records why there is no test rather than leaving a silent gap.

**Correction to the earlier Major on `sarek_sm_`** — that same reservation
refutes its premise. There is no "valid scalar parameter named `sarek_sm_n`"; the
frontend refuses it. So the prefix approach was a real guarantee rather than a
displacement. I had also repeated the opposite claim in a commit message, taken
from an internal review pass without checking it, and that message is corrected.

The structural fix replacing the prefix is still the better one — it covers the
two pre-existing cases the prefix never touched, including the silent WGSL pair,
and it moves no goldens — but the stated reason for abandoning the prefix was
wrong and is now recorded as such.

PR #356, opened 2026-07-28
fix(codegen): the plain generate was a lossy copy of generate_with_types, in all five backends

mathiasbourgoin:
Filed (backlog-155) as an OpenCL bug: `Opencl_plugin.generate_with_fp64` re-exports a
path that emits no record typedefs. It is the same defect in **all five**
source-emitting backends, and the plugin re-export is just where it was noticed.

## The defect

Each backend carried its own 34–84 line copy of the emit sequence under `generate`,
and every copy dropped the same step: no record typedefs, no variant typedefs, and
on four of the five no `current_variants` assignment either. A kernel with a record
emitted source that **uses** the struct and never **declares** it:

```c
__global__ void record_kernel(Point2* __restrict__ pts, int sarek_pts_length) {
```

against no `typedef struct { ... } Point2`. Not a diagnostic — source the device
compiler rejects. WGSL was worse: `generate` set `current_variants`, which arms
SMatch payload extraction, for types it then never declared.

## The fix

```ocaml
let generate k = generate_with_types ~types:k.kern_types k
```

Sound because `~types` has exactly the type of the `kern_types` **field** of the
kernel it travels with (`spoc/ir/Sarek_ir_types.ml:379`), and every production call
site already passed that field — the parameter was redundant with the record it
rides in. **Net −178 lines.**

Verified as a strict superset **before** relying on it: diffed each `generate`
against its `generate_with_types`, and the delta is only the three type-emitting
lines, on all five. So output is byte-identical for a kernel carrying neither a
record nor a variant — which the golden suite confirms, 74 goldens unchanged.

## Two things this repairs that the filed item did not mention

**`Sarek_transpile.emit_backend` routes all five backends through the plain
`generate`.** So `sarek-transpile` on any kernel carrying a record or a variant
emitted undeclared-type source, on every backend, for every user. Fixed by
inheritance; the transpiler is not touched.

**`generate_with_fp64` calls `generate`**, so the filed OpenCL symptom is fixed by
inheritance too.

Evidence the duplication had already bitten once: Metal's `generate_with_types`
carries a comment saying variant defs were *"previously omitted here, unlike the
CUDA/OpenCL backends"*. Someone found that divergence and repaired **one** of the
two copies. The second copy kept the bug for both of them.

## The test

15 cases — 5 backends × {record, variant, scalar control} — asserting **byte
equality** of the two entry points, rather than a substring probe for `typedef` or
a singleton-identifier diff. One entry point is *defined* as a special case of the
other, so identical output is both the strongest available assertion and the
cheapest: no per-backend builtin list, no external validator, and a partial fix
cannot satisfy it (emitting the record typedef but not the variant typedef still
differs).

Record and variant are separate kernels because the omission was per-step — a
backend could restore one and not the other.

**Prove-red split exactly as designed:** reverting `sarek/codegen/` to `origin/main`
turns the **10** record/variant cases red and leaves all **5** controls green. The
control is load-bearing: it carries neither type, both entry points agreed on it
even before the fix, so a red there would indict the test rather than the backends.
Measured from a checkpoint commit and restored **by sha** — on a branch with nothing
committed, `HEAD` is the base, so restoring from `HEAD` discards the work.

Refusals are compared, not skipped: a backend may legitimately reject a kernel
(Metal has no f64, GLSL no f16), and that is agreement only if both entry points
refuse with the same error. Treating a raise as "not applicable" would let a
backend drop out of the matrix silently.

## Scope and non-findings

- **Agreement, not correctness.** If both entry points emit the same wrong typedef,
  this is green; the goldens and the device-compiler gates own that.
- **PTX untouched**, and the test header says why: its `generate_with_types` takes
  `~types:_` and ignores it, so the two entry points are equal by construction and a
  green PTX row would assert nothing.
- `Sarek_ir_glsl.set_framework` looked like an adjacent omission — the transpiler
  sets `current_framework` on four modules and GLSL is not one of them. Checked
  before reporting it: **GLSL has no `current_framework` at all**, so the code is
  correct and there is nothing to fix.

## Verification

`dune build` clean, `dune test --force` 4885 lines / **0 `[FAIL]`** — up from
4864, the +15 cases being visible rather than asserted.

`@fmt` **failed in CI on the first push, and the earlier claim here that it was
clean was measured on the tree before the test file existed.** I promoted
formatting, then added `test_entry_point_agreement.ml`, and never re-ran the
gate — so the check was real but the state it certified was not the state that
shipped. Fixed by reformatting and force-pushing; `@fmt` now exits 0 on the
actual branch head. Worth stating rather than quietly amending: a verification
that runs before the last change is the same shape as the rest of this branch —
a check whose result is narrower than the claim made from it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed generated CUDA, GLSL, Metal, OpenCL, and WGSL code to include required record and variant type definitions.
  * Improved handling of variant match payloads during shader and kernel generation.
  * Ensured generated output remains consistent across equivalent generation entry points.

* **Tests**
  * Added cross-backend tests covering record, variant, and scalar kernels to verify identical generated output.

### DIFF
--- a/sarek/codegen/Sarek_ir_cuda.ml
+++ b/sarek/codegen/Sarek_ir_cuda.ml
@@ -903,44 +903,6 @@ let cuda_header_for (k : kernel) =
     cuda_fp16_include ^ sarek_f32_barrier_decl ^ cuda_header
   else cuda_header
 
-(** Generate complete CUDA source for a kernel *)
-let generate (k : kernel) : string =
-  let buf = Buffer.create 4096 in
-
-  (* Header *)
-  Buffer.add_string buf (cuda_header_for k) ;
-
-  (* Generate helper functions before kernel *)
-  List.iter (gen_helper_func buf) k.kern_funcs ;
-
-  (* Kernel signature *)
-  Buffer.add_string buf "__global__ void " ;
-  Buffer.add_string buf k.kern_name ;
-  Buffer.add_char buf '(' ;
-
-  (* Parameters *)
-  List.iteri
-    (fun i p ->
-      if i > 0 then Buffer.add_string buf ", " ;
-      gen_param buf p)
-    k.kern_params ;
-
-  Buffer.add_string buf ") {\n" ;
-
-  (* Local declarations *)
-  List.iter (gen_local buf "  ") k.kern_locals ;
-
-  (* Body *)
-  gen_stmt buf "  " k.kern_body ;
-
-  (* Close kernel *)
-  Buffer.add_string buf "}\n" ;
-
-  (* Close extern "C" *)
-  Buffer.add_string buf "}\n" ;
-
-  Buffer.contents buf
-
 (** Generate CUDA variant type definition *)
 let gen_variant_def buf v =
   Sarek_ir_codegen.gen_variant_def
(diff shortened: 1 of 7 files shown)


mathiasbourgoin:
@coderabbitai review

PR #357, opened 2026-07-28
fix(scripts): the count gate passed for green off a run that had failed

mathiasbourgoin:
backlog-157. `scripts/test-suite-counts.sh` is the tool an operator or an agent
uses to report "N cases, 0 FAIL" — the instrument this repo audits its other gates
with. It reported green off runs that had failed.

## Two ways, neither caught

Gates 1–3 all ask whether the total is *complete for the suites present*. None asked
the prior question: **did the run succeed**.

**A build failure part-way through.** `dune test --force` builds and runs suite by
suite, so a compile error in one test leaves every earlier epilogue intact. Measured
by breaking one test file on this tree: **45 parsed suites**, gate 2 satisfied (45
epilogues, 45 counts), gate 3 satisfied (45 > 10) — a plausible total, **exit 0**,
off a `dune test` that exited **1**. The number is not wrong for what it counted; it
is wrong as an answer to "how many tests are there", which is the question it gets
asked.

**Failing cases.** `FAIL : 3` printed with exit 0 means
`test-suite-counts.sh log && echo all good` prints "all good". The counts were
honest and the exit code was not — a shell reads only the code.

## What changed

A run that did not complete is **exit 5**; failing cases are **exit 4**. Counts still
print in both — they are useful, they just must not pass for green in a chain.

Separate codes on purpose: a complete-but-red run is a different fact from a run that
died, and one catch-all could not say which. A died run **outranks** its own FAIL
tally, because that tally is taken over a partial run and would understate it.

`--dune-exit` is the mechanism; the log scan is the fallback:

```
dune test --force > t.log 2>&1; scripts/test-suite-counts.sh --dune-exit $? t.log
```

The script cannot derive dune's status from a log, and the advertised pipe form
structurally cannot either — in `a | b` the shell reports b's code, and `pipefail` is
the caller's to set. So the caller passes the one authoritative fact.

The `^Error:` / `^File "` heuristic covers the pipe form. **Measured before the
pattern was chosen, not after:** 0 matches in a green 4885-line log, 0 in a log of
genuinely *failing* alcotest cases (alcotest indents its output, dune does not), 2 in
a deliberately broken build. So it separates "the build died" from "a test failed" —
which is why those are different exit codes rather than a guess.

## A defect in the fix, found by testing the flag rather than reading it

`--dune-exit ""` was accepted and silently treated as "not told". So
`--dune-exit "$SOME_UNSET_VAR"` demoted the authoritative check back to the heuristic
and let a failed run exit 0 — **the exact shape this flag exists to close, reached
through the flag that closes it.** Given-empty is now a usage error; only an *absent*
flag means nobody asked dune. My in-source comment had also described the failure in
the wrong direction, and is corrected rather than left.

## Verification

- Covering test: **37 PASS / 0 FAIL** (was 27; +10 cases), including
  `a died run outranks its failing cases (5, not 4)` and
  `empty --dune-exit exits 2 (not silently ignored)`.
- `scripts/prove-red.sh`: **4 subjects, 16 mutations**, every one red as declared.
  Three are new — `build-failed` (exit 5), `dune-exit-nonzero` (exit 5, an
  *environment* mutation like the existing `empty-stdin`: no file changes, because the
  authoritative fact is one a log cannot carry) and `failing-cases` (exit 4). Each
  pins its exit code specifically, so 4 and 5 being confused would be caught.
- `scripts/check-kb-properties.sh` green; `KB-GATE-SUITE-COUNTS` restated to say the
  rule out loud — "0 FAIL of N" is trustworthy only if the run that produced N
  completed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test-result classification to clearly separate interrupted/incomplete runs from completed runs with failing cases.
  * Added precise exit-code behavior and precedence: build failures or non-zero runner exit produce **5**, failing cases produce **4**, success stays **0**; invalid options remain **2**.
  * Added support for passing through the test runner exit status via `--dune-exit` with stricter validation.
* **Documentation**
  * Updated the gate documentation to clarify partial suite-count semantics and the **4/5** failure-mode mapping.


mathiasbourgoin:
@coderabbitai review

mathiasbourgoin:
@coderabbitai review

PR #358, opened 2026-07-28
fix(interp): an interpreter error must name what went wrong (+ delete the dead id allocator)

mathiasbourgoin:
backlog-161 and backlog-162.

## backlog-161 as filed was three claims. One true, one inaccurate, one false.

I checked each before writing code, and all three verdicts are in the tree rather
than three fixes quietly shipped.

**TRUE — no `Printexc` printer.** Uncaught, an interpreter error printed as an opaque
constructor with none of `error_to_string`'s detail, so every test that had to report
one carried its own `describe` wrapper — there is one in `test_helper_scoping.ml`
written for exactly this. The three sibling error modules (`Execute_error`,
`Fusion_error`, `Kirc_error`) all register a printer; this was the odd one out.

**INACCURATE — "`LArrayElem` has no bounds check, so an out-of-range store is silent
memory corruption."** OCaml bounds-checks `a.(i) <- v`. The store raises
`Invalid_argument "index out of bounds"` and nothing was ever corrupted. The real
defect is smaller and still worth fixing: the arm raised a bare OCaml exception naming
neither array, index nor length, while the `LArrayElemExpr` arm **directly below it**
raised `Array_bounds_error` naming all three — two adjacent arms of one function
disagreeing about whose error this is. It matters because the interpreter is the
cross-backend oracle: the same store on a GPU backend is genuine undefined behaviour,
and this is the one side able to describe it.

**FALSE — "`default_for_type` is missing float arms."** `TReg Float32 -> TEFloat 0.0`
and `TReg Float64 -> TEDouble 0.0` are both there and always were. What *is* missing is
`Float16`, which fell through to the `TEInt 0` wildcard — an int-shaped initializer
carrying an f16 type. **Latent, not observed**: the value is overwritten before use,
which is why nothing caught it. Shipped anyway because "it's just a placeholder" is
exactly the silently-succeeding-wildcard justification #94 exists to refuse — and
claimed as latent, not as a repaired bug.

## backlog-162 — the dead allocator

`fresh_id` over a `next_var_id` field had **one** occurrence in the tree: its own
definition. Deleted with the field. The measurement is why it is deleted rather than
left: an earlier commit message of mine called it one of three *live* id allocators,
and that was false. After #355 there are two conventions, disjoint by construction —
the typer's monotone counter, and the negative range for tuple temporaries.

## Tests

Two new cases, both proven red from a checkpoint commit against `origin/main`:
`assign_out_of_bounds` fails with *"raised OCaml's Invalid_argument: the diagnostic
names neither array, index nor length"*, and `error_has_printer` fails outright.

The `Float16` arm has **no test**, because it has no observable effect. Saying so beats
inventing an assertion that would pass either way.

Both assertions are built to fail for the right reason: `assign_out_of_bounds` matches
the **constructor**, not the rendered string, so a reworded `error_to_string` cannot
silently reclassify it; `error_has_printer` requires the message to **contain the array
name, index and length** rather than be non-empty — `Printexc.to_string` always returns
something, so a length check would pass with no printer registered at all.

## Verification

`dune build` clean, `@fmt` clean (re-run **after** the last file was added — that
ordering is what broke #356's first push). Suite via `scripts/test-suite-counts.sh`:
**1712 cases across 118 suites, 0 FAIL**, 23 SKIP.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved array bounds errors with clear details, including the array name, invalid index, and length.
  * Fixed initialization for tail-recursive `Float16` code paths.
  * Standardized interpreter error messages for clearer diagnostics.

* **Tests**
  * Added coverage for array bounds handling and readable interpreter error output.

### DIFF
--- a/sarek/interp/Interp_error.ml
+++ b/sarek/interp/Interp_error.ml
@@ -57,3 +57,14 @@ let error_to_string = function
 
 (** Raise an interpreter error *)
 let raise_error err = raise (Interpreter_error err)
+
+(* Uncaught, [Interpreter_error] printed as an opaque constructor with none of
+   [error_to_string]'s detail, so every test that had to report one carried its
+   own `describe` wrapper — see the one in
+   sarek/tests/e2e/test_helper_scoping.ml, written for exactly this. The three
+   sibling error modules (Execute_error, Fusion_error, Kirc_error) all register a
+   printer; this one was the odd one out. *)
+let () =
+  Printexc.register_printer (function
+    | Interpreter_error err -> Some (error_to_string err)
+    | _ -> None)
--- a/sarek/interp/Sarek_ir_interp_eval.ml
+++ b/sarek/interp/Sarek_ir_interp_eval.ml
@@ -587,7 +587,23 @@ and assign_lvalue state env lv value =
   | LArrayElem (arr, idx_expr) ->
       let a = get_array env arr in
       let i = to_int (eval_expr state env idx_expr) in
-      a.(i) <- value
+      (* Same check as the LArrayElemExpr arm below, which had it while this one
+         did not — adjacent arms of one function disagreeing about whether an
+         out-of-range store is an interpreter error or an OCaml one.
+
+         NOT a memory-safety fix, and the backlog entry that claimed it was is
+         wrong: OCaml bounds-checks [a.(i) <- v] and raises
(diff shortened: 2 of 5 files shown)


mathiasbourgoin:
@coderabbitai review

mathiasbourgoin:
@coderabbitai review

PR #360, opened 2026-07-28
refactor(ir): a base record so a new kernel field costs zero call sites

mathiasbourgoin:
**Pure refactor**, PR A of the backlog-160 split you approved. No behaviour change, no
golden artifact touched, suite identical. The field and the top-level emission come in
PR B.

## Why

OCaml requires every field at every record literal, so adding one field to the kernel
record meant editing every construction site — which is why the type has been worked
*around* rather than extended. A record **update** names only the fields it sets, so
once a site is written that way the next field costs it nothing.

`default_kernel` is added to **both** kernel types, because there are two:
`Sarek_ir_types.kernel` (consumed by the five backends) and `Sarek_ir_ppx.kernel`
(produced by lowering), bridged by `Sarek_ir_conv.conv_kernel`. A field added to one
almost always has to be added to the other and threaded through the converter;
migrating one and not the other would just move the problem. `make_kernel` is provided
alongside, with labels, for new code.

## Form

`{default_kernel with ...}` rather than restructuring every site into
`make_kernel ~name ~body ()`. Same property, but an update is an insertion plus the
deletion of the fields already at their default, whereas the labelled form rewrites
each multi-line nested literal whole.

**44 files, +179 −712, net −533 lines.**

Dropping the default-valued fields is *not* cosmetic: warning 23
(`useless-record-with`) is an error in this tree and fires whenever a `with` lists
every field — which most of these sites did. Removing them is what makes the update an
update.

## The count was wrong twice, and both corrections are in the commit

| stage | figure | why it was wrong |
| --- | --- | --- |
| first report | 149 literals, one type | didn't know `Sarek_ir_ppx.kernel` existed |
| second | 150 sites | `grep kern_name =` counts field READS and other records |
| **measured** | **119 literals**, two IR types | see below |

`PtxKernelSpec.ml` reads `k.kern_name` into a `ptx_kern_name`. `Sarek_typed_ast`'s
kernel has `kern_name : string option` next to `kern_module_items`/`kern_loc` — a
**third** record sharing the field name. `test_typer.ml` was converted by the script
and reverted whole once the compiler named it.

## Verification

The claim is *"nothing changes"*, so the evidence has to be that nothing changed:

- `dune build` clean, `@fmt` clean
- `dune test --force`: **1725 cases across 119 suites, 0 FAIL** — the same numbers as
  before the change
- **zero non-`.ml` files modified**, i.e. no generated golden artifact moved

The codegen goldens are the real oracle here: they are byte-compared, so a literal that
lost a field it needed would surface there rather than in a pass/fail count.

### DIFF
--- a/formal/codegen-ptx/PtxKernelSpec.ml
+++ b/formal/codegen-ptx/PtxKernelSpec.ml
@@ -98,6 +98,7 @@ let ex_st = {regs = ex_empty_regs; tc = ex_zero_tc; mem = ex_zero_mem}
 
 let ex_k_no_shared =
   {
+    default_kernel with
     kern_name =
       String
         ( Ascii (false, true, true, true, false, true, true, false),
@@ -151,7 +152,6 @@ let ex_k_no_shared =
                                               true,
                                               false ),
                                           EmptyString ) ) ) ) ) ) ) ) );
-    kern_params = [];
     kern_shared = [];
     kern_body = ISEmpty;
   }
@@ -1138,6 +1138,7 @@ let ex_shared_decl =
 
 let ex_k_with_shared =
   {
+    default_kernel with
     kern_name =
       String
         ( Ascii (true, true, true, false, true, true, true, false),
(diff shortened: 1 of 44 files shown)


mathiasbourgoin:
@coderabbitai review

mathiasbourgoin:
@coderabbitai review

PR #364, opened 2026-07-29
fix(ci): the anti-vacuity gate covered one hardcoded path; widen it, with a declaration

mathiasbourgoin:
`scripts/check-test-alias-coverage.sh` guards against the *vacuous test* shape — an
`(executable)` with no run rule builds green and never executes. It read **one hardcoded
path**, `sarek/tests/e2e/dune`. So a gate against unwired tests was itself unwired
everywhere else in the tree.

## Measured on widening — and the composition is the result, not the count

**45 unwired out of 275 declared.** The backlog item said 30; that was wrong, and the
breakdown matters more than either number:

| | | |
|---|---|---|
| 27 | benchmarks | GPU timing loops, run by `run_all_benchmarks.sh` |
| 8 | browser targets | js_of_ocaml / WebGPU spikes, not native executables |
| 4 | hardware probes | report what a device advertises; no device in CI |
| 1 | tool | `device_info`, an operator tool, not a test |
| **5** | **real** | the ones that mattered |

The five: `test_transpile_proof`, `test_native_runtime`, `test_runtime_comparison`,
`gate_numeric`, `gate_framework`. **All wired now**, and all five verified passing
**before** wiring so this does not turn CI red.

**Two of them are gates that nothing invoked** — the exact shape this repo keeps
closing, found by widening a gate against exactly that. And `test_transpile_proof` is
the **only test of the transpiler** — the component #356 found routing all five
backends through a lossy emit path that dropped record and variant typedefs. It had
never run.

## Why the declaration ships in the same commit

A gate reporting 45 red on 39 non-problems gets switched off. So the widening comes
with `scripts/unwired-targets.tsv`: each legitimately-unwired target states *why*, with
a tag from a **closed set** (`benchmark | browser-target | hardware-probe | tool`) so
"TODO" or "flaky" cannot become an exemption.

**Both directions are red**, and the second is what stops the file rotting:

```
unwired with no declaration            -> UNWIRED
a declaration for something now wired  -> STALE
```

A stale exemption silently keeps covering whatever takes that name next — the same
declared-but-unverified shape the declaration exists to close.

## Root and declaration are arguments, and that is forced

This gate's input is the *whole tree*, so prove-red's copy-a-few-files scratch world
left most dune files absent and **every declaration for them read STALE** — a baseline
red for a reason unrelated to the property. It now runs against a fixture carrying a
wired target **and** a legitimately-unwired declared one, so *"refuses an undeclared
target"* and *"refuses every unwired target"* cannot be confused. Defaults are the real
tree, which is what CI runs.

## Verification

- **Six prove-red mutations**, one per refusal and one per direction:
  `unwired-undeclared`, `stale-declaration`, `bad-reason-tag`,
  `malformed-declaration-line`, `missing-declaration-file`, `empty-declared-set`.
  prove-red now **4 subjects / 19 mutations**, every one red as declared.
- prove-red also caught that `scripts/unwired-targets.tsv` was **untracked** — absent on
  a fresh clone, so the gate would pass here and exit 2 in CI.
- `dune test --force` **1725 cases / 119 suites / 0 FAIL**, and the five newly wired
  targets verified to have **run** by finding their own output in the suite log rather
  than assuming the rules fired.

## Scope

This is #69's hygiene half (A/B), which the audit recommended doing now. It does **not**
certify feature coverage — #69's actual subject — and it deliberately does not wire the
22 benchmarks to `runtest`: that would put device-dependent timing loops on a CPU-only
runner, which the audit flagged as the plausible-but-wrong version of this fix.

## Summary by CodeRabbit

* **Tests**
  * Added runtime test execution for native and CUDA comparison tests when CUDA is available.
  * Added test-run coverage for FFI-free gate checks and transpiler proof validation.
  * Expanded automated checks to detect unwired, stale, or invalid test-alias declarations.
  * Added fixtures covering wired and intentionally exempt test targets.

### DIFF
--- a/sarek/core/ffi_free_gate/dune
+++ b/sarek/core/ffi_free_gate/dune
@@ -16,3 +16,11 @@
  (modules gate_numeric)
  (libraries spoc_core_base stub_ops)
  (preprocess no_preprocessing))
+
+; backlog-69: WIRED. A GATE that nothing invoked — the exact shape this repo
+; keeps closing. Passing when wired, verified before wiring.
+
+(rule
+ (alias runtest)
+ (action
+  (run %{dep:gate_numeric.exe})))
--- a/sarek/tests/new_runtime/dune
+++ b/sarek/tests/new_runtime/dune
@@ -31,3 +31,20 @@
  ; broke `dune build` on machines with CUDA_PATH set but libnvrtc absent.
  (flags
   (:standard -w -33)))
+
+; backlog-69: WIRED. Both were declared as executables with no run rule, so
+; they built green and never executed. Both pass, verified before wiring.
+; test_runtime_comparison keeps its `enabled_if CUDA_PATH` guard, so dune skips
+; it entirely on a runner without CUDA rather than reporting a skip as a pass.
+
+(rule
+ (alias runtest)
+ (action
+  (run %{dep:test_native_runtime.exe})))
+
+(rule
+ (alias runtest)
+ (enabled_if
+  (<> %{env:CUDA_PATH=} ""))
+ (action
+  (run %{dep:test_runtime_comparison.exe})))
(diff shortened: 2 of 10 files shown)


mathiasbourgoin:
@coderabbitai review

PR #368, opened 2026-07-29
test(lower-ir): pin the trigger conditions of both ppx refusals

mathiasbourgoin:
backlog-165. `Sarek_lower_ir` refuses two module-constant-in-helper shapes with a located `raise_errorf`, and **neither was pinned by anything**.

Both are ppx-time errors and this tree has no negative-compile infrastructure — no "this `.ml` must FAIL to build, with this message" rule. So each refusal was code that had been observed firing *exactly once, by hand*, during a mutation run. A refactor making either unreachable would have been invisible: the same gate-that-cannot-fail shape the refusals themselves guard against.

## The choice

Test the **predicates that decide to refuse** rather than build negative-compile plumbing:

- the decision is where the history of being wrong is — `expr_barrier`'s first version returned `false` unconditionally, a guard that could not fire;
- `sarek_frontend` is a plain `(wrapped false)` library with no `.mli`, so the collectors are directly reachable, and `sarek/tests/e2e/test_stdlib_meta_proof` already links it the same way.

**The boundary is stated in the test header, not left for a reader to over-read:**

| | |
|---|---|
| covered | the decision — a predicate that stops discriminating goes red here |
| **not** covered | that a positive decision reaches the user *as a located error*. Still needs a negative-compile rule. |

## Coverage — 11 cases

Each barrier intrinsic asserted **individually, not as a set**, so dropping one arm fails on that arm and names it. A nested-expression case, because an initializer is a tree and a top-level-only check would miss the realistic shape. Two negative controls (a benign intrinsic; a barrier-free expression) — without them "detects a barrier" and "reports everything as a barrier" are the same observation, and the second would make module constants unusable rather than safe.

## Proven red four ways

Each mutation hits distinct, correctly-named cases:

| mutation | result |
|---|---|
| empty the synchronising-intrinsic list | 5 failures (4 arms + nested) |
| drop the `memory_fence_device` arm | exactly that arm, by name |
| free-name collector reverts to over-approximating | the shadow-only discriminator |
| `stmt_binders` stops recording `SLet` | the rebinding-collision case |

Wired into `sarek/tests/unit/runtest` (verified: the alias-coverage gate reports 278 targets / 40 declared, and the suite log contains the new output).

### DIFF
--- a/sarek/tests/unit/dune
+++ b/sarek/tests/unit/dune
@@ -257,3 +257,18 @@
   alcotest
   str
   unix))
+
+; backlog-165: the TRIGGER CONDITIONS of Sarek_lower_ir's two ppx-time refusals
+; (barrier-bearing constant initializer; a helper that both references and
+; rebinds a constant name). Both refusals are `raise_errorf`s and this tree has
+; no negative-compile infrastructure, so before this test each was code observed
+; firing exactly once by hand. This pins the PREDICATES that decide to refuse,
+; not the raise — see the test's header for that boundary stated explicitly.
+;
+; Links sarek_frontend directly (a plain wrapped-false library with no .mli),
+; the same way sarek/tests/e2e/test_stdlib_meta_proof does.
+
+(test
+ (name test_lower_ir_refusal_triggers)
+ (modules test_lower_ir_refusal_triggers)
+ (libraries sarek_frontend))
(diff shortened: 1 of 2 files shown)


mathiasbourgoin:
@coderabbitai review

PR #370, opened 2026-07-29
feat(ci): a resolved CodeRabbit finding must carry a test, or a recorded escape

mathiasbourgoin:
backlog-164. CodeRabbit found something real on **six consecutive pull requests in one day** whose CI was fully green. Each got fixed; whether a regression test came with the fix was decided by whoever happened to be writing it. That's a habit — and a habit that has to hold on the twentieth tired iteration is exactly what this repo replaces with a check.

## What made it checkable

Measured, not assumed — this is why the item sat open:

- `isResolved` **is** readable, **GraphQL only** (REST does not expose it).
- `resolvedBy.login` is `coderabbitai[bot]` on every resolved thread inspected — so resolution is **third-party verification** that the fix landed, not the author's own say-so. If it were self-attested this gate would be worthless.
- resolution is **not** a side effect of outdating: threads exist that are resolved-and-not-outdated, and unresolved-and-not-outdated.
- each finding body carries a stable `cr-comment:v1:<hex>` id — the escape key. **Line numbers are unusable**: `line` comes back `null`.

## The ceiling, stated in the header rather than implied

This asserts the PR **touched a test path**. That is weaker than *"a test that fails without the fix"*, and touching a comment in a test file satisfies it. It is a **floor** against fix-with-no-test-at-all, not proof of coverage, and must not be reported as though it were. The strong form is prove-red, which needs a per-finding red command.

## Fail-closed, because every failure mode here is "verified nothing"

| input | exit |
|---|---|
| no `reviewThreads` array (auth failure, wrong PR, schema change) | **2**, never 0 |
| unparseable data | **2** |
| malformed or blank-field escape | **2** — not a silent excuse |
| zero findings | 0, but prints `NOTHING TO VERIFY` so a clean line can't be misread |

Scope is deliberately narrow: only CodeRabbit's own findings, resolved by CodeRabbit. A **human**-resolved thread is self-attestation; an **unresolved** one isn't yet a fix (the review is still blocking it).

## Coverage — 24 offline cases

Fixtures via `--threads`/`--files`, so CI needs no token. Each test-path shape asserted individually, plus a control that `latest_results.ml` and `contest.ml` do **not** count as test paths.

## My own test was too weak, and the mutation caught it

Disabling the refusal made the code throw on an undefined escape — and **a crash also exits 1**, so every status-only case still "passed". 1 failure where there should have been 4; only the id-naming assertion noticed. Refusal cases now assert status 1 **and** the `REFUSED:` marker **and** the id. Re-mutated: 4 failures, each reporting `status=1 refused=false`.

## Wiring

Its own `pull_request`-guarded CI step, with `if:` rather than a silent no-op — so a push build does not *look* like it ran the gate and found nothing. The covering test runs on every event.

Verified live against merged PRs: #367 (1 finding, 3 test paths → OK) and #362 (2 findings, 2 test paths → OK).

## Summary by CodeRabbit

* **New Features**
  * Added automated verification for resolved review findings, requiring either a related test change or a documented exception.
  * Added support for validating findings from pull requests and reporting clear pass, refusal, or error outcomes.

* **Tests**
  * Added comprehensive offline coverage for test-path detection, exceptions, malformed data, missing findings, and error handling.


mathiasbourgoin:
## The pagination finding was real — and the *other* unbounded call was already misfiring today

Rebased onto `origin/main` first.

### CONFIRMED — `reviewThreads(first:100)` truncated silently

Two pieces of evidence, because "it could truncate" and "it fails open when it does" are different claims:

- The 100 ceiling is hard rather than a convention: `reviewThreads(first:101)` is rejected with `EXCESSIVE_PAGINATION` — *"Requesting 101 records on the `reviewThreads` connection exceeds the `first` limit of 100 records."* Past thread 100 there is no single-page read to be had at all.
- The fail-open is reproduced, not inferred. With a stub `gh` whose page 1 holds 100 out-of-scope threads and page 2 holds the only resolved finding, the pre-fix subject **exited 0 and printed `NOTHING TO VERIFY`**.

**Stated limit, because it changes what this is:** no Sarek PR has yet exceeded 100 review threads. Sweeping the last 100 PRs by `reviewThreads.totalCount`, the maximum is **11** (PR #305). So the thread truncation is *confirmed-latent* — the API cap and the stub prove the behaviour, but it has not yet misfired on this repo's history.

### Fixed by actually paginating, not by detecting truncation

The suggested diff detects `hasNextPage` and exits 2. That closes the fail-open but makes the gate unusable on the first PR that crosses 100 threads, so this follows the cursor instead: `pageInfo.hasNextPage`/`endCursor` with `after:`, accumulating until the connection is exhausted. Fail-closed (exit 2, "NOTHING was verified") on every way the walk can fail to complete:

- a page that errors — and it names the cursor it died after;
- missing or malformed `pageInfo`;
- `hasNextPage:true` with a null or absent `endCursor`;
- **a cursor that does not advance** (not in the finding — a repeated `endCursor` would otherwise spin forever);
- a 50-page budget, so an unbounded walk ends as a refusal rather than a hang.

The empty-input trap now applies **per page** and says which page, because "the connection ended here" and "page 2 came back empty" must not print the same line.

Offline `--threads` is documented and enforced rather than left ambiguous: a fixture is one GraphQL document, i.e. one page, and there is nowhere to fetch page 2 from. So a fixture must carry a `pageInfo`, one declaring `hasNextPage:true` exits 2 instead of reporting on page 1, and one with **no** `pageInfo` also exits 2 — it is indistinguishable from a truncated one.

### The same audit found a second instance, and that one is live today

| Call | Verdict | Evidence |
|---|---|---|
| `fetchFiles` → `gh pr view --json files` | **CONFIRMED, misfiring now** — fixed | Measured on PR #388, whose diff is 294 files: `gh pr view 388 --json files --jq '.files[].path'` prints **100**; `gh api .../pulls/388/files --paginate --jq '.[].filename'` prints **294**, a strict superset. Now uses the paginated REST endpoint; the live re-run reports `294 changed file(s)`. |
| `comments(first:1)` | **not a bug — intentional** | Only the first comment is CodeRabbit's finding body (author login + `cr-comment` id); replies are human discussion and are never read, so there is nothing on a second page for this gate to miss. Left alone, with a comment saying not to "fix" it into a cursor loop. |

The file-list truncation is the more dangerous of the two, and in the direction that is easy to miss: dropping the one test path in the tail of a diff turns an acceptable PR into a `REFUSED` one, and makes the verdict depend on where in the diff a file happens to sort.

### A defect introduced and caught, recorded because the shape of it matters

The first cut used `--jq '.[].path'`. The REST representation names that field **`filename`**, so it emitted one *blank* line per file — and `wc -l` still said 294. The initial "294 verified" was counting blank lines. What caught it was the live zero-files trap (exit 2), which had **no test**; it has one now. Blind compare: the count agreed while every value was empty.

### Prove-red

**Fixed hunk reverted** with `git checkout c7b5cba5 -- scripts/check-review-finding-tests.js` (not `git stash`) — exit **1**, 19 failures:

```
  FAIL: a finding found only on PAGE 2 is seen -- status=0 refused=false names-id=false
  FAIL:   and the walk did not stop at page 1 -- expected false, got true
  FAIL: incomplete pagination fails closed: a page-2 query that FAILS is exit 2 -- expected 2, got 0
  FAIL: incomplete pagination fails closed: hasNextPage:true with a NULL endCursor is exit 2 -- expected 2, got 0
  FAIL: incomplete pagination fails closed: a cursor that does not advance is exit 2 -- expected 2, got 0
  FAIL: the page budget is a refusal, not a loop or a verdict -- expected 2, got 0
  FAIL: a fixture declaring hasNextPage:true is exit 2, not a verdict on page 1 -- expected 2, got 1
passed: 27   failed: 19
```

`status=0` on the first line is the whole point: the pre-fix gate passed clean over a page-2 finding. Restored → exit 0.

**A subject that always exits 0** (`process.exit(0)` and nothing else) — exit **1**, 33 failures:

```
  FAIL: a document with NO pageInfo is exit 2 -- expected 2, got 0
  FAIL: no PR number is a usage error -- expected 2, got 0
  FAIL: --threads without --files is a usage error -- expected 2, got 0
passed: 13   failed: 33
```

The 13 that still read green are the exit-0-expecting cases (out-of-scope threads, recognised test paths); every refusal and fail-closed case rejects the stub. Per-case `says` regexes rather than one shared "NOTHING was verified" match, so no assertion is wider than the branch it covers.

### Gates

Baseline before editing was green on all of these, so nothing pre-existing is being inherited or hidden.

`dune build` 0 · `dune runtest` 0 · `dune build @fmt` 0 · `check-license-headers.sh` 0 · `check-kb-properties.sh` 0 · `prove-red.sh` 0 (6 subjects, 28 mutations) · `check-review-finding-tests.test.js` 0 — **46 pass, 0 fail** (was 24). Live end-to-end: PR #388 exit 0, PR #370 exit 0.

No new gate, so no new `kb/properties.md` row — the cases extend the existing `KB-GATE-REVIEW-FINDING-TESTS` red path, already invoked from `ci.yml`. Its row now records the pagination hole and the measured file-list cap.

### One thing to know: this branch was force-updated

`git push` was rejected non-fast-forward — the remote held `c8fe714c`/`7cd7dffe`, the **un-rebased twins** of these same two commits on the older `d9cbed3b` base. Verified before acting: `origin/main..origin/<branch>` contained only those two, and the gate script blob at `c8fe714c` is byte-identical (`e08775b0`) to the one at the rebased `c7b5cba5`. No third-party work existed to lose. Pushed with `--force-with-lease` so it would have aborted if anything had moved.

PR #373, opened 2026-07-29
fix(license): name the file on a duplicate copyright line instead of dying in sed

mathiasbourgoin:
Found while fixing the SPDX headers in #371, and it cost me real time there.

## The defect

The year-update path in `scripts/add-license-headers.sh` extracts the contributor's copyright line with a grep that **can match more than one line**. A header carrying two `SPDX-FileCopyrightText` lines for the same email — the exact state the case-insensitivity bug documented just above it used to produce, and which **four files in this repo were in** — makes that a multi-line value, which then reaches:

```sh
sed -i "s/$existing_years\(.*$contributor_email\)/.../I"
```

as the pattern. That is not a well-formed `s` command:

```
sed: -e expression n°1, caractère 6 : commande « s » inachevée
```

The fixer stopped mid-walk having silently skipped every remaining file, exited 1, and **never printed the offending path**. A red run with no location.

## The fix

A duplicate line is itself the defect to fix, so this refuses it by path rather than papering over it by taking the first match:

- `DUPLICATE` names the file *and* the email, on stderr
- `DUPLICATE_COUNT` is surfaced in the summary
- exit code **2**, distinct from 1 which already means "files need updating"
- the years extraction also takes `head -1`, so a second 4-digit match on a single line cannot reintroduce the same shape

## Scope, honestly

This is the **operator-facing fixer, not the CI gate**. `ci.yml:84` runs `check-license-headers.sh`, which was green throughout — the blast radius was developer time, not a broken build.

## Coverage

Case 10 in `scripts/check-license-headers.test.sh`, which CI *does* run (`ci.yml:168`). It asserts four things, because **the exit status alone was satisfied by the old crash too**: exit 2, the output names the file, it says `DUPLICATE`, and `unterminated` is absent.

Prove-red — removing the guard and restoring the un-`head`ed extraction:
```
FAIL: case10: duplicate copyright line -- expected exit 2, got 1
11 passed, 1 failed
```
With the fix: **12 passed, 0 failed**, and the CI gate stays green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved license header validation to detect duplicate copyright lines reliably.
  * Reports the affected file and duplicate count without triggering misleading errors.
  * License checks now fail consistently when duplicate entries are found, including check-only runs.

* **Tests**
  * Added coverage for duplicate OCaml copyright headers, including expected error codes and diagnostic output.

### DIFF
--- a/scripts/add-license-headers.sh
+++ b/scripts/add-license-headers.sh
@@ -35,6 +35,7 @@ MAINTAINER="Mathias Bourgoin <mathias.bourgoin@gmail.com>"
 
 # Colors
 GREEN='\033[0;32m'
+RED='\033[0;31m'
 YELLOW='\033[1;33m'
 BLUE='\033[0;34m'
 NC='\033[0m'
@@ -42,6 +43,10 @@ NC='\033[0m'
 # Counters
 UPDATED_COUNT=0
 SKIPPED_COUNT=0
+# Files carrying two copyright lines for the same contributor. Not fixable
+# here (this script cannot choose which to keep), so it reports and exits
+# non-zero rather than feeding a multi-line value to sed and dying namelessly.
+DUPLICATE_COUNT=0
 
 # Get copyright info from git
 get_copyright_years() {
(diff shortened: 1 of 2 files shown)


mathiasbourgoin:
Fixed, and the finding was right — it also lands on the duplicate counter this PR adds, which I had not seen.

**Impact, worst first:** a *different* contributor whose address differs only where the dots sit satisfied `grep -i "$contributor_email"`, so the fixer believed the maintainer already had a line, took the year-update branch, and **never added the maintainer's own copyright** — a silently missing attribution. With two such lines present, the new counter would then report a false `DUPLICATE` and exit 2 on a correct header. An address containing `[ ] * ^ $ \` would be a grep *error*, not merely loose.

All three greps now use `-F` on the **bracketed** address, which also pins the match to a whole address rather than a substring of a longer one. `-i` is kept — it is load-bearing and documented (providers are case-insensitive by spec, and case-sensitivity was the original cause of the duplicate lines).

**One thing beyond the three greps you named:** the same variable was also interpolated into two `sed` **patterns**, with the identical defect. Those are gone too — not by escaping metacharacters, but by removing regex from that path entirely. The line is already located, so the year is substituted with bash `${var/pat/repl}` (glob, not regex; the year is digits and a dash) and the line is rewritten by *number* with `awk`. Carrying the rest of the line verbatim also preserves the address's casing, which the old code needed a `\1` backreference for.

**Three cases added** — the year-update path I rewrote had **no case at all**, so the suite read 12/12 green over it and refactoring it was indistinguishable from breaking it:

- `case11` — stale single year becomes a span, address intact
- `case12` — a regex-matching but *different* address is not this contributor, asserted on the **outcome** (the maintainer's address must end up in the file) so it holds however the branch is worded
- `case13` — the opposite polarity: same address, different casing, is still the same person and appends no second line, pinning that `-F` did not eat `-i`

**case12 was wrong on the first cut and prove-red caught it.** I had replaced the `@` as well as the dots — but `@` is *literal* in the regex, so that counterexample never matched the buggy pattern and the case passed against the very defect it claimed to detect. Only the dots may differ. Verified after correcting it: with the greps reverted to regex matching, `case12: the maintainer's own copyright line was never added` — 14 passed, 1 failed. With the fix: **15 passed, 0 failed**, and the CI gate stays green.

PR #375, opened 2026-07-29
feat(soa): carry an SoA binding on the vector, not a host_storage constructor

mathiasbourgoin:
**backlog-54, first slice of the transparency work — and it corrects the persisted plan, which is not buildable as written.**

## The plan's design violates a gated invariant

The Tier 1b handoff's §1 says to add a `Custom_storage_soa` constructor to the `host_storage` GADT in `spoc_core_base`, holding the plan and N leaf sub-buffers. Two facts the doc doesn't record:

- **`spoc_core_base` is deliberately FFI-free** — its dune stanza must not list `ctypes`/`unix`/`spoc_framework` — and it is *additionally* compiled to `.bc.js`. `sarek/core/ffi_free_gate` enforces **both** at build time (`(modes byte js)`).
- **`Soa` uses `Ctypes` directly** to copy raw words, and both it and `Vector_types` live in `spoc_core`, *above* that layer.

So naming `Soa.plan` down there either breaks that gate or forces the plan representation to be duplicated across a layer boundary — the drift hazard the `ir_fields` trust note warns about, and probably the real reason the earlier work chose the layered `Soa_vector` in the first place.

## Resolved with behaviour instead of types

`soa_binding` carries the transpose as two closures plus the only two numbers a launch needs (leaf count, AoS stride). Nothing ctypes-shaped crosses down, and the value is one a jsoo build can hold. The producer is `Vector.create` in `spoc_core`, which may use ctypes freely.

## A record field, not a GADT constructor — measured, not guessed

| Approach | Forced edits |
|---|---|
| new `host_storage` constructor | **51 match arms across 5 files** (`Spoc_core_base` 16, `Vector` 14, `Interpreter_plugin_base` 8, `Native_plugin_base` 7, `Vector_jsx` 6) |
| record field | **6 sites**, all in `Spoc_core_base.ml` |

The field keeps the property that made the GADT attractive — the compiler finds every site — without the fan-out, and without a hidden global side table keyed by vector id. The doc's "~25 forced match sites" estimate was low by 2×.

`soa = None` on the copy/sub-vector paths is not a placeholder: the handoff keeps `of_ctypes_ptr` and `sub_vector` AoS-only, so `None` is the correct value there.

## Scope of this PR

**No behaviour change — nothing reads the field yet.** Host access is untouched *by construction*, which is why `get`/`set` and the PPX accessors needed no edits at all. This is the foundation plus the corrected design record.

Still to come on this branch: `Vector.create ~layout:SoA` populating the binding; the two device-facing closures the launch needs (`Ops.device_t -> Ops.device_buf list` — expressible because `core_base` already parameterises over `Ops`); `Execute.run` auto-dispatch on CUDA/PTX with the AoS fallback elsewhere; then extending `test_soa_emitter_equiv` through the transparent path.

Build green, `@fmt` green, **`ffi_free_gate` builds both `byte` and `js`** (the invariant this design exists to preserve), **1729 cases across 119 suites, 0 failures**.

## Summary by CodeRabbit

* **New Features**
  * Added transparent SoA vectors that work with standard execution flows.
  * Added support for SoA data across device transfers and migrations.
* **Bug Fixes**
  * Improved device-to-host read-back and stale-data handling.
  * Fixed cleanup, freeing, and cross-device behavior to prevent incorrect data reuse.
* **Tests**
  * Expanded coverage for mixed-width records, relaunches, round trips, and cross-device migration.
* **Documentation**
  * Updated correctness evidence with completed mixed-width SoA device results.

### DIFF
--- a/docs/optimization/tier1b-emitter-soa-handoff.md
+++ b/docs/optimization/tier1b-emitter-soa-handoff.md
@@ -65,8 +65,10 @@ existing enumeration — the STOP condition in the task did not fire).
   reference. **PASS** for `point3d` (3× f32) and `dpair` (2× f64) on RX 7900 XTX
   and CPU-as-CUDA under ZLUDA; `point3d`'s AoS leg also passes on
   OpenCL/Vulkan/Native/Interpreter. i32/i64 leaves are covered at the
-  PTX-instruction + ptxas-assembly level (device e2e for them is folded into
-  Tier 1c once the launch plumbing lands).
+  PTX-instruction + ptxas-assembly level AND on device — see the table below and
+  "Device e2e for the two integer combos: DONE" under it. The Tier 1c deferral
+  this bullet used to carry was discharged once ZLUDA supplied a CUDA/PTX
+  device.
 
 | Leaf type combo | PTX markers | ptxas | device e2e (ZLUDA) |
 |---|---|---|---|
@@ -198,7 +200,10 @@ codegen-level `~soa_params` knob:
 This is pure plumbing (no new coalescing) — the roadmap (§1.2c/e) rates it the
 bulk of the SoA cost; it was descoped here so the emitter could ship complete
 and proven. When it lands, extend `test_soa_emitter_equiv` to drive SoA through
-`Vector.create ~layout:SoA` + `run_vectors`, and add the i32/i64 device rows.
+the transparent constructor + `run_vectors`. The i32/i64 device rows this
+sentence used to also ask for are NOT outstanding: they landed with
+`check_mixed_widths` (see the table above), and leaving them listed here had a
+reader draw the opposite conclusion from the table.
 
 > **✅ CLOSED (2026-07-29) — was: PRECONDITION, generated param-name namespace.**
 > Option (a) shipped in `63ab6df1`: the emitter now mangles leaves as
--- a/docs/optimization/tier1b-emitter-soa-handoff.md
+++ b/docs/optimization/tier1b-emitter-soa-handoff.md
@@ -72,13 +72,28 @@ existing enumeration — the STOP condition in the task did not fire).
 |---|---|---|---|
 | f32 × 3 (`point3d`) | ✓ | ✓ | ✓ |
 | f64 × 2 (`dpair`) | ✓ | ✓ | ✓ |
-| i32 + f64 (`{i;d}`) | ✓ | ✓ (`soa_mixed`) | (Tier 1c) |
-| i64 + i32 (`{p;q}`) | ✓ (`ld.global.s64`+`s32`) | ✓ (`soa_long`) | (Tier 1c) |
+| i32 + f64 (`{i;d}`) | ✓ | ✓ (`soa_mixed`) | ✓ (2026-07-30, ZLUDA) |
+| i64 + i32 (`{p;q}`) | ✓ (`ld.global.s64`+`s32`) | ✓ (`soa_long`) | ✓ (2026-07-30, ZLUDA) |
 
 All three SoA kernels (`soa_field_sum_f32`, `soa_mixed_i32_f64`,
 `soa_long_i64_i32`) go through `test_ptxas_assembles`; the i32+f64 and i64+i32
-combos additionally have dedicated marker tests. Device e2e for the two integer
-combos folds into Tier 1c with the launch plumbing.
+combos additionally have dedicated marker tests.
+
+**Device e2e for the two integer combos: DONE 2026-07-30.** They were open only
+for want of a CUDA/PTX device; ZLUDA on the RX 7900 XTX provides one.
+`check_mixed_widths` in `test_soa_emitter_equiv.ml` drives both through the
+TRANSPARENT path (`Soa_vector.create_transparent` + the generic
+`Execute.run_vectors`), so one case covers the mixed-width leaf addressing and
+the generic dispatch together. Each leaf is written to its own output array at
+its own width — deliberately no int<->float conversion, since a conversion folds
+both leaves into one number and a per-leaf stride error could then be masked by a
+compensating error in the other leaf.
+
+Observed: both rows OK on the N-leaf PTX ABI (2 ZLUDA devices) and OK on the AoS
+fallback (5 non-PTX devices); the i32+f64 row SKIPs on the 2 devices that report
+no fp64. The rows are gated on the DEVICE CAPABILITY (`Device.allows_fp64` /
+`allows_int64`) rather than on `framework = "CUDA/PTX"` the way the dpair row is,
+so they run wherever they can actually execute instead of only on PTX.
 
 ## Benchmark — the emitter's coalescing win (`benchmarks/bench_soa_emitter`)
 
--- a/docs/optimization/tier1b-emitter-soa-handoff.md
+++ b/docs/optimization/tier1b-emitter-soa-handoff.md
@@ -89,9 +89,12 @@ its own width — deliberately no int<->float conversion, since a conversion fol
 both leaves into one number and a per-leaf stride error could then be masked by a
 compensating error in the other leaf.
 
-Observed: both rows OK on the N-leaf PTX ABI (2 ZLUDA devices) and OK on the AoS
-fallback (5 non-PTX devices); the i32+f64 row SKIPs on the 2 devices that report
-no fp64. The rows are gated on the DEVICE CAPABILITY (`Device.allows_fp64` /
+Observed: both rows OK on the N-leaf PTX ABI (2 ZLUDA devices). On the AoS
+fallback the two rows differ and are counted separately — the earlier single "5
+non-PTX devices" figure was the i32+f64 count applied to both: the **i64+i32** row
+is OK on **all 7** non-PTX devices, while the **i32+f64** row is OK on **5** and
+SKIPs on the 2 rusticl OpenCL devices that do not report fp64. The rows are gated
+on the DEVICE CAPABILITY (`Device.allows_fp64` /
 `allows_int64`) rather than on `framework = "CUDA/PTX"` the way the dpair row is,
 so they run wherever they can actually execute instead of only on PTX.
 
--- a/sarek/core/Soa_vector.ml
+++ b/sarek/core/Soa_vector.ml
@@ -236,15 +236,31 @@ let create_transparent (custom : 'a Vector.custom_type) (length : int) :
                device the results were NOT on discarded them.
 
                One rule for both cases, so there is no [None] special case to
-               drift: after freeing, the leaves are authoritative iff some leaf
-               still has a device buffer somewhere. For [None] every leaf was
-               just released on every device, so this evaluates to [false] — and
-               it does so by observing the leaves rather than by asserting what
-               the loop above was supposed to have done. *)
+               drift: after freeing, the leaves are authoritative iff they
+               already WERE and some leaf still has a device buffer somewhere.
+               For [None] every leaf was just released on every device, so the
+               second conjunct is [false] — and it is so by observing the leaves
+               rather than by asserting what the loop above was supposed to have
+               done.
+
+               The first conjunct is not redundant, and leaving it out was a
+               resurrection bug: allocation is not authority. A packed launch
+               CLEARS this flag after gathering the leaves into host storage
+               (Execute.transfer_vectors_to_device), and the leaves it gathered
+               stay allocated on their device — so deriving from allocation alone
+               turned the flag back ON for leaves that now hold PRE-packed-launch
+               data. Measured (test_soa_cross_device_migration,
+               "a free does not resurrect stale leaf ownership"): SoA launch on
+               A, packed launch on B, [free_buffer sv B], packed launch again —
+               the last launch gathered A's stale leaves over the host copy of
+               B's result and ran on them. Narrowing only: this can clear the
+               flag, never set it. *)
             leaves_live :=
-              Array.exists
-                (fun (Leaf lv) -> Hashtbl.length lv.Vector.device_buffers > 0)
-                t.leaves);
+              !leaves_live
+              && Array.exists
+                   (fun (Leaf lv) ->
+                     Hashtbl.length lv.Vector.device_buffers > 0)
+                   t.leaves);
         soa_leaf_bufs =
           (fun dev -> Array.to_list (Array.map (leaf_buf dev) t.leaves));
       } ;
--- a/sarek/core/Soa_vector.ml
+++ b/sarek/core/Soa_vector.ml
@@ -94,7 +94,31 @@ let scatter t =
     t.plan
     ~aos:(Vector.to_ctypes_ptr t.aos)
     ~length:t.length
-    ~leaves:(leaf_ptrs t)
+    ~leaves:(leaf_ptrs t) ;
+  (* The transpose above writes each leaf's HOST buffer through a raw ctypes
+     pointer, which performs no location bookkeeping. So without this a leaf
+     keeps whatever location it already had — and after a previous launch that
+     is [Both dev], on which [Transfer.to_device] logs "skip (Both)" and returns
+     without copying. A SECOND launch on the same vector then ran against the
+     FIRST launch's device data, silently, with no user workaround.
+
+     Marking [Stale_GPU dev] is the accurate statement and not a nudge: the host
+     copy is the one just written and the device copy is now out of date. It is
+     also the one location [to_device] does not short-circuit.
+
+     Only leaves that actually HAVE a buffer on a device are touched; a leaf
+     never transferred stays [CPU], which is already correct.
+
+     Done here rather than in the [soa_to_device] closure so the explicit
+     [Soa_launch.run_soa] path gets it too — both callers scatter through this
+     function, and this invariant must not have two answers. *)
+  Array.iter
+    (fun (Leaf lv) ->
+      match lv.Vector.location with
+      | Vector.Both d | Vector.GPU d | Vector.Stale_CPU d ->
+          lv.Vector.location <- Vector.Stale_GPU d
+      | Vector.CPU | Vector.Stale_GPU _ -> ())
+    t.leaves
 
 let gather t =
   Soa.gather
--- a/sarek/core/Soa_vector.ml
+++ b/sarek/core/Soa_vector.ml
@@ -217,14 +217,34 @@ let create_transparent (custom : 'a Vector.custom_type) (length : int) :
                 Array.iter
                   (fun (Leaf lv) -> Transfer.free_all_buffers lv)
                   t.leaves) ;
-            (* The leaves now hold nothing a read-back could fetch, so the flag
+            (* The freed leaves hold nothing a read-back could fetch, so the flag
                that says "the leaves are authoritative" must stop saying it —
-               otherwise the next read-back follows freed buffers. Safe to clear
-               unconditionally even for the [Some d] case: this binding has one
-               flag for the whole vector, not one per device (see
-               [Transfer.has_device_data]), so there is no per-device answer to
-               preserve. *)
-            leaves_live := false);
+               otherwise the next read-back follows freed buffers.
+
+               DERIVED from what is still allocated, not assigned [false]. The
+               previous version cleared it unconditionally and justified that
+               with "this binding has one flag for the whole vector, so there is
+               no per-device answer to preserve". The premise is right and the
+               conclusion is backwards: precisely BECAUSE the flag covers the
+               whole vector, a per-device free ([Some d]) must not clear it while
+               leaves are still live on another device. Measured after a
+               migration — leaves on A, packed buffer on B — [free_buffer sv B]
+               released B's half and then reported the whole vector leaf-free
+               with A's leaves still allocated. [Transfer.has_device_data sv A]
+               then answered [false], so the drain-before-free in
+               [Transfer.free_buffer]/[free_all_buffers] skipped A: freeing a
+               device the results were NOT on discarded them.
+
+               One rule for both cases, so there is no [None] special case to
+               drift: after freeing, the leaves are authoritative iff some leaf
+               still has a device buffer somewhere. For [None] every leaf was
+               just released on every device, so this evaluates to [false] — and
+               it does so by observing the leaves rather than by asserting what
+               the loop above was supposed to have done. *)
+            leaves_live :=
+              Array.exists
+                (fun (Leaf lv) -> Hashtbl.length lv.Vector.device_buffers > 0)
+                t.leaves);
         soa_leaf_bufs =
           (fun dev -> Array.to_list (Array.map (leaf_buf dev) t.leaves));
       } ;
--- a/sarek/core/Soa_vector.ml
+++ b/sarek/core/Soa_vector.ml
@@ -102,3 +102,52 @@ let gather t =
     ~leaves:(leaf_ptrs t)
     ~length:t.length
     ~aos:(Vector.to_ctypes_ptr t.aos)
+
+(* ── Transparent SoA: a plain Vector the generic launch path can dispatch on ──
+   backlog-54. Returns the AoS vector with its [soa] binding populated, so
+   [Execute.run] can bind the N-leaf ABI without ever naming this module.
+
+   Why that indirection exists: sarek/execute/jsoo/dune copies Execute.ml but
+   NOT Soa_launch.ml, so Execute.ml is compiled in a build where Soa_vector does
+   not exist. It therefore cannot call into here — the binding's closures are the
+   only channel, which is why they are closures and not a plan.
+
+   Why the constructor is HERE and not [Vector.create ~layout:SoA] as the Tier 1b
+   handoff proposed: Soa_vector depends on Vector, so putting it there would be a
+   layer inversion. The transparency that the item is actually about is at the
+   LAUNCH site — the user calls the generic Execute.run and the N-pointer ABI is
+   selected for them — and that is unaffected by which module names the
+   constructor. *)
+let create_transparent (custom : 'a Vector.custom_type) (length : int) :
+    ('a, unit) Vector.t =
+  let t = create custom length in
+  let leaf_buf (dev : Device.t) (Leaf lv) =
+    match Vector.get_buffer lv dev with
+    | Some b -> b
+    | None ->
+        (* Unreachable via the launch path: soa_to_device runs first and
+           allocates every leaf. Loud rather than silent, because a missing leaf
+           buffer would otherwise bind a short argument list to an N-pointer
+           signature — the kernel would then read whatever followed. *)
+        invalid_arg
+          (Printf.sprintf
+             "Soa_vector: leaf buffer not allocated on device %d; \
+              soa_to_device must run before soa_leaf_bufs"
+             dev.Device.id)
+  in
+  let v = t.aos in
+  v.Vector.soa <-
+    Some
+      {
+        Vector.soa_num_leaves = num_leaves t;
+        soa_aos_stride = t.plan.Soa.aos_stride;
+        soa_scatter = (fun () -> scatter t);
+        soa_gather = (fun () -> gather t);
+        soa_to_device =
+          (fun dev ->
+            scatter t ;
+            Array.iter (fun (Leaf lv) -> Transfer.to_device lv dev) t.leaves);
+        soa_leaf_bufs =
+          (fun dev -> Array.to_list (Array.map (leaf_buf dev) t.leaves));
+      } ;
+  v
--- a/sarek/core/Soa_vector.ml
+++ b/sarek/core/Soa_vector.ml
@@ -160,11 +160,12 @@ let create_transparent (custom : 'a Vector.custom_type) (length : int) :
              dev.Device.id)
   in
   let v = t.aos in
-  (* SET only by [soa_to_device] below; cleared or narrowed by three other
-     writers outside this module (see [soa_leaves_live]'s doc in
-     Spoc_core_base.ml for the exhaustive list). No reader re-derives "SoA or
-     AoS?" from the vector's shape — that is the point of the flag, so that
-     read-back follows the launch's ABI decision. *)
+  (* SET only by [soa_to_device] below. Three other writers can clear or narrow
+     it: [soa_free_leaves], also in this record, narrows it; and
+     [Execute.transfer_vectors_to_device] and [Transfer.to_device] clear it. That
+     list is exhaustive — see [soa_leaves_live]'s doc in Spoc_core_base.ml. No
+     reader re-derives "SoA or AoS?" from the vector's shape; that is the point of
+     the flag, so that read-back follows the launch's ABI decision. *)
   let leaves_live = ref false in
   v.Vector.soa <-
     Some
--- a/sarek/core/Soa_vector.ml
+++ b/sarek/core/Soa_vector.ml
@@ -197,6 +197,34 @@ let create_transparent (custom : 'a Vector.custom_type) (length : int) :
               (fun (Leaf lv) -> Transfer.to_cpu ~force:true lv)
               t.leaves ;
             gather t);
+        soa_free_leaves =
+          (fun dev ->
+            (* No read-back here. The CALLER (Transfer.free_buffer /
+               free_all_buffers) drains through [read_back_to_host] first, and
+               doing it again here would either be redundant or — once the flag
+               below is cleared — read the packed buffer instead. One drain, at
+               the site that owns the ordering.
+
+               [Transfer.free_buffer]/[free_all_buffers] on a LEAF, not
+               [B.free ()] directly: the leaves are ordinary scalar vectors, so
+               they get the location bookkeeping and the [Gpu_memory.track_free]
+               accounting for free, and a leaf whose buffer is already gone is a
+               no-op rather than a double free. *)
+            (match dev with
+            | Some d ->
+                Array.iter (fun (Leaf lv) -> Transfer.free_buffer lv d) t.leaves
+            | None ->
+                Array.iter
+                  (fun (Leaf lv) -> Transfer.free_all_buffers lv)
+                  t.leaves) ;
+            (* The leaves now hold nothing a read-back could fetch, so the flag
+               that says "the leaves are authoritative" must stop saying it —
+               otherwise the next read-back follows freed buffers. Safe to clear
+               unconditionally even for the [Some d] case: this binding has one
+               flag for the whole vector, not one per device (see
+               [Transfer.has_device_data]), so there is no per-device answer to
+               preserve. *)
+            leaves_live := false);
         soa_leaf_bufs =
           (fun dev -> Array.to_list (Array.map (leaf_buf dev) t.leaves));
       } ;
--- a/sarek/core/Soa_vector.ml
+++ b/sarek/core/Soa_vector.ml
@@ -136,6 +136,9 @@ let create_transparent (custom : 'a Vector.custom_type) (length : int) :
              dev.Device.id)
   in
   let v = t.aos in
+  (* Set by [soa_to_device] below and only read by the read-back path, so that
+     read-back follows the launch's ABI decision rather than re-deriving it. *)
+  let leaves_live = ref false in
   v.Vector.soa <-
     Some
       {
@@ -143,10 +146,33 @@ let create_transparent (custom : 'a Vector.custom_type) (length : int) :
         soa_aos_stride = t.plan.Soa.aos_stride;
         soa_scatter = (fun () -> scatter t);
         soa_gather = (fun () -> gather t);
+        soa_leaves_live = leaves_live;
         soa_to_device =
           (fun dev ->
             scatter t ;
-            Array.iter (fun (Leaf lv) -> Transfer.to_device lv dev) t.leaves);
+            Array.iter (fun (Leaf lv) -> Transfer.to_device lv dev) t.leaves ;
+            leaves_live := true ;
+            (* The AoS vector itself gets no device buffer under this ABI — the
+               leaves hold the data — so without this its location stays [CPU]
+               and EVERY read-back path short-circuits: [Transfer.to_cpu],
+               [Transfer.sync] and the auto-sync callback all treat [CPU] as
+               "nothing on a device to fetch" and return before looking at the
+               SoA binding at all. Recording the device as authoritative and the
+               host copy as stale is not bookkeeping; it is what makes the
+               read-back reachable. *)
+            v.Vector.location <- Vector.Stale_CPU dev);
+        soa_from_device =
+          (fun _dev ->
+            (* [~force:true]: after a launch the leaves are [Vector.Both dev] —
+               the host copy is the pre-launch scatter, the device copy is the
+               result — and an unforced [to_cpu] treats [Both] as already in
+               sync and returns without reading anything back. That is exactly
+               the silent-stale-output failure this closure exists to prevent,
+               so the force is load-bearing and not defensive. *)
+            Array.iter
+              (fun (Leaf lv) -> Transfer.to_cpu ~force:true lv)
+              t.leaves ;
+            gather t);
         soa_leaf_bufs =
           (fun dev -> Array.to_list (Array.map (leaf_buf dev) t.leaves));
       } ;
--- a/sarek/core/Soa_vector.ml
+++ b/sarek/core/Soa_vector.ml
@@ -160,8 +160,11 @@ let create_transparent (custom : 'a Vector.custom_type) (length : int) :
              dev.Device.id)
   in
   let v = t.aos in
-  (* Set by [soa_to_device] below and only read by the read-back path, so that
-     read-back follows the launch's ABI decision rather than re-deriving it. *)
+  (* SET only by [soa_to_device] below; cleared or narrowed by three other
+     writers outside this module (see [soa_leaves_live]'s doc in
+     Spoc_core_base.ml for the exhaustive list). No reader re-derives "SoA or
+     AoS?" from the vector's shape — that is the point of the flag, so that
+     read-back follows the launch's ABI decision. *)
   let leaves_live = ref false in
   v.Vector.soa <-
     Some
--- a/sarek/core/Soa_vector.mli
+++ b/sarek/core/Soa_vector.mli
@@ -27,10 +27,17 @@
  * constructor on the core [host_storage] GADT. That much is unchanged; the
  * DEFERRAL this header used to describe is not.
  *
- * Generic [Execute.run] auto-dispatch is no longer deferred: {!create_transparent}
- * provides it, and a vector from it takes the N-leaf ABI through the ordinary
- * [Execute.run]/[run_vectors] with no SoA-specific entry point. Two earlier
- * halves are likewise closed — the leaf enumeration is derived from
+ * Auto-dispatch is no longer deferred: {!create_transparent} provides it, and a
+ * vector from it takes the N-leaf ABI through the ordinary
+ * {!Sarek.Execute.run_vectors} with no SoA-specific entry point. Named
+ * precisely, because bare [Execute.run] is NOT the same promise: [run_vectors]
+ * is what performs the leaf upload before dispatching. On CUDA/PTX — the only
+ * backend where the leaf ABI is selected at all — [run] called directly on a
+ * vector whose leaves were never transferred raises rather than launching, which
+ * is what [check_direct_run_refuses_untransferred] in test_soa_emitter_equiv.ml
+ * pins (and why that case carries the SoA-ABI blocker: on the CPU backends [run]
+ * executes against host vectors and there is no leaf upload to have skipped). Two earlier halves are likewise closed — the leaf
+ * enumeration is derived from
  * [custom_type.ir_fields] rather than supplied by the caller, and the launch
  * path threads the SoA param names itself.
  *
--- a/sarek/core/Soa_vector.mli
+++ b/sarek/core/Soa_vector.mli
@@ -24,16 +24,24 @@
  * only. This module is pure host storage + transpose and is backend-agnostic.
  *
  * This is deliberately layered above {!Vector} + {!Soa} rather than being a new
- * constructor on the core [host_storage] GADT. A fully transparent
- * [Vector.create ~layout:SoA] + generic [Execute.run] auto-dispatch would
- * require threading SoA param names through every backend and generalising the
- * 1-buffer-per-device table — deferred (see
- * docs/optimization/tier1b-emitter-soa-handoff.md and the impl brief).
+ * constructor on the core [host_storage] GADT. That much is unchanged; the
+ * DEFERRAL this header used to describe is not.
  *
- * The layout half of that deferral is now closed: the leaf enumeration is
- * derived from [custom_type.ir_fields] rather than supplied by the caller, so
- * "the custom_type carries no record layout" is no longer a reason this cannot
- * be transparent. What remains is the backend threading and the buffer table.
+ * Generic [Execute.run] auto-dispatch is no longer deferred: {!create_transparent}
+ * provides it, and a vector from it takes the N-leaf ABI through the ordinary
+ * [Execute.run]/[run_vectors] with no SoA-specific entry point. Two earlier
+ * halves are likewise closed — the leaf enumeration is derived from
+ * [custom_type.ir_fields] rather than supplied by the caller, and the launch
+ * path threads the SoA param names itself.
+ *
+ * What is still deferred, and the only thing this header now claims: the
+ * dispatch is CUDA/PTX-only. No other backend's emitter lowers a record
+ * parameter to N base pointers, so everywhere else the same vector stays an
+ * ordinary packed AoS custom vector and takes the packed path — the documented
+ * "never wrong data" fallback, not a gap that produces different results. There
+ * is also still no [Vector.create ~layout:SoA]: the constructor lives here (see
+ * {!create_transparent} for why that is not the same question as transparency at
+ * the launch site).
  ******************************************************************************)
 
 (** A per-leaf host buffer, type-erased: a width-matched scalar vector used as a
--- a/sarek/core/Soa_vector.mli
+++ b/sarek/core/Soa_vector.mli
@@ -93,3 +93,20 @@ val scatter : 'a t -> unit
 (** [gather t] transposes the N per-leaf host buffers back into the AoS host
     buffer (call after reading leaves back from a device that wrote them). *)
 val gather : 'a t -> unit
+
+(** [create_transparent custom length] is a SoA custom vector presented as an
+    ordinary {!Vector.t}: the generic {!Sarek.Execute.run_vectors} recognises it
+    and binds the N-leaf ABI on CUDA/PTX, falling back to the packed AoS path on
+    every other backend. Callers need no SoA-specific launch entry point
+    (backlog-54).
+
+    The returned vector IS the AoS source of truth — [Vector.get]/[set] and the
+    PPX accessors work on it exactly as on any custom vector, and the transpose
+    to and from the per-leaf buffers happens inside the launch.
+
+    Use this rather than {!create} + {!Sarek.Soa_launch.run_soa} unless you need
+    to drive the leaves yourself. {!create} remains the lower-level entry point.
+
+    Raises {!Soa.Unsupported} on an element type with no derivable flat-record
+    layout, exactly as {!create} does. *)
+val create_transparent : 'a Vector.custom_type -> int -> ('a, unit) Vector.t
--- a/sarek/core/Transfer.ml
+++ b/sarek/core/Transfer.ml
@@ -307,9 +307,10 @@ end = struct
       not a cosmetic asymmetry: [soa_free_leaves] used to CLEAR the flag on a
       per-device free, which made this function answer [false] for a device
       whose leaves were still live, and the drain-before-free in
-      {!free_all_buffers} consults exactly this answer. It now re-derives the
-      flag from the leaves that survive, so a per-device release cannot disown
-      another device's. *)
+      {!free_all_buffers} consults exactly this answer. It now NARROWS the flag
+      — still set iff it already was and some leaf survives — so a per-device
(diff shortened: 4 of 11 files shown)


mathiasbourgoin:
## Update: dispatch landed, and the PTX half is now *measured* (ZLUDA)

Two more commits on this branch. The transparency itself is in.

### `Execute.run` now selects the ABI

`Soa_vector.create_transparent` returns an ordinary `Vector.t` with its `soa` binding populated; `Execute.run_vectors` recognises it and binds **N leaf base pointers + one shared length** instead of a packed `(ptr, len)` pair. No SoA-specific launch entry point.

**One predicate, `soa_dispatch`, decides SoA vs AoS**, and all three touch points read it: the host→device transfer, the argument expansion, and the `~soa_params` handed to `generate_source`. That is deliberate — the emitted signature and the bound argument list must be two consequences of *one* decision. Two sites answering it separately is how you get N pointers bound to a packed AoS param block, which is not a crash but silently wrong data.

`~soa_params` needed **no threading work**: `Framework_sig` already declares it (`?soa_params:string list`, `Framework_sig.ml:360`) and the CUDA/PTX backend already honours it. The handoff doc still lists that as to-do — stale.

### The closure design is *required*, not just cheaper

`sarek/execute/jsoo/dune` copies `Execute.ml` but **not** `Soa_launch.ml` (modules: `Smoke Execute Execute_error`), so `Execute.ml` is compiled in a build where `Soa_vector` does not exist and can never name it. The binding's closures are the only channel. No GADT-carrying-`Soa.plan` design could put dispatch in `Execute.run` at all. **Verified: the jsoo Execute target builds with dispatch in.**

Also, `create_transparent` lives in `Soa_vector`, not as `Vector.create ~layout:SoA` — `Soa_vector` depends on `Vector`, so the doc's placement is a layer inversion. The transparency the item is about is at the *launch* site and is unaffected by which module names the constructor.

### Evidence — both halves, on real hardware

The new `check_transparent` case asserts the transparent result equals the **pure-OCaml reference** (not a second GPU run, so a bug common to both device paths cannot hide).

| Path | Devices | Result |
|---|---|---|
| PTX N-leaf ABI (**via ZLUDA**) | RX 7900 XTX, Ryzen 7950X iGPU | ✅ `transparent SoA == reference (PTX: N-leaf ABI) OK` |
| AoS fallback | OpenCL ×2, Vulkan ×2, Native, Interpreter | ✅ `transparent SoA == reference (non-PTX: AoS fallback) OK` |

I had reported the PTX half as unverifiable on this host; **ZLUDA makes it measurable, and it passes.** Asserting the *fallback* half matters too — it's the only thing distinguishing "the fallback works" from "the fallback was never exercised".

**Prove-red on the gate.** Dropping the `CUDA/PTX` condition from `soa_dispatch` so it fires everywhere:

```
Fatal error: exception [OpenCL Runtime] Context error during clSetKernelArg: CL_ERROR_OTHER
EXIT=2
```

N leaf pointers bound to a signature expecting one packed pointer. Note *where* the loudness comes from — OpenCL's own argument checking, not our code. On a driver that didn't check, this would be wrong data instead of an error, which is precisely why the gate exists.

Full suite **1729 cases / 0 failures**, `@fmt` green, and the suite is green both with and without ZLUDA on `PATH`.

Still to come: extending the equivalence test with the **i32/i64 device rows** the handoff asks for (now reachable, since ZLUDA gives a CUDA/PTX device).

mathiasbourgoin:
@coderabbitai review

Both actionable findings from the previous review are addressed in 4b15a323, and both were real:

1. **No SoA read-back path.** Confirmed, and worse than reported: the AoS vector gets no device buffer at all under the SoA ABI, so its location stayed `CPU` and *every* read-back path (`to_cpu`, `sync`, the auto-sync callback) returned before it ever consulted the SoA binding. `soa_to_device` now records `Stale_CPU`, and a new `soa_from_device` downloads each leaf with `~force:true` (unforced would treat post-launch `Both dev` as already in sync) and gathers.

2. **`run_source` bound the SoA ABI to an external kernel.** `soa_dispatch` now takes `~soa_abi`, defaulting to **false**; only the two generated-JIT sites pass `true`. Fail-closed, because the dangerous direction is a source string Sarek did not emit.

`soa_leaves_live` is set by the upload closure and only *read* by read-back, so the round trip has one answer to "SoA or AoS?" rather than two sites deriving it independently.

Why the existing tests all passed: `check_transparent` held the SoA vector input-only, and `check_roundtrip` covers the *explicit* `Soa_launch` path where the caller does the D2H and gather by hand. Neither read an output back through the transparent path. `check_transparent_roundtrip` now does, with a plain `Vector.get` — red before the fix, green after, measured on CUDA/PTX via ZLUDA.

mathiasbourgoin:
## Follow-up: the `Stale_CPU` write-loss was a **class**, not one function (backlog-190)

The H5 fix in this branch covered `Vector.set`. I stopped at the function I was reading. The class is *every host writer that can run on a `Stale_CPU` vector*, and there were three — `unsafe_set` carried the identical arm, and `fill`'s was a `| _ -> ()` catch-all, **wider still**.

Each got its own decision, because the right answer differs:

| writer | decision | why |
|---|---|---|
| `unsafe_set` | **syncs**, same as `set` | documented "no bounds check", silent on staleness — so the only difference a caller can expect from `set` is the bounds check. `ensure_cpu_sync` is a branch, not a copy, so the performance contract is intact |
| `fill` | **does not sync**; marks `Stale_GPU` from `Stale_CPU` | it overwrites *every* element, so pulling the old buffer down first is pure waste. Arms now spelled out — the `| _ -> ()` is how this hid |
| `kernel_set` | **stays exempt, and now says so** | the only writer without staleness handling; a per-element update would race across the threads it exists to serve |

### Evidence: a diagonal, not three green ticks

`check_relaunch` is parameterized over the writer, one case each — separate cases so a fix to one cannot make the others *read* as covered, which is exactly how this survived the first fix.

6 cases green (3 writers × 2 PTX devices via ZLUDA). And each fix proven red **independently**:

| mutation | `set` | `unsafe_set` | `fill` |
|---|---|---|---|
| `unsafe_set` reverted | OK | **FAILED** | OK |
| `fill` reverted | OK | OK | **FAILED** |
| `set` reverted | **FAILED** | OK | OK |

No case is carrying another.

One slip worth recording: I first wrote `fill`'s expectation as 1000, forgetting the kernel *doubles* it (round 2 fills y=1000 → 2000; a stale round 2 re-doubles round 1 to 4). Caught by running it rather than reasoning about it.

mathiasbourgoin:
## Five open findings triaged — two fixed, two refuted with a reason, one already landed and now proven load-bearing

Rebased onto `origin/main` first (one additive conflict in `test_soa_emitter_equiv.ml`: `main`'s `check_short_arg_list` and this branch's `check_transparent` are neighbours, both kept).

### CONFIRMED and fixed — `soa_leaves_live` was only ever SET

`3678702680` / `3678702683`. Confirmed by grep: `leaves_live := true` in `Soa_vector.create_transparent`'s `soa_to_device` closure is the only write in the tree. Nothing clears it, so once a transparent CUDA/PTX launch has run, **every** later read-back on that vector follows the leaves for the rest of its life, whichever ABI the most recent launch used. `Execute.run_source` deliberately keeps the packed AoS ABI, so the sequence is reachable through the public API with no unusual step.

Clearing the flag is not the whole fix, and the two extra lines next to it are not bookkeeping. After an SoA launch the **authoritative** copy is the leaves, so the gather has to happen before the packed launch can upload anything meaningful — and the location then has to be one `to_device` does not short-circuit. It records `Stale_CPU` on this very device, which `to_device` skips (`"skip (Stale_CPU)"`), leaving the packed launch to run against pre-SoA-launch bytes, or against **no buffer at all**, since the transparent path never allocates one. That last case is why the pre-fix red below is an exception rather than wrong data.

Normalised on **both** outcomes of the one `soa_dispatch` predicate, so the flag states the ABI of the *most recent* launch rather than of *some* launch.

### CONFIRMED and fixed — cleanup had no SoA arm at all

`3678702684`. `Transfer.free_buffer` and `free_all_buffers` read a vector back before releasing its memory, but called `copy_device_to_host` **directly**. Traced both:

- `free_all_buffers` — `location = Stale_CPU d` → `copy_device_to_host` → `get_buffer = None` → `Failure "to_cpu: no device buffer to transfer from"`, **raised from inside cleanup**.
- `free_buffer` — the outer `match get_buffer with None -> ()` returns first, so the transparent launch's output is discarded **silently**.

Both now go through one `read_back_to_host`, the same function `to_cpu` uses, so the SoA-vs-packed decision exists once instead of three times. A new `has_device_data` preserves the no-op that early return was actually for: an AoS vector with nothing on this device.

### REFUTED as filed — `sync_vectors_to_cpu` already routes SoA results

`3678702687` asks to select `soa_from_device` in `Execute.sync_vectors_to_cpu` "instead of calling `Transfer.to_cpu` unconditionally". Calling `Transfer.to_cpu` **is** the routing: `Transfer.ml` dispatches on `vec.soa` / `soa_leaves_live` inside `to_cpu` itself, so the SoA arm is reached one level down, and every other caller (`Transfer.sync`, `Vector.get`'s auto-sync, the cleanup paths) inherits it for free. Adding the same decision in `sync_vectors_to_cpu` would put a second, independently-derived answer to "SoA or AoS?" on the round trip — which is the thing that produces an upload of leaves against a download of a packed buffer. The finding also asks to pass "the active device handle"; `to_cpu` already takes it from `vec.location`, which is the only place that records it.

Not fixed, deliberately. The behaviour it describes is real and is covered — by the arm that already exists.

### REFUTED as filed — the date was not future-dated

`3678385949`. The comment reads `CORRECTED 2026-07-30`; the commit that wrote it is authored `2026-07-30 00:09:26 +0200`. The review ran at `2026-07-29T22:15Z` and compared against UTC, which is where the one-day gap comes from. Changing it to 2026-07-29 would make it wrong in the timezone the commit was authored in. Left as is.

### Already landed in 4b15a323 — and now proven load-bearing rather than asserted

`3678369465` ("`soa_gather` is never called") was the headline of this review round and is fixed: `Transfer.to_cpu` reaches `b.soa_from_device dev`, which force-downloads every leaf and gathers. The thread is resolved, but a resolved thread is not evidence, so the gather was mutated and re-measured — replacing the `soa_from_device` call with a no-op, leaving everything else intact:

```
  transparent round-trip mismatch @0: got {x=0;y=1;z=1024} want {x=0;y=2;z=1024}
  transparent SoA output read back (PTX: leaf D2H + gather) FAILED
EXIT=1
```

`y=1` is the pre-launch host value — the original silent-stale-output failure mode exactly, reproduced on demand. The assertion is not vacuous.

### Evidence — per device, and the CUDA rows are ZLUDA on AMD

`transparent SoA output read back` is a plain `Vector.get` after `run_vectors`, no leaf iteration and no explicit gather, which is the end-to-end claim.

| Device | ABI | read-back | SoA→packed | free-then-read |
|---|---|---|---|---|
| `[CUDA/PTX] AMD Radeon RX 7900 XTX [ZLUDA]` | N-leaf | ✅ | ✅ | ✅ |
| `[CUDA/PTX] AMD Ryzen 9 7950X iGPU [ZLUDA]` | N-leaf | ✅ | ✅ | ✅ |
| `[OpenCL] RX 7900 XTX (radeonsi navi31)` | AoS fallback | ✅ equals reference | n/a (SoA is PTX-gated) | n/a |
| `[OpenCL] Ryzen 9 7950X (radeonsi)` | AoS fallback | ✅ equals reference | n/a | n/a |
| `[Vulkan] RX 7900 XTX (RADV NAVI31)` | AoS fallback | ✅ equals reference | n/a | n/a |
| `[Vulkan] Ryzen 9 7950X (RADV)` | AoS fallback | ✅ equals reference | n/a | n/a |
| `[Native] CPU (Parallel, 32 cores)` | AoS fallback | ✅ equals reference | n/a | n/a |
| `[Interpreter] Sequential` / `Parallel` | AoS fallback | ✅ equals reference | n/a | n/a |

**Every CUDA/PTX row above is measured through ZLUDA on AMD hardware. There is no NVIDIA device on this host and none of these rows should be read as native NVIDIA.** The two SoA-only cases are PTX-gated because `soa_dispatch` is, and the read-back row skips on the CPU backends for an unrelated reason (`v.(i).f <-` is blocked there — backlog-172), which the run states rather than hides.

### Prove-red: a diagonal, not two green ticks

Each fix reverted independently with `git checkout <pre-fix-sha> -- <path>` (not `git stash`), rebuilt, re-run:

| reverted | `packed AoS launch after a transparent SoA one` | `free_all_buffers preserves a transparent SoA result` |
|---|---|---|
| `sarek/execute/Execute.ml` | **FAILED** — `packed launch after a transparent one raised: Vector transfer failed for unknown: Vector has no device buffer`, EXIT=1 | OK |
| `sarek/core/Transfer.ml` | OK | **FAILED** — `free_all_buffers raised: Failure("to_cpu: no device buffer to transfer from")`, EXIT=1 |

Neither case carries the other, which is why they are two cases and not one: the two paths that got a stale answer are separate — the launch path and the cleanup path. And neither does a read-back before the step under test; inserting one would clear the staleness by accident and both would pass unfixed.

### Gates

`dune build` 0 · `dune runtest` 0 · `dune build @fmt` 0 · `check-license-headers.sh` 0 · `check-kb-properties.sh` 0 · `prove-red.sh` 0 (6 subjects, 28 mutations) · SoA e2e suite exit 0, zero `FAILED` lines.

mathiasbourgoin:
## Round 3 — and the round-2 refutation was wrong

**The "second unconverted read-back site" finding was correct as filed. I refuted it. That refutation was wrong, and this is not a quiet fix.**

`Transfer.to_device`'s cross-device migration arm called `copy_device_to_host` **directly**. It never went through `read_back_to_host` — the function whose entire purpose is deciding whether a read-back reads the packed AoS buffer or the N SoA leaves. My round-2 reply said the site was already covered. It was not; I checked the three call sites the fix touched and not the fourth one that existed.

Worse, the comment I wrote **on** `read_back_to_host` asserted the thing that was false:

> "The ONE place that decides WHERE a read-back reads from, shared by **every** path that pulls device data into host storage — `to_cpu` and both buffer cleanup paths."

The migration arm is such a path and did not share it. The comment named three sites and claimed to name all of them, which is precisely the shape that made the finding easy to dismiss: the code read as if it were already complete.

Both halves of the failure pair this PR claims to have eliminated were live through that arm. Measured on two CUDA/PTX devices (ZLUDA on an AMD RX 7900 XTX — a CUDA/PTX device, **not** NVIDIA hardware; there is no NVIDIA GPU in this host):

| sequence | result |
|---|---|
| `to_device(other dev)` after a transparent SoA launch | **RAISED** `Failure("to_cpu: no device buffer to transfer from")` |
| packed buffer present, then `to_device(other dev)`, then read | **DATA WRONG** `@0 got=1 want=2`, silently |

The same error message, and the same silent discard of the SoA output.

### What changed

The migration arm now routes through `read_back_to_host`, and the comment is an enumeration of **four named call sites** — exhaustive by construction, because `copy_device_to_host` is now called nowhere else in the module:

- `to_cpu` (and `sync` through it) — the explicit read-back
- `to_device`'s cross-device **migration** arm — drain the source device before uploading to another
- `free_buffer` and `free_all_buffers` — drain before release

It also records that an earlier version of the comment claimed the migration arm was already included, that it was not, and that both measured failures were reachable through it.

New file `sarek/tests/e2e/test_soa_cross_device_migration.ml` pins both halves. Separate from `test_soa_emitter_equiv.ml` deliberately: the raise half's pre-fix red is a crash, which would abort a shared binary before the other half reported.

**Prove-red** (revert `Transfer.ml` to `6bc76be0`, keep the new test):

```
  migration raised: Failure("to_cpu: no device buffer to transfer from")
  migration after a transparent SoA launch (no packed buf)   FAILED
  migration data wrong @0: got=1 want=2 (the stale packed buffer holds 1)
  migration after a transparent SoA launch (packed present)  FAILED
```

Exactly the two symptoms as filed. Restored: both OK.

---

### Also in this round

**Four e2e cases never ran under a plain `dune runtest`, and printed nothing.** All are behind `if is_ptx dev`, and nothing put ZLUDA on the loader path — `LD_LIBRARY_PATH` appeared only in dune **comments**. So `runtest` exited 0 with this PR's last three commits unverified and no skip notice: `check_relaunch` ×3, `check_soa_then_packed`, `check_free_preserves_soa`, all zero lines. Two changes, because either alone is insufficient:

1. every guarded case now prints a skip line naming the device class it needs and why it was absent — **9 → 51 skip lines** with no CUDA/PTX device present, so a silent no-op is impossible;
2. the dune rules `setenv LD_LIBRARY_PATH`, so the device is actually there. Verified: `dune build @sarek/tests/e2e/runtest` with no ambient `LD_LIBRARY_PATH` now enumerates both ZLUDA devices and runs all seven previously-silent assertions.

Trade-offs of (2) are written into the dune comment rather than assumed: a nonexistent ZLUDA directory is ignored by the loader (no-op), a native CUDA driver lives on the default loader path (unaffected), and a host that reaches CUDA *only* through an inherited `LD_LIBRARY_PATH` degrades to the loud skip — never to a false green.

**`free_all_buffers` freed nothing and reported nothing.** `device_buffers` is empty under this ABI, so the call succeeded having released zero bytes — `Gpu_memory.usage()` 3840 → 4224 across launch+free, delta +384 = 32 × 3 leaves × 4 B. A leak, not wrong data: leaves carry `register_finalizer`. Fixed by freeing (new `soa_free_leaves` closure on the binding, wired into both cleanup paths, clearing `soa_leaves_live`), not by reporting. New case `check_free_releases_leaves` asserts bytes come back, and asserts `allocated > 0` so it cannot pass vacuously — it reds at "released 0 of the 12288 bytes the launch allocated" while `check_free_preserves_soa` stays green, which is the evidence that correctness and release were genuinely separate claims.

**Four doc comments documented an API this PR decided not to build** — "present iff the vector was created with `~layout:SoA]`". No `layout` parameter exists anywhere; `Soa_vector.ml` records that `Vector.create ~layout:SoA` was *rejected*. All four now describe `create_transparent`.

**`test_soa_emitter_equiv.ml` contradicted itself about what runs locally** — "this machine has no NVIDIA device, so locally it is the fallback half that runs" against "it needs a CUDA/PTX device. ZLUDA provides one" in the same file. Nine devices enumerate here and two are CUDA/PTX. "No NVIDIA device" and "no CUDA/PTX device" are different statements and only the second is what `is_ptx` tests; the wrong half pointed a reader away from the arm carrying every fix.

**LOWs.** `has_device_data` was documented "for `[vec]` on `[dev]`" while its SoA arm ignores `dev` — the binding holds one `soa_leaves_live` for the whole vector, not one per device, so it *cannot* answer "on `[dev]`". The doc now says that and names the change that would fix it, rather than pretending. The handoff doc's "OK on the AoS fallback (5 non-PTX devices)" applied the i32+f64 count to both rows: i64+i32 is OK on **all 7**, i32+f64 on **5**, skipping the 2 rusticl devices without fp64. Counted separately now.

**Deliberately not done.** I considered clearing `soa_leaves_live` in `to_device`'s packed-upload branch, for symmetry with `Execute.transfer_vectors_to_device`. I traced the reachable sequences instead of adding it defensively: after the migration gather the host and the leaves hold the same bytes, and only `Execute` writes device buffers (normalising the flag itself), so no wrong data is reachable from the stale flag. Left out, and recorded as such.

### Gates

Every exit code captured directly, never `$?` after a pipe.

| gate | exit |
|---|---|
| `dune build` | 0 |
| `dune runtest` (full) | 0 |
| `dune build @fmt` | 0 |
| `scripts/check-kb-properties.sh` | 0 (34 declarations hold) |
| `scripts/prove-red.sh` | 0 (6 subjects, 28 mutations) |
| SoA e2e **with** ZLUDA | 0 |
| SoA e2e **without** ZLUDA | 0 (51 skip lines, every case named) |
| migration e2e **with** ZLUDA | 0 (both halves OK) |
| migration e2e **without** ZLUDA | 0 (loud skip: 7 devices, none CUDA/PTX) |

### Not verified

No native NVIDIA hardware is available here, so every CUDA/PTX row in this PR is ZLUDA on an AMD RX 7900 XTX. Nothing was run against a real NVIDIA driver, and no row should be read as one. The two-device migration pair is two ZLUDA devices on one physical GPU + CPU, not two discrete GPUs.

mathiasbourgoin:
@coderabbitai review

Round 6 pushed as `6e96b89c`. Comment-only, no code change — the one blocker from the last adversarial pass was a sentence claiming more than the code does.

The HIP bullet in the framework-floor comment (`sarek/tests/e2e/test_soa_emitter_equiv.ml`) said HIP had "no registering backend today". False: `sarek-hip/Hip_plugin.ml:86` calls `Framework_registry.register_backend ~priority:80` under the name `"HIP"` (`Hip_plugin_base.ml:17`), forced at module init on `:93`. HIP is in the same category as CUDA/C and WebGPU — a `register_backend` backend absent from *this host's* table only because nothing forced its registration or `is_available ()` returned false.

Two adjacent corrections in the same comment, same "claims more than the code does" family:

- The paragraph said `Cuda_ptx_plugin` and `Cuda_c_plugin` "are both `[lazy]` behind an explicit `[register ()]`". Only the latter is; `Cuda_ptx_plugin.ml:99`, `Hip_plugin.ml:93` and `Webgpu_plugin.ml:195` force at module init. Both shapes are now named, plus the consequence that HIP and WebGPU cannot enter this executable's table at all because module-init forcing needs the module LINKED and neither is in `sarek/tests/e2e/dune`'s libraries.
- The corrected bullet cites `Device.init`'s default framework list, which also contains the bare `"CUDA"` — not in the floor. Stated why: no backend registers under it, `resolve_framework` expands a family name to its variants, so no `Device.t` carries it.

Worth your attention on:
- Whether the "all three sit in one category" paragraph is actually true of all three, or whether I have flattened a distinction that matters.
- Whether the `"CUDA"` paragraph clarifies or just adds noise.
- Any remaining sentence in that comment block asserting more than its cited lines support.

mathiasbourgoin:
## Round 5 — all 10 points from the review on `bbb7df7`, then what the review of the fix found

Rebased onto current `main` (#390, #393). Five commits: `93ff505`, `59430ac`, `2c84a40`, `3dd445c`, `b332de2`.

Nine fixed, one refuted. Then a cross-runtime and a fresh-read review of my own fix found a **data-loss regression I had introduced**, plus a check of mine that could not fail — three more commits below.

---

### The 5 inline comments

**1. `Transfer.ml#L385-389` + `Execute.ml#L335-341` — normalise `soa_leaves_live` on the packed upload; don't let a location skip the gather. FIXED, both halves, but the second one twice.**

`Transfer.to_device` now clears the flag after a packed host→device upload — only in the arm that actually uploaded, since the already-resident arms are exactly where the leaves *are* the device copy. Reaching that arm with the flag set means the host copy was authoritative or in sync first, which is what makes the upload meaningful. Being straight about the value: **nothing goes red without this**. The suites were green before it. Through the launch API the flag was already normalised, so this is the flag meaning what its own contract says, not a bug fix.

`Execute.transfer_vectors_to_device` was the substantive half, and it took two attempts. Unconditional (`93ff505`) was wrong: `CPU`/`Stale_GPU` assert the *host* copy is newer, and `Stale_GPU` with the flag set is reachable through public API — a transparent launch leaves `Stale_CPU dev`, then `Vector.set` gathers, writes and records `Stale_GPU dev`. An unconditional gather replays the leaves over that write and it vanishes silently: the backlog-181/190 class, on the packed-after-SoA path. It is now decided on `v.location` — gather on `GPU`/`Stale_CPU`/`Both`, skip on `CPU`/`Stale_GPU`, clear the flag either way. Pinned by a new case, proven red against the unconditional version:

```
host write lost @0: got=4 want=1 (re-gathered leaves would give 4)
```

(`got=4` is exactly `4*y0` at `i = 0` — the value the discarded write would leave.)

**2. `Execute.ml#L361-377` — no arity guard between `soa_leaf_bufs` and the N-pointer signature. FIXED, with its scope corrected.**

Guard added: `List.length bufs` against `b.soa_num_leaves`, `Transfer_failed` on mismatch. Proven red by truncating `soa_leaf_bufs`:

```
Vector transfer failed for SoA-dispatched vector: the SoA binding produced 2 leaf
buffer(s) on device 0 but declares 3 leaves; binding a list of the wrong length ...
```

Two corrections to how I first described it. The condition is `<>`, so it rejects a long list too — the message no longer says only "short". And it does **not** check the vector's leaf count against the *kernel's*: both sides come from the vector, while the emitter derives the parameter block from the `TRecord` on the kernel's own `DParam`. A transparent vector bound to a parameter of a different record type still satisfies this check. `Soa_launch.run_soa` has that comparison (`check_soa_layout`); the generic transparent path does not, and did not before — `check_launch_args` skips the element comparison for `Custom` elements because both sides "derive from the same registered layout", which is the assumption the SoA ABI makes load-bearing. Not closed here (`Execute` cannot call `Soa_launch`, and this function has no IR); stated at the guard.

Your premise that `soa_leaf_bufs` "may return fewer buffers" is not true of the only implementation — it maps over the leaf array and raises `invalid_arg` for a missing buffer (`Soa_vector.ml:148-160`). The guard is for the cross-module convention, since the binding is a record of closures that exists precisely so `Execute` cannot see `Soa_vector`.

**3. `test_soa_cross_device_migration.ml#L37-38` — exit code 2 unreachable. FIXED.** Removed; every skip exits 0, which is the intent documented at the skip sites. The header now says so, and the same header's device requirement was updated again when point 4's fix added a case with a stricter one.

**4. `test_soa_cross_device_migration.ml#L313-321` — restrict `dst` to a real device-buffer target. REFUTED for the three migration cases, and the byte accounting is the citation.**

`Transfer.to_device` on a Native/Interpreter target still produces a non-zero byte delta: `ensure_buffer` calls `Gpu_memory.track_alloc (size * elem_size)` for **every** framework (`Transfer.ml:232`) on a wrapper whose `size`/`elem_size` are the requested length and element size (`Transfer.ml:150-152`), and `free_buffer` tracks the same product back. The accounting is not the backend's, so `after_migration - after_launch2` is not 0 and the split stays measurable. Restricting `dst` would drop the heterogeneous pair, which is the more interesting migration. Written at the selection site.

Your review did find a real `dst` problem, just not this one — see the fourth commit below.

**5. `test_soa_emitter_equiv.ml#L1841-1847` — `dpair(f64)` skips silently. FIXED.** Named skip, your wording. Observed live on this host's two rusticl devices:

```
SoA-emitter dpair(f64) [OpenCL] AMD Radeon RX 7900 XTX (radeonsi, navi31, ...): SKIP (device reports no fp64)
```

---

### The 5 nitpicks

**6. `mixed_ty` collides with the new `mixed` record. FIXED.** `mismatch_ty` over the record `"ab_pair"`. Nothing compares the name — `check_soa_layout` compares leaves and stride only (`Soa_launch.ml:117-121`) — so the rename cannot change what those cases assert.

**7. `&& !ok` suppresses later diagnostics. FIXED, and it is the substantive one of the five.** One `first_mismatch` ref per claim in `check_free_buffer_coherent`, for claims 2 and 5. Proven red by making claim 1 fail *and* claim 2 mismatch:

```
location after free_buffer: Stale_GPU 0, want CPU
to_cpu after free_buffer corrupted @0: got=2 want=5 (pre-launch host value is 1)
```

Both lines. Before, the second was suppressed.

**8. Duplicated relaunch label. FIXED.** One `relaunch_label writer_name`, used by the pass line and the `guarded ~label`.

**9. `LD_LIBRARY_PATH` overwritten in both e2e dune rules. FIXED.** Prepended. The loss was wider than the CUDA driver the rule is about: these tests assert on every device the host enumerates, so a host reaching its OpenCL ICD or Vulkan loader through an inherited path lost those devices — real rows, failing nothing. The trailing-colon consequence of `%{env:LD_LIBRARY_PATH=}` when unset is noted at the rule.

**10. Document why `copy`/`sub_vector` drop the binding. FIXED,** and they are not the same reason: a copy could not safely inherit closures that capture the source's leaves (a launch on the copy would write the original's), while a subvector's length simply differs from the parent's leaves, so the binding does not describe it at all.

---

### Two earlier points you asked about, verified against current head

**Packed launch never clears `soa_leaves_live`** — was already fixed at `Execute.ml`, in the `None` arm of the one `soa_dispatch` predicate: gather, clear, record `Stale_GPU dev`. This round corrected *which locations* that gather runs on, twice.

**Missing reverse transfer** — already present: `soa_from_device` in the binding (`Soa_vector.ml`, `to_cpu ~force:true` per leaf then `gather`), selected by `Transfer.read_back_to_host`, which is the single decision point shared by `to_cpu`, `to_device`'s migration arm, `free_buffer` and `free_all_buffers`.

**Device-independent plan checks for `mixed`/`longpair`** — this had **not** landed. Added to `check_field_derivation`, which runs before the no-device exit. `point3d` and `dpair` are both uniform-width, so neither can distinguish a correct offset from one ignoring padding. Discrimination proven: dropping trailing padding from the stride fails **only** the `longpair` row —

```
derived plan for point3d (3 x f32)                       OK
derived plan for dpair (2 x f64)                         OK
derived plan for mixed (i32 then f64: pad after the i32) OK
derived plan for longpair (i64 then i32: trailing pad)   FAIL (stride 12, expected 16)
```

This matters here because the executed rows need a CUDA/PTX device plus fp64/int64, so in CI they do not run at all.

---

### What the review of the fix found (commits 3-5)

Three further commits, because the fix for one direction opened another:

- **`soa_free_leaves` re-derived the flag from what is still ALLOCATED.** Allocation is not authority: a packed launch gathers the leaves and clears the flag, and those buffers stay allocated, so a later free on a device the leaves are *not* on turned ownership back on. Harmless while the gather was location-gated; data loss once it was not. Now a conjunction with the previous value — that free can clear the flag, never set it. New two-device case, proven red.
- **That new case was then a gate that could not fail.** It asserted only data, and re-gating the gather on `location` removed the data consequence: measured, it printed OK against the broken `soa_free_leaves`. It now reads the invariant directly.
- **`"THREE writers, and the list is exhaustive"`** — I had added a fourth. Four now, with which one can *set* it spelled out; plus five more claims of mine with wrong counts or wrong mechanisms (`fill` does not gather; `to_device` records `Both` twice; a citation into `Vector.ml` that resolved to nothing; a separation claim that was false at `i = 98`).

**One exposure documented and not closed.** Two transparent launches on A then B leave leaves on both; `free_buffer sv B` drains B and releases B's leaves, and the flag correctly stays set because A's survive — that rule is what stops a per-device free from disowning another device's live leaves. A's leaves are then stale, and a **direct** `read_back_to_host sv A` gathers them over the host copy. No path inside `Transfer` reaches it, but there is no `Transfer.mli`. Closing it means making `soa_leaves_live` name a device — a change to the binding contract, which I am not landing in a review round. Written where `has_device_data` explains why the whole-vector flag cannot answer "on `dev`".

---

### Gates

`dune build` 0 · `dune runtest --force` 0 · `dune build @fmt` 0 · `scripts/prove-red.sh` 0 (6 subjects, 33 mutations) · `check-kb-properties.sh` 0 · `.test.sh` 0 · `check-license-headers.sh` 0 · `check-opam-clean.sh` 0 · `check-cited-paths-exist.sh` 0.

**What rests on hardware CI does not have.** Every executed SoA row needs a CUDA/PTX device; CI has none, so there the SoA legs print named skips. Verified locally on two CUDA/PTX devices via ZLUDA on AMD hardware (RX 7900 XTX + Ryzen 7950X), which is also what makes the two-device migration cases runnable at all. Specifically device-dependent: the `soa_leaves_live` lifecycle cases, `check_host_write_survives_packed`, `check_free_does_not_resurrect_leaves`, and the leaf-count guard's red proof. Device-**independent**, and therefore the part that does run in CI: the four `check_field_derivation` rows, the layout-refusal cases, and the skip-reason scope check.

@coderabbitai review

mathiasbourgoin:
### SoA → SoA on another device — the gap is real, the loss is not. `4544864`

You are right that nothing covered it. Every case in `test_soa_cross_device_migration.ml` was SoA → **packed**, where `Transfer.to_device`'s migration arm is what drains, and `Execute.transfer_vectors_to_device` does call `soa_to_device` directly rather than going through `to_device`. So the sequence you describe was uncovered.

It does not lose the first launch's output. The drain is inside `scatter` itself: `Soa_vector.scatter` opens with `Vector.ensure_cpu_sync t.aos`, and on the `Stale_CPU A` that a transparent launch leaves, that fires the sync callback → `Transfer.to_cpu` → `read_back_to_host` → the SoA arm → A's leaves. So by the time the transpose reads the AoS host buffer, it holds A's result.

Executed rather than argued, on two CUDA/PTX devices (ZLUDA on an RX 7900 XTX and a Ryzen 7950X):

```
a second SoA launch on another device sees the first       OK
```

`y0 = i+1`, each launch doubles `y`, so two launches give `4*y0` and a second launch fed pre-A host bytes gives `2*y0`. Proven red by replacing that `ensure_cpu_sync` with a no-op — which is also the positive control for your reproduction:

```
second SoA launch ran on pre-first-launch host bytes @0: got=2 want=4
(scattering the un-drained host copy gives 2)
```

`got=2` is exactly `2*y0`, i.e. precisely the failure you predicted. It is one `ensure_cpu_sync` away, four indirections deep, two of them function pointers — which is why the case is now in the file instead of the reasoning being in a comment.

One precondition named at the case, same as `check_host_write_survives_packed`: `ensure_cpu_sync` is a no-op when `auto_sync` is false on the vector or auto mode is off globally (`Vector_transfer.ml`), so with either disabled the drain does not happen. That is the pre-existing auto-sync-off hazard, not something this path decides, and the case does not disable either.

Gates after this commit, exit codes captured directly: `dune build` 0 · `dune runtest --force` 0 · `dune build @fmt` 0 · `scripts/prove-red.sh` 0 · `check-kb-properties.sh` 0 · `.test.sh` 0 · `check-license-headers.sh` 0 · `check-opam-clean.sh` 0 · `check-cited-paths-exist.sh` 0.

The two-device cases need a second CUDA/PTX device and print a named skip without one, so this row does not execute in CI.

@coderabbitai review

mathiasbourgoin:
### Round 5, final state — rebased onto current `main`, `22c40c5`

Rebased past #394/#395/#396 and the ppx/gate commits. The rebase is content-identical: `git diff 3bdf2ae2..45448647` and `git diff origin/main..22c40c5f` are byte-for-byte the same, so nothing in the eight commits changed under it. Verified against both parents rather than trusting a clean rebase, since `rerere`'s cache is shared here.

Consolidated disposition, since this round ran long. **Nine fixed, two refuted with citations, one exposure documented and deliberately not closed.**

| # | Point | Disposition |
|---|---|---|
| 1 | `Transfer.ml` L385-389 + `Execute.ml` L335-341 — normalise `soa_leaves_live` | **fixed** `93ff505`, corrected twice: `2c84a40`, `3dd445c` |
| 2 | `Execute.ml` L361-377 — no arity guard on `soa_leaf_bufs` | **fixed** `93ff505`, scope corrected `59430ac` |
| 3 | migration test L37-38 — exit code 2 unreachable | **fixed** `93ff505` |
| 4 | migration test L313-321 — restrict `dst` to a device-buffer target | **refuted**, citation below |
| 5 | `dpair(f64)` silent skip | **fixed** `93ff505` |
| 6 | `mixed_ty` collides with the new `mixed` record | **fixed** `93ff505` |
| 7 | `&& !ok` suppresses later claims' diagnostics | **fixed** `93ff505` |
| 8 | duplicated relaunch label | **fixed** `93ff505` |
| 9 | `LD_LIBRARY_PATH` overwritten in both e2e dune rules | **fixed** `93ff505` |
| 10 | document why `copy`/`sub_vector` drop the binding | **fixed** `93ff505` |
| — | SoA → SoA on another device (raised in your re-review) | **refuted**, executed; gap closed `4544864` |

**Point 4, refuted.** The byte accounting is not the backend's: `Transfer.ensure_buffer` calls `Gpu_memory.track_alloc (size * elem_size)` for every framework (`Transfer.ml:232`), on a wrapper whose `size`/`elem_size` are the requested length and element size (`Transfer.ml:150-152`), and `free_buffer` tracks the same product back. A Native or Interpreter `dst` therefore still yields a non-zero packed-byte delta and the split stays measurable. Restricting `dst` would drop the heterogeneous pair, which is the more interesting migration. Written at the selection site. Your review did surface a real `dst` problem — the case added in `59430ac` launches PTX on `dst` and would have exited 1 on a non-PTX second device; that one is fixed with a named skip in `2c84a40`.

**The two earlier data-integrity points, checked against the current head rather than assumed.**

*Packed launch never clears `soa_leaves_live`* — was already fixed before this round, in the `None` arm of the single `soa_dispatch` predicate in `Execute.transfer_vectors_to_device`: gather, clear the flag, record `Stale_GPU dev`. What this round changed is **which locations that gather runs on**, and it took two attempts to get right. Unconditional was wrong: `Stale_GPU` with the flag set is reachable through public API (transparent launch leaves `Stale_CPU dev`; `Vector.set` gathers, writes, records `Stale_GPU dev`), and gathering there replays the leaves over the host write. It is now `gather on GPU/Stale_CPU/Both, skip on CPU/Stale_GPU, clear either way`, pinned by `check_host_write_survives_packed` and proven red against the unconditional version:

```
host write lost @0: got=4 want=1 (re-gathered leaves would give 4)
```

*Missing reverse transfer* — already present: `soa_from_device` in the binding (`to_cpu ~force:true` per leaf, then `gather`), selected by `Transfer.read_back_to_host`, the single decision point shared by `to_cpu`, `to_device`'s migration arm, `free_buffer` and `free_all_buffers`. Not re-fixed; cited.

**`check_field_derivation` for `mixed`/`longpair`** — had not landed; added. `point3d` and `dpair` are both uniform-width, so neither distinguishes a correct offset from one ignoring padding. Discrimination proven — dropping trailing padding from the stride fails **only** the `longpair` row:

```
derived plan for point3d (3 x f32)                       OK
derived plan for dpair (2 x f64)                         OK
derived plan for mixed (i32 then f64: pad after the i32) OK
derived plan for longpair (i64 then i32: trailing pad)   FAIL (stride 12, expected 16)
```

This is the row that matters for CI, which has no GPU: the executed mixed-width rows need CUDA/PTX plus fp64/int64, so without these the padding-sensitive layouts had no check that runs there.

**Two defects this round introduced and then removed**, both found by review of my own fix rather than by the gates — worth recording since they are the same class the PR is being reviewed for:

- the unconditional gather above (data loss on a host write);
- `check_free_does_not_resurrect_leaves` was a gate that could not fail: it asserted only data, and re-gating the gather removed the data consequence. Measured — it printed OK against the broken `soa_free_leaves`. It now reads the invariant directly, and is red again without the fix.

**One exposure documented, not closed.** Two transparent launches on A then B leave leaves on both; `free_buffer sv B` drains B and releases B's leaves, and the flag correctly stays set because A's survive — that rule is what stops a per-device free from disowning another device's live leaves. A's leaves are then stale, and a **direct** `read_back_to_host sv A` gathers them over the host copy. No path inside `Transfer` reaches it, but there is no `Transfer.mli`, so the function is exported. Closing it means making `soa_leaves_live` name a device: a change to the binding contract, not something to land in a review round. Written where `has_device_data` explains why the whole-vector flag cannot answer "on `dev`".

### Gates on the rebased tip, exit codes captured directly

`dune build` 0 · `dune runtest --force` 0 (5544 lines, not a cached alias) · `dune build @fmt` 0 · `scripts/prove-red.sh` 0 (6 subjects, 33 mutations) · `check-kb-properties.sh` 0 · `check-kb-properties.test.sh` 0 · `check-license-headers.sh` 0 · `check-opam-clean.sh` 0 · `check-cited-paths-exist.sh` 0.

### What rests on hardware CI does not have

Every SoA row that executes needs a CUDA/PTX device, and the cross-device cases need two. This host provides them through ZLUDA on AMD hardware (RX 7900 XTX + Ryzen 7950X); **GitHub CI has no GPU, so there those rows print named skips and the following claims rest on local runs only**: the `soa_leaves_live` lifecycle cases, `check_host_write_survives_packed`, `check_free_does_not_resurrect_leaves`, `check_soa_then_soa_other_device`, and the leaf-count guard's red proof. What does run in CI: the four `check_field_derivation` rows, the layout-refusal cases, the skip-reason scope check, and every red proof driven by `scripts/prove-red.sh`.

@coderabbitai review

mathiasbourgoin:
The `CHANGES_REQUESTED` on this PR is still recorded against `bbb7df7`, which is eight commits and a rebase behind the current head `22c40c5`. Every point from that review has a disposition in the comment above — nine fixed, two refuted with citations from the current code — and CI is green on the new head.

Requesting a fresh full review so the verdict lands on what is actually there now.

@coderabbitai full review

mathiasbourgoin:
@coderabbitai full review

The previous request hit the per-user review quota before it could start, so this is a re-trigger rather than a second review of the same state.

The outstanding `CHANGES_REQUESTED` is recorded against `bbb7df72`, which is three heads behind. Since then, round 5 addressed all ten points from that review — nine fixed, two refuted with code citations, one of them (`SoA → SoA loses the first launch`) refuted by execution rather than argument: the drain is inside `scatter` via `ensure_cpu_sync`, and the coverage gap that finding identified was real, so the case now exists and is red without that call.

Worth a look in particular, since they were introduced *by* the fixes and then caught and repaired in the same round:

- an unconditional gather that discarded a host write (`Stale_GPU` is reachable via `Vector.set` after a transparent launch);
- a new test that could not fail — it asserted only data, and re-gating the gather removed the data consequence, so it printed OK against the broken code.

CI is 5/5 green on `22c40c5f`, rebased onto current main (past #394, #395, #396, #376, #397).

mathiasbourgoin:
@coderabbitai full review

The previous request hit the per-user review quota before it could start, so this is a re-trigger rather than a second review of the same state.

The outstanding `CHANGES_REQUESTED` is recorded against `bbb7df72`, which is three heads behind. Since then, round 5 addressed all ten points from that review — nine fixed, two refuted with code citations, one of them (`SoA → SoA loses the first launch`) refuted by execution rather than argument: the drain is inside `scatter` via `ensure_cpu_sync`, and the coverage gap that finding identified was real, so the case now exists and is red without that call.

Worth a look in particular, since they were introduced *by* the fixes and then caught and repaired in the same round:

- an unconditional gather that discarded a host write (`Stale_GPU` is reachable via `Vector.set` after a transparent launch);
- a new test that could not fail — it asserted only data, and re-gating the gather removed the data consequence, so it printed OK against the broken code.

CI is 5/5 green on `22c40c5f`, rebased onto current main (past #394, #395, #396, #376, #397).

mathiasbourgoin:
## Round 6 — all eleven, `d163defe`

Rebased past #376/#397 (which also touches `sarek/tests/e2e/dune`); the rebase is content-identical, verified by diffing my delta against both parents rather than trusting a clean tree.

**Two of these I had previously answered from reading the code. Both answers were right in substance and neither was evidence, so both are now executed cases.** That distinction matters here: reading is what produced a wrong refutation earlier in this branch.

### 1. `Execute.ml` — reverse transfer (Major)

**Refuted, and now executed.** The line is exactly as you describe:

```ocaml
let sync_vectors_to_cpu (args : vector_arg list) : unit =
  List.iter (function Vec v -> Transfer.to_cpu v | _ -> ()) args
```

No `soa_from_device` in it. The conclusion does not follow, because **`Transfer.to_cpu` is not the packed read — it is the dispatcher.** It calls `read_back_to_host` (`Transfer.ml:507`), whose SoA arm calls `soa_from_device` when `soa_leaves_live`. Selecting the arm at this call site instead would put a second site in the business of answering "SoA or AoS?", which is the divergence the single decision point exists to prevent.

You asked for an executed case showing host data correct after a transparent launch with no explicit read-back. `check_sync_vectors_to_cpu_gathers` does that with **auto-sync disabled on the vector**, which is what makes the observation about `sync_vectors_to_cpu` rather than about `Vector.get` — with auto-sync on, `get` gathers through `ensure_cpu_sync` and the row would pass whatever the function did. Result on two CUDA/PTX devices: `OK`. Red-proved by making `soa_from_device` a no-op:

```
sync_vectors_to_cpu did not gather @0: got=1 want=2 (the un-gathered host value is 1)
```

It is also that function's only coverage anywhere — nothing in the tree calls it — which is how a claim about it could stand unchecked for a round.

### 2. `Execute.ml` — the arity guard on a direct `run` (Major)

**Refuted on the placement, and you were right that my report was incomplete.** I said "guard added" without saying which path it sits on. It sits inside `expand_to_run_source_args`, and `run` calls that itself on the generated-JIT path (`Execute.ml:955`) — so **a direct `run` does reach it**, not only `run_vectors`.

And `soa_leaf_bufs` does not return a short list in the first place: `leaf_buf` raises `invalid_arg` for a leaf with no buffer on `dev` (`Soa_vector.ml:158`), which is precisely the never-transferred case. `check_direct_run_refuses_untransferred` calls `run` with no transfer and accepts either refusal but never a normal return. Result: `OK (refused)`.

Its red state is worth reporting exactly: with the guard disabled **and** `soa_leaf_bufs` changed to drop missing leaves instead of raising, the call **SIGSEGVs** — exit 139, immediately after the row above it prints. That is your driver over-read, observed. The mutation kills the binary before the row can print, so the positive control for that case is the exit code, not the line; the row's own job is the narrower claim that the current code refuses by name. Written into the case.

### 3. `Transfer.ml` — the lifecycle enforced only in `Execute` (Major)

**Already fixed, earlier in this round** — `Transfer.ml:466` clears `soa_leaves_live` after the packed host→device upload, in the upload arm only (the three already-resident arms upload nothing, and `Stale_CPU dev` on a transparent vector is exactly where the leaves *are* the device copy). The comment names why the `Execute` site is a different site: that one runs before a launch and gathers first, this one after an upload that has already made the host copy authoritative.

### 4. `test_soa_emitter_equiv.ml` — "both launches use `run_vectors`" (Major)

**Refuted from the current code.** Launch 2 of `check_soa_then_packed` is `Sarek.Execute.run_source`, which defaults to `~soa_abi:false` — the packed external launch you ask for, with no intermediate read-back and a `Vector.get` assertion after it. The case is red-proved: removing the gather gives `SoA->packed mismatch @0: got=2 want=4 (stale leaves would be 2)`. A second case added this round, `check_host_write_survives_packed`, pins the other direction of the same condition.

### 5. `test_soa_cross_device_migration.ml` — `Device.all ()` and `Gpu_memory` (Major)

**My refutation survives, and it is now measured instead of cited.** `check_alloc_tracking_is_framework_agnostic` asserts the tracked delta is **exactly** `length * elem_size` on every enumerated device — 9 rows here including Native and both Interpreters — because a tracker counting allocations rather than bytes would satisfy a `> 0` form and the migration arm compares deltas against each other. Red-proved by gating `track_alloc` on `Device.is_gpu`: `0 bytes tracked, want 768` on Native and Interpreter. **This one needs no GPU, so unlike most of this branch it executes in CI.**

### Security & Privacy — `sarek/tests/e2e/dune` (treated as more than a nit)

**Fixed, and the first fix was wrong in a way worth stating.** Giving `%{env:...}` non-empty defaults does *not* close it: dune substitutes a default only when the variable is **unset**, so a defined-but-empty `LD_LIBRARY_PATH` still produced the trailing colon. The fix is the shape. Both rules now compose the value in `sh`, where `${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}` emits the colon and the value together or not at all, and `${HOME:?}` makes an unset HOME a loud failure rather than an absolute `/opt/zluda`. Measured in every state:

```
unset          -> [/home/mathias/opt/zluda]
defined-empty  -> [/home/mathias/opt/zluda]
/opt/x         -> [/home/mathias/opt/zluda:/opt/x]
HOME unset     -> sh: HOME: parameter null or not set, exit 127
```

Scope stated at the rule, because the obvious reading is wider: this stops the rule from **adding** an empty entry. It does not sanitize empty components already inside an inherited non-empty value (`":"` → `...zluda::`), and it cannot — such a value is already in effect for the loader that starts `sh`. Nor does the comment still claim `/usr/lib` "adds nothing that was not searched": prepending changes loader **precedence**, which is the point of the rule.

`dune runtest --force` caught my first `sh` form (`exec` on a bare relative `%{dep:}` path searches PATH: `exec: test_soa_emitter_equiv.exe: not found`). Fixed with a `case` on absolute-vs-relative, and re-verified with an empty parent `LD_LIBRARY_PATH`: the alias is green and still produces the CUDA/PTX rows, so the rule now supplies the path rather than inheriting it from my shell.

### The five Minor points

- **per-loop first-mismatch flag** — fixed. This is the defect I fixed in the sibling file this round and did not carry across; the shape of a fix that stops at the file it was reported in. Swept both files after: no `&& !ok` remains in either. Red-proved by firing the byte assertion *and* a data mismatch together — both lines now print.
- **misplaced comment block** — fixed; the `dst`-PTX rationale sits above `let d`, and `let e` carries its own (different) reason.
- **future-dated "CORRECTED 2026-07-30"** — undated, both occurrences. The remaining `2026-07-30` stamps are measurement dates and accurate.
- **superseded Tier 1c deferral** — fixed at both sentences, so the document carries one status for those rows.
- **stale `Soa_vector.mli` header** — narrowed to what is still deferred: CUDA/PTX-only dispatch, packed fallback elsewhere. Two further corrections after review of my own replacement: it wrongly said "`Execute.run`/`run_vectors`" when only `run_vectors` uploads the leaves (contradicting this branch's own new test), and the replacement caveat then lacked the CUDA/PTX qualification — on the CPU backends `run` executes against host vectors and there is no leaf upload to have skipped.

### Two outside-diff comments and three nitpicks, also addressed

- **`Vector.ml` overstates `Stale_CPU` unreachability** — fixed at both sites (`set` and `unsafe_set`): unreachable on the default path only, since `ensure_cpu_sync` returns immediately when `auto_sync` is cleared and discards the callback result. The no-op is still right, and now says why.
- **`Soa_vector.scatter` with auto-sync off** — **not changed, and flagged rather than refuted.** The behaviour is real. Making `scatter` transfer anyway would mean `auto_sync` does not mean what it says, and I could not refute it by citing a contract: `set_auto_sync` had no doc comment and nothing in `docs/` or `kb/` described the semantics — the same missing-citation trap this round already caught me in once. So the contract is now written at the switch itself, this consequence included, and whether the SoA path should override it is left for you rather than settled by me.
- **`soa_dispatch` called twice** — kept as two calls, with the reason stated: the predicate is pure, and merging into one `match` would duplicate the packed arm that the non-dispatching `Vec` case shares.
- **`check_roundtrip` silent skip** — fixed; routed through `guarded` with `blocker_needs_soa_abi` and its internal `is_ptx` early return removed, so every device prints exactly one row. It was the last bare `if is_ptx dev &&` in the loop, and my own comment two commits earlier pointed at it as the counterexample.
- **comment/definition inversions** — fixed; the `check_transparent_roundtrip` and `check_soa_then_packed` headers now sit above their own definitions.

### Gates, exit codes captured directly

`dune build` 0 · `dune runtest --force` 0 (5641 lines, not a cached alias) · `dune build @fmt` 0 · `scripts/prove-red.sh` 0 (6 subjects, 33 mutations) · `check-kb-properties.sh` 0 · `.test.sh` 0 · `check-license-headers.sh` 0 · `check-opam-clean.sh` 0 · `check-cited-paths-exist.sh` 0.

### What rests on local hardware

The SoA rows that execute need a CUDA/PTX device, and the cross-device cases need two; provided here by ZLUDA on AMD hardware (RX 7900 XTX + Ryzen 7950X). **CI has no GPU, so these rest on local runs only:** `check_sync_vectors_to_cpu_gathers`, `check_direct_run_refuses_untransferred` (including its SIGSEGV positive control), the lifecycle and cross-device cases, and `check_roundtrip`. **Executing in CI:** the new `check_alloc_tracking_is_framework_agnostic` on all enumerated devices, the four `check_field_derivation` rows, the layout refusals, the skip-reason scope check, and every `scripts/prove-red.sh` mutation.

mathiasbourgoin:
@coderabbitai full review

All eleven findings on `22c40c5f` are addressed in `d163defe` — eight fixed, three refuted with executed evidence rather than argument. CI is 5/5 green.

The two worth re-reading, because in both cases the factual observation was correct and the conclusion did not follow:

**Reverse transfer (`Execute.ml:1254`).** `sync_vectors_to_cpu` is indeed `List.iter (function Vec v -> Transfer.to_cpu v | _ -> ())` with no `soa_from_device` at the call site. But `Transfer.to_cpu` is the *dispatcher*, not the packed read: it calls `read_back_to_host` (`Transfer.ml:507`), whose SoA arm calls `soa_from_device`. Selecting at the call site would create the second decision site the single-flag design exists to prevent. Executed with auto-sync **disabled**, so the observation is about that function and not about `Vector.get`: `OK`. Red-proved by no-opping the gather: `sync_vectors_to_cpu did not gather @0: got=1 want=2`. It previously had no coverage anywhere — nothing in the tree called it.

**Arity guard (`Execute.ml:481`).** The guard lives inside `expand_to_run_source_args`, which `run` calls directly at `Execute.ml:955`, so a direct `run` does reach it; and `soa_leaf_bufs` cannot return a short list because `leaf_buf` raises for a leaf with no buffer (`Soa_vector.ml:158`). Executed: `OK (refused)`. Worth noting the over-read this guards against is real — the red state is a **SIGSEGV (exit 139)**, so the positive control in that case is the exit code rather than a result row, and that is now written into the case.

Also refuted: the packed-AoS handoff coverage — launch 2 of `check_soa_then_packed` *is* a `run_source` call with `~soa_abi:false`, and it is red-proved.

The `soa_leaves_live` normalisation in `Transfer` was already fixed at `Transfer.ml:466`. The `Gpu_memory` question is now **measured** rather than cited: exact `length * elem_size` across all 9 devices including Native and Interpreter, red-proved by gating `track_alloc` on `is_gpu` — and that check needs no GPU, so it runs in CI.

**On the `LD_LIBRARY_PATH` finding: the first fix was wrong and is replaced.** Non-empty `%{env:...=}` defaults do not close it, because dune substitutes only when a variable is **unset** — defined-but-empty still produced the trailing colon. It is now composed in `sh` with `${LD_LIBRARY_PATH:+:...}` and `${HOME:?}`, measured in all four states. The scope is stated at the rule: it stops the rule from *adding* an empty entry; it cannot sanitize one already present in an inherited value.

One item is deliberately **not** fixed and is flagged for the maintainer rather than resolved: `Soa_vector.scatter` with auto-sync off. Forcing a transfer there would make `auto_sync` not mean what it says, and it could not be refuted by citing a contract because `set_auto_sync` had no doc comment anywhere. The contract is now written at the switch, this consequence included.

PR #376, opened 2026-07-29
fix(interp,native): `v.(i).f <- e` was refused on one CPU backend and silently dropped on the other

mathiasbourgoin:
backlog-172. A documented DSL construct, used by a shipped kernel (`p3_scale_y_kernel`), behaved three different ways.

| Backend | Before |
|---|---|
| CUDA/PTX | correct |
| Interpreter | **refused** — `Unsupported_operation "record field assignment" / "not fully supported"` |
| Native, immutable record | **did not compile** — `The record field b is not mutable` |
| Native, mutable record | **compiled and silently discarded the store** |

The last two are one cause with two symptoms, and the first pushes you into the second: `Vector.get` returns a fresh record marshalled out of the vector's storage, so a `setfield` on it hits a temporary. Declaring the fields `mutable` makes the *compile error* go away and converts a build failure into wrong data. `point3d` in `test_soa_emitter_equiv` carries exactly that `mutable`, with a comment calling it necessary "to write a leaf in place". After this PR `mutable` is not required for either shape.

## Native

A functional update written back through the same `kernel_set` that `TEVecSet` uses — read-modify-write rather than making the field mutable and storing into it, because the record's fields may legitimately be immutable and only the RMW shape is correct either way. The element and the index are each evaluated once, so `v.(f i).x <- e` does not call `f` twice.

## Interpreter

`VRecord` holds a **mutable** `value array`, so an in-place store is expressible; what was missing was name→index. `get_field` can only read, and `to_values`/`from_value` round-trips through the OCaml record and therefore **copies** — a store through it would vanish exactly the way Native's did.

So `HELPERS` gains `field_names`, generated by the PPX from the same `labels` list that already produces `to_values` and `get_field`. One source of truth, so a name's position **is** its index by construction rather than by convention. An index off by one is wrong data, not an error, which is why the new unit test checks `field_index` against `to_values`' actual output rather than against a hardcoded 0/1 — the two cannot be made to agree by editing only the test.

`read_lvalue` is deliberately separate from `eval_expr`: it **shares** rather than copying, which is what lets a nested `r.a.b <- v` land.

Registry lookup precedes the positional (`_0`/`_1` tuple) fallback, matching the read path — a declared record with a field literally named `_0` must resolve through its own declaration order.

Refusals now name the type and field instead of saying "not fully supported", and an unknown field is refused rather than defaulting to slot 0.

## Evidence

`test_record_field_store` asserts on **every** available device and checks the two neighbouring fields as well, because a dropped store is indistinguishable from a kernel that never ran unless the values are read back and compared. A run that finds zero devices exits non-zero rather than reading as a pass.

**18 cases green** — 9 devices × 2 record shapes (Interpreter ×2, Native, CUDA/PTX ×2 via ZLUDA, OpenCL ×2, Vulkan ×2).

**Prove-red, both halves, real output:**

Interpreter — `field_index` forced to `None`:
```
field-store [Interpreter]: FAILED (raised: Unsupported operation
  'record field assignment Test_record_field_store.triple.b': no such field ...)
Native, CUDA/PTX: still OK   <- the two halves are independent
```

Native silent path — the `TEVecGet` arm disabled:
```
@0 field target: got 1 want 2
@1 field target: got 2 want 4
```
That one needed **two** variables, stated plainly: the mutation *plus* removal of the immutable case, because that case turns the same mutation into a compile error which masks the silent drop. The compile error is the red for the immutable half — it just cannot be observed in the same build as the other.

## Summary by CodeRabbit

* **New Features**
  * Added in-place updates for record fields, including nested fields and records inside vector elements.
* **Changed**
  * Record type helpers now require ordered field names to preserve field/value alignment.
  * Local record and variant bindings are copied to prevent unintended aliasing.
* **Bug Fixes**
  * Improved nested-record backend support and diagnostics for unsupported kernel-declared record updates.
* **Tests**
  * Added comprehensive unit, end-to-end, and negative coverage for field stores, aliasing, field resolution, and cyclic values.


mathiasbourgoin:
Adversarial review: **NO-GO**. The depth-1 fix is real and proven red-on-revert, but the immediately adjacent lvalue form is still broken — and this PR's comments claim otherwise.

## `v.(i).f.g <- e` is still silently dropped on the CPU backends

Executed, on a probe kernel storing `v.(tid).op.ix <- 42.0`:

```
NESTED [Interpreter] Sequential:            SILENTLY DROPPED (op.ix = 0, want 42)
NESTED [Interpreter] Parallel, 32 cores:    SILENTLY DROPPED (op.ix = 0, want 42)
NESTED [Native]      Parallel, 32 cores:    SILENTLY DROPPED (op.ix = 0, want 42)
NESTED [OpenCL]      AMD RX 7900 XTX:       STORE LANDED
NESTED [OpenCL]      AMD Ryzen 9 7950X:     STORE LANDED
NESTED [Vulkan]      RADV NAVI31:           STORE LANDED
NESTED [Vulkan]      RADV RAPHAEL:          STORE LANDED
```

Six backends correct, PTX refuses loudly, **two CPU backends silently wrong** — a value divergence, which is the exact failure mode this PR exists to remove.

### Where

- `Sarek_ir_interp_eval.ml`, `read_lvalue`'s `LRecordField` arm calls `Sarek_type_helpers.get_field`, which is `H.get_field (H.from_value v)` — the `from_value`/`to_value` round trip through the OCaml record. That is the copying path the sibling comment twenty lines above warns is what loses the store. So the nested store lands only on synthesized/tuple records, where the positional fallback returns the shared `fields.(idx)`; on any `[@@sarek.type]` record it is lost.
- `Sarek_native_gen_expr.ml`'s new arm matches only `record.te = TEVecGet`. A nested store lowers to `TEFieldSet (TEFieldGet (TEVecGet ...), ...)` and falls into the generic `pexp_setfield` branch — a setfield on the fresh record `Vector.get` just marshalled out, i.e. the original bug verbatim.

The IR supports the nesting (`Sarek_lower_ir.ml:1275-1276` recurses through `TEFieldGet` into chained `LRecordField`) and the C-family emitters recurse structurally, which is why the GPU paths are fine.

### The part that makes this a NO-GO rather than a follow-up

The interpreter arm's comment asserts *"Nested: r.a.b <- v reads r.a here... Sharing all the way down is what makes the nested store land."* The gap is documented as coverage, so nothing downstream would ever look for it. That is the fourth time today a comment claimed more than the code it sits on.

## Prove-red — both hunks are load-bearing

| state | result |
|---|---|
| branch as-is | test exit 0, 7 devices |
| interpreter hunk reverted | exit 1 — `Unsupported operation 'record field assignment'` |
| native hunk reverted | exit 1 — `Error: The record field "b" is not mutable` |
| native hunk reverted, mutable-only probe | exit 1 — `SILENTLY DROPPED (mb=1 want 2)` |

One test-design note worth acting on: the committed test's red signal for the native hunk is a **compile error** from its immutable-record case, so the assertion that would catch the *silent* drop never executes in the red state. Isolating with a mutable-only probe was needed to confirm that path is genuinely pinned.

## Attacks that did not break it

`v.(i).b <- w.(i).a` (cross-vector rhs), a field store inside a `for` loop indexed by the loop variable, and a two-threads-one-element lost-update probe against Native's read-modify-write — all correct on all 7 devices. Index double-evaluation is genuinely prevented (`sarek_fs_idx` / `sarek_fs_vec` bound once before the get).

## Gates

`dune build` 0 · `dune runtest` 0 · `dune build @fmt` 0 · license 0 · KB 0 (33 declarations) · prove-red 0 (5 subjects, 25 mutations) · Alcotest registration 0 (598 files, 81 suites). All green — none of them see this class.

## To flip to GO

Handle the chained lvalue on both CPU paths, or refuse it LOUDLY there rather than dropping it; correct the two comments either way; and add a chained-lvalue case to the test, which currently has none.

mathiasbourgoin:
@coderabbitai review

This PR has completed an independent adversarial review of its current head and carries no outstanding blockers. CI is green with and without a CUDA/PTX device present.

Requesting a review so the PR can satisfy the repository's review requirement.

Worth knowing for the read: the round-5 commits correct two prose claims that earlier rounds got wrong, and deliberately *narrow* what the PR asserts to what was measured — including a note that the depth-65 refusal is read off the marshaller rather than observed, and that no committed test covers the 65-link chain.

mathiasbourgoin:
Heads-up from #394 (backlog-203, `fix/struct-typedef-dependency-order`): the nested-record struct-ordering gap this branch tolerates is **fixed** on `main` now — the declaration emitters topologically sort `kern_types`, and a read-only nested access verifies on OpenCL x2 and Vulkan x2 on the RX 7900 XTX host.

So before this lands, the tolerance needs to come out or it excuses a gap that no longer exists:

- `sarek/tests/e2e/test_record_field_store.ml` — `is_backlog203_struct_gap` and its device-free self-check (4 references), and the nested store required on every backend instead of excused.
- `sarek/tests/e2e/test_record_local_alias_agreement.ml` — the backlog-203 "no dependency sort" tolerance around the `l2` ordering.

I could not do it from #394: neither file exists on `main`, and they depend on this branch's interpreter/native changes, so they cannot be lifted across on their own.

mathiasbourgoin:
## Unblocked: rebased onto current `main`, CI failure diagnosed, backlog-203 tolerance deleted

Head is now `ea23235c` (force-with-lease over `564c0784`).

---

### 1. The failing check was not this PR's

`build (ubuntu-latest, 5.4.0, true)` failed in `scripts/check-workflow-toolchain-env.sh`:

```
.github/workflows/ci.yml:178: job 'build' runs `ocaml` on the HOST, but the job
provisions no host OCaml toolchain (no ocaml/setup-ocaml) — every toolchain
action here must go through `docker run ... spoc-ci:latest`
```

**This branch touches neither `.github/workflows/ci.yml` nor that script** — measured, `git diff --name-only <merge-base> 564c0784` lists neither. It was a two-hour window on `main` in which a guard and a workflow step disagreed:

| commit | authored (UTC) | what it did |
|---|---|---|
| `361dff0d` | 14:22 | added an apt `ocaml-nox` install + `ocaml -version` to ci.yml's `build` job |
| — | **15:27** | **the failing run** |
| `93b865f5` | 17:27 | taught the guard that apt provisioning counts, not only `ocaml/setup-ocaml` |

Reproduced exactly rather than inferred: running `93b865f5^`'s version of the guard against current `main`'s `ci.yml` reprints that message byte-for-byte at the same line, exit 1. With the rebase in place the current guard passes:

```
check-workflow-toolchain-env: OK — examined 26 host `run:` step(s) of 48 across
9 job(s) in 6 workflow file(s); none invokes the toolchain in a job that does
not provision one (2 job(s) skipped as ocaml/setup-ocaml provisions a host
switch; 1 step(s) provision a host tool via apt-get + a version assertion)
```

So the fix was already on `main` and the rebase is the fix. No workaround, and nothing to measure the cost of.

### 2. The conflicts, and evidence both parents survived

Rebased twice — `main` moved again mid-round (**#394 merged at 19:56:40Z**), so the second rebase is the one that ships.

Three conflicts, all the same shape: **two sides appending at one anchor over an empty common ancestor**. Each resolution asserted the ancestor was empty before merging, and kept both sides.

| file | ours (`main`) | theirs (this PR) |
|---|---|---|
| `sarek/tests/e2e/dune` | `test_record_nested_struct_order` stanza (#394) | `test_record_field_store` stanza |
| `CHANGES.md` | #394's `### Fixed` entry | this PR's two `### Fixed` entries |
| `sarek/tests/negative/dune` | `neg_test_ktype_advice` + `neg_test_sarek_include_advice` | `neg_test_inline_type_field_store` |

`rerere` is enabled with a cache in the shared git dir and **did** silently replay a resolution on the first rebase, so no clean tree was taken on faith. Evidence, per the both-parents rule:

- **Marker sweep**: `git grep -In '^<<<<<<< \|^>>>>>>> \|^||||||| '` over the whole tree — exit 1, no hits.
- **Every file `main` owns is byte-identical to `origin/main`.** `git diff --stat origin/main -- .github README.md benchmarks scripts specs formal docs kb spoc sarek/codegen sarek/ppx/Sarek_error.ml sarek/ppx/Sarek_lower_ir.ml sarek/tests/e2e/test_record_nested_struct_order.ml sarek/tests/negative/test_ktype_advice.ml sarek/tests/negative/test_sarek_include_advice.ml` → empty.
- **Restricted diffs on each shared file, against BOTH parents.** vs `origin/main`: additions only, exactly this PR's stanza/entry. vs the branch commit being applied: additions only, exactly `main`'s. Zero deletion lines in either direction on all three files.
- On the first rebase, the branch's whole payload (a 2726-line diff) was **byte-identical** pre- and post-rebase.
- `test_record_nested_struct_order.exe` — `main`'s own #394 test — exits 0 on this tree.

### 3. The backlog-203 tolerance: not correct earlier today, correct now, done

The precondition #394 recorded was verified against the repo rather than trusted, and the answer changed inside this round.

**When first checked, deleting would have been wrong.** #394 was OPEN, `sort_record_types_by_dependency` occurred nowhere on `main`, and the gap was live: 4 rows (OpenCL ×2, Vulkan ×2) took the known-gap branch. Deleting then would not have surfaced the gap — CI enumerates no GPU, so those rows never run there. CI would have stayed green while every OpenCL/Vulkan host went red. That reasoning is committed as `148bfa56`.

**#394 then merged.** Re-measured against it:

| | before #394 | after #394 |
|---|---|---|
| `test_record_field_store` rows tolerated as "known gap" | 4 | **0** |
| rows green | 24 | **28** |
| `test_record_local_alias_agreement` "not launched" skips | 4 | **0** |

So the tolerance became dead — the dangerous state, since it admits precisely the failure a regression of #394 produces. Deleted in `ea23235c`: `is_backlog203_struct_gap` + its 11-case self-check, both dialect error-line helpers, `declaration_offset`, `uses_type_before_declaring` + its 10-case self-check, `struct_emitting_frameworks`, `predict_struct_gap`, the `~inner` thread through `nested_case`, and the "no longer bites" note. In the sibling file the second copy went too, including `nested_can_compile` — a **skip** predicate, the worse half, which would have declined to launch on both shader backends forever regardless of codegen.

**Proven, not asserted**: unwiring #394's three call sites (definition kept, mutation asserted landed) now makes `test_record_field_store.exe` exit 1 with **4 hard failures** on exactly the rows that were previously tolerated. Before the deletion the same mutation exited 0 printing "known gap — expected".

### 4. The PR's substance still holds, both polarities proven red

Fix committed first, then mutated, mutation asserted to have landed, then restored.

- **Interpreter no longer refuses** — revert `Sarek_ir_interp_eval.ml` to `origin/main` (asserted: 0 `backlog-172`, 1 old refusal): exit 1, 8 Interpreter rows `FAILED (raised: Unsupported operation 'record field assignment': not fully supported)`.
- **Native no longer silently drops** — keep the functional rebuild, discard the write-back (asserted: 0 write-back sites, 2 drops). Exit 1, 4 Native rows with the silent-drop signature `@0 field target: got 1 want 2`. This is an **assertion**-level red; reverting the whole file instead gives only a build-level red (`The record field b is not mutable`), which is weaker, so the narrow mutation is the one reported.

### 5. Review

Both blockers are code the previous review saw, and the conflict resolutions edit reviewed code, so no verdict was carried. A full fan-out ran on the rebased tree: an unscoped fresh-read pass plus a cross-runtime pass (codex). Between them **twelve** findings, every one the house defect class — a claim wider than its code — and none a wrong algorithm. Fixed across `3e24e774` and `84c6b108`. Highlights:

- `Sarek_ir_interp_eval.ml` said the Native half "is still open"; it is fixed three commits earlier in the same range.
- "Cases 3 and 4 would not pass with the typedef arm removed" — case 4 **does**. Measured in both files. That arm was pinned in one direction only, so a direct `declaration_offset` assertion was added ahead of the case loop (ahead, because the loop exits first and an assertion behind it can never be observed failing) — then deleted with everything else in §3.
- Both refusal messages named `[%%sarek.type]`, wrong twice: `%%` renders as one `%`, and the construct bears no attribute. The correction then **overshot**, dropping the attribute from the *advice*, where it belongs — moving a type out of the kernel does nothing until `[@@sarek.type]` registers it. Four `@`s in the literal, verified on rendered output.
- New negative test `sarek/tests/negative/test_inline_type_field_store.ml`: those two refusals were reachable from no test at all, which is how both bad spellings shipped. The Makefile case greps diagnosis and advice **separately** — they have been wrong independently. Proven red on three mutations, each asserted landed first; two of them initially matched nothing and the gate read PASS, which is the false-separation shape.
- "no error on any path" was wider than the construct: Native had two modes, silent drop on a mutable field and a hard compile error on an immutable one.
- The lost-update note stated a **race** as a certainty. Now stated as a race, with the interleaving spelled out.
- CHANGES.md claimed the depth guard replaces "looping" on a cyclic value; it is `Stack_overflow` — and the implementation comment had already retired the "infinite loop" wording, which the changelog reintroduced.

Independent confirmation from `main`'s side: `scripts/check-ppx-construct-names.sh` — new on `main` and aimed at exactly this `%%`/`@@` class — passes on this tree (437 sources, 471 bracket-bearing literals, every named construct declared under `sarek/ppx/` and surviving Format rendering).

The conflict resolutions and the deletion are unreviewed code, so: **@coderabbitai review**

### 6. Gates, each exit code captured directly

| gate | exit |
|---|---|
| `dune build` | 0 |
| `dune runtest --force` | 0 — **5223 lines**, not a cached no-op |
| `dune build @fmt` | 0 (after `dune promote`) |
| `make test_negative` | 0 — 33 PASS, including the new case |
| `scripts/prove-red.sh` | 0 — 6 subjects, 33 mutations |
| `scripts/check-kb-properties.sh` + `.test.sh` | 0 / 0 |
| `scripts/check-cited-paths-exist.sh` + `.test.sh` | 0 / 0 |
| `scripts/check-ppx-construct-names.sh` + `.test.sh` | 0 / 0 |
| `scripts/check-review-bundle-tracked.sh` | 0 |
| `scripts/check-opam-clean.sh` | **1 — environmental, see below** |

`git status --porcelain` was empty before each commit, and `git checkout -- '*.opam'` ran before each. `git stash` was never used; foreign stashes from other branches are present in the shared git dir and were left untouched.

### 7. What is NOT verified

- **`check-opam-clean.sh` exits 1 here and I did not fix it.** Local dune 3.24.1 regenerates `sarek.opam` with `"dune" {>= "3.15" & >= "3.15"}` — a duplicated constraint. This branch modifies **neither** `dune-project` nor any `.opam` file (`git diff --name-only origin/main HEAD -- dune-project '*.opam'` is empty), and those files are generated from `dune-project` alone, so the output is a pure function of the local dune version and is identical on `origin/main`. I did not run a CI-version dune to confirm that, so it is argued from the file list, not measured.
- **No GPU in CI.** Everything device-shaped here — 28 rows across 7 devices, the 4-row before/after on the backlog-203 gap, both prove-red polarities, the #394-regression proof — was measured on this host (RX 7900 XTX + Ryzen 9 7950X iGPU, radeonsi/RADV). CI enumerates no device, so CI cannot reproduce any of it.
- **No CUDA/PTX row in the final run.** ZLUDA was not on `LD_LIBRARY_PATH`, so the run printed its `NOT MEASURED HERE` line for CUDA/PTX and the header's CUDA/PTX claim is not reproduced by that run.
- **Metal and HIP: nothing measured, nothing claimed.** No such device on this machine.
- **The Native lost-update race (backlog-207) is read off the code, not observed.** No kernel in the suite has two writers per element.
- **The depth-65 refusal remains read off the marshaller and `bind_var`, not observed**, and no committed test covers a 65-link chain. That honesty is intact and was checked for re-widening in both directions.

No review was dismissed, nothing was merged, and `--admin` was not used.

mathiasbourgoin:
## Round 6 — two findings fixed, one refuted by measurement and then pinned

Head is now `81a58314` (was `ea23235c`). Two commits: `beacb14f` addresses the review, `81a58314` addresses a cross-runtime pass over `beacb14f`. All 3 findings answered by name below.

---

### 1. MAJOR, `Sarek_ir_interp_eval.ml:814` — positional fallback under `Some h`. **FIXED** (`beacb14f`)

Confirmed against the code, and the in-code comment was already admitting it ("unreached in practice"). `read_lvalue`'s `LRecordField` arm resolves a registered type with `h.field_index` alone and raises on `None`; `assign_lvalue`'s went on to try `positional_field_index`. So for a registered record and a name that merely *looks* positional, the READ raised `record T has no field _0` and the WRITE landed in slot 0 reporting success.

Fallback removed under `Some h`; write now resolves exactly as read does. The `ERecordField` registry tier is deliberately left alone — adding it here would change which stores **succeed**, not just which fail.

Correcting one thing I wrote in that commit, caught by the cross-runtime pass: I justified the severity with "a dropped store leaves correct data". It does not — a dropped store leaves the *target* holding its old value, already wrong under the requested semantics. The real difference is blast radius: a dropped store corrupts the field the program was writing, a misdirected one corrupts a field nothing in the source mentions. Reworded in `81a58314`.

New test `sarek/tests/unit/test_interp_field_store_resolution.ml`, registered mock with `field_names = ["a"; "b"]` — no `_0`, so the divergence is reachable. **Proven red**, mutation applied by asserted *line position* (the two arms now carry identical text, so a `perl -0pi` for it hits `read_lvalue` first — which is exactly what my first attempt did, reddening only case 3 and reading as a successful proof of case 1):

| mutation | case 1 | case 2 | case 3 |
|---|---|---|---|
| restore fallback in `assign_lvalue` | **FAIL** `Expected 1, Received 99` | OK | **FAIL** `Expected refused, Received slot 0` |
| restore fallback in `read_lvalue` | OK | OK | **FAIL** `Expected slot 0, Received refused` |

The slot assertion is ordered *before* the raise assertion on purpose: Alcotest stops at the first failure, and "slot 0 holds 99" names the defect where "it did not raise" only names a missing diagnostic.

`81a58314` also strengthened case 3 after codex pointed out that "read and write **agree**" compared only whether each raised — two resolvers landing on different slots agree on every boolean. It now compares `Refused | Slot i`, with the slot recovered by observation (distinct seeds identify the read's slot; a sentinel identifies the write's).

### 2. MAJOR, `"CUDA/PTX"` is not a real `Device.framework` value. **REFUTED**, then pinned

It is a real value. `sarek/core/Device.ml:30-38` — `resolve_framework` expands a **family** name to registered backend names; the comment at `:138` states it outright: *"CUDA backends register as `CUDA/PTX` and `CUDA/C`"*. `Device.init ~frameworks` accepts the family `"CUDA"` as a **request**; `d.Device.framework` — which is what `frameworks_seen` is built from — is the resolved **backend** name `"CUDA/PTX"`. Two vocabularies; only the second can appear on the left of that `List.mem`.

Settled by measurement, both directions, on this host:

| | devices | frameworks reported | NOT-MEASURED lines |
|---|---|---|---|
| ZLUDA on `LD_LIBRARY_PATH` | **9** | `CUDA/PTX, Interpreter, Native, OpenCL, Vulkan` | **0** |
| without | 7 | `Interpreter, Native, OpenCL, Vulkan` | 1 |

Rows read `field-store immutable [CUDA/PTX] AMD Radeon RX 7900 XTX [ZLUDA]: OK`. So the predicate is true when it should be true and false when it should be false, and **the previously-missing CUDA/PTX row was ZLUDA genuinely absent, not a predicate that could never be true** — the spot checks used a bare `dune exec`, while the dune *rule* wraps the test in `(setenv LD_LIBRARY_PATH %{env:HOME=}/opt/zluda ...)`, so `dune runtest` has been exercising the CUDA leg all along. With it on the path the suite reports **36 OK rows across 9 devices**.

Applying the suggestion would have **created** the defect it describes: with the key changed to `"CUDA"`, `List.mem "CUDA" frameworks_seen` is false on a ZLUDA run where the leg executed and passed — a loud named skip over a measured leg, indistinguishable in the log from an honest one.

Pinned in both files so it cannot be reintroduced. That pin took three tries, and the two failures are the interesting part:

- **v1** required each key to be in `Framework_registry.all_backend_names ()`. Measured: `Cuda_plugin.init ()` registers nothing without a CUDA driver, so `"CUDA/PTX"` is absent from the registry on a driver-less host and this hard-failed there. It traded a false green for a false red.
- **v2** compared against enumerated devices — a key absent from `frameworks_seen` whose *family* was enumerated. Caught `"CUDA"`, but also rejected `"CUDA/C"`, a valid backend name, whenever only `CUDA/PTX` enumerates. Overshoot one step further out.
- **v3** asks the question that separates the vocabularies: is the key a **family name** — not itself registered, but the prefix before `/` of something that is?

Five polarities measured on v3:

| key | ZLUDA | result |
|---|---|---|
| `CUDA/PTX` | yes | exit 0, 9 devices, no skip line |
| `CUDA/PTX` | no | exit 0, 7 devices, honest skip |
| `CUDA` | yes | **exit 1** — *"is a FAMILY name, not a backend name: it is not itself registered, but CUDA/PTX is"* |
| `CUDA` | no | exit 0, no false red |
| `CUDA/C` | yes | exit 0 + honest skip — the v2 overshoot, gone |

Bounded honestly in the comment: it catches the vocabulary confusion, **not** a plain misspelling (`"Vulkann"` is indistinguishable at runtime from a framework that did not enumerate), and it **cannot fire in CI**, which registers no GPU plugin.

### 3. MINOR, `Makefile:123` — missing `@echo`. **FIXED** (`beacb14f`)

`make test_negative` now prints `Testing inline-type field-store rejection and its rendered advice (backlog-172)...` before the check. 33 PASS.

---

### Correction to an earlier comment on this PR

My round-5 comment listed `check-opam-clean.sh` as exit **1**, "environmental", and attributed it to local dune 3.24.1. **That entry was wrong to publish and I withdraw it.** Measured properly:

- ambient `dune` **3.23.0** → `check-opam-clean.sh` exit **0**, `PASS: 'make opam' is clean and idempotent`
- the project switch `/home/mathias/dev/SPOC/_opam`, `dune` **3.24.1** → exit **1**, `sarek.opam` regenerated with `"dune" {>= "3.15" & >= "3.15"}`

And the 3.23.0 pass is a real pass, not a vacuous one — I appended a canary to `sarek.opam` and both dune versions removed it, so both genuinely regenerate the file. The check passes at the dune this tree uses; the exit 1 came from building under a different switch. Nothing here is a defect in this branch, and the branch touches neither `dune-project` nor any `.opam`.

### Gates, each exit code captured directly

`dune build` 0 · `dune runtest --force` 0 (**5232 lines**) · `dune build @fmt` 0 · `make test_negative` 0 (33 PASS) · `prove-red.sh` 0 (6 subjects, 33 mutations) · `check-kb-properties.sh` +`.test.sh` 0/0 · `check-cited-paths-exist.sh` +`.test.sh` 0/0 · `check-ppx-construct-names.sh` +`.test.sh` 0/0 · `check-license-headers.sh` 0 · `check-negative-case-coverage.sh` 0 · `check-test-alias-coverage.sh` 0 · `check-dune-dir-visibility.sh` 0 · `check-workflow-steps.sh` 0 · `check-workflow-toolchain-env.sh` 0 · `check-review-bundle-tracked.sh` 0 · `check-alcotest-registration.js` 0.

`git status --porcelain` empty before each commit; `git checkout -- '*.opam'` before each; `git stash` never used.

The PR's own two polarities were re-proved on the rebased tree after the interpreter change: reverting the interpreter fix gives exit 1 with 8 Interpreter rows raising the original `Unsupported operation 'record field assignment'`; discarding the Native write-back gives exit 1 with 4 Native rows showing the silent-drop signature `got 1 want 2`.

### Not verified

- **CI has no GPU.** The 36 rows across 9 devices, the 9-vs-7 CUDA measurement, all five self-check polarities and both prove-red polarities were measured on this host (RX 7900 XTX + Ryzen 9 7950X iGPU, radeonsi/RADV, CUDA via ZLUDA). CI reproduces none of it. The new framework self-check specifically cannot fire there.
- **Metal and HIP**: no such device here, nothing claimed.
- **The Native lost-update race (backlog-207)** is still read off the expansion, not observed; no kernel in the suite has two writers per element.
- **The depth-65 refusal** remains read off the marshaller and `bind_var`, with no test covering a 65-link chain.

`@coderabbitai review`

Issue #75, opened 2026-01-11
Fix missing Jekyll setup documentation


mathiasbourgoin:
## Problem

The benchmark preview Makefile target (`make bench-preview`) requires Jekyll to be installed and configured, but there's no documentation explaining how to set it up. Users trying to preview benchmark results locally will encounter errors.

## Current State

`Makefile` line 426 runs:
```bash
@cd gh-pages && bundle exec jekyll serve
```

But there's no:
- Installation instructions for Jekyll/Ruby
- `Gemfile` setup instructions  
- Bundle install steps
- Prerequisites documentation

## Required Changes

1. **Add `gh-pages/README.md`** with setup instructions:
   - Ruby installation (version requirements)
   - Bundler installation: `gem install bundler`
   - Bundle config: `bundle config set --local path 'vendor/bundle'`
   - Dependencies install: `bundle install`
   - Running server: `bundle exec jekyll serve`

2. **Update root `README.md`** (optional):
   - Add "Previewing Documentation" section
   - Link to `gh-pages/README.md`

3. **Verify `gh-pages/Gemfile`** exists and is correct

## Files to Update

- `gh-pages/README.md` (create)
- `README.md` (optional section)
- Verify: `gh-pages/Gemfile`, `gh-pages/Gemfile.lock`

## Related

- Makefile target: line 426
- PR #68 benchmark infrastructure

mathiasbourgoin:
Resolved by PR #81 - Added comprehensive Jekyll setup documentation

PR #379, opened 2026-07-29
fix(lower-ir): seed constant-prefixing with helper PARAMETER names (H3)

mathiasbourgoin:
backlog-180 (review finding H3). Verified at the lines before changing anything.

`Sarek_lower_ir.ml` collected the helper's referenced-constant set with `stmt_names fun_body_ir [] acc` — an **empty** bound list. So a helper parameter that merely shares a module constant's name made that constant look free, and backlog-160's prefixing pass inserted its `SLet` into a body that already binds the identifier as a parameter.

The `expr_names` header documents this exact redeclaration bug for **local** binders (`SLet`/`SLetMut`/`SFor`, "caught by review on #362"). That fix landed for locals and stopped one binder class short. And the refusal message on the same path advises *"pass the constant in as a parameter"* — precisely the unchecked case, so the guidance walked the user into the defect.

## The reported backend split was inverted, and the truth is worse

Measured on the pre-fix code:

| Backend | Pre-fix |
|---|---|
| Interpreter ×2 | **OK** — correct; does *not* ignore its argument |
| Native | **OK** |
| CUDA/PTX ×2 | **`got 200 want 2` — silently wrong data** |
| OpenCL ×2 | compile failure (loud) |
| Vulkan ×2 | compile failure (loud) |

PTX is the silent one because it doesn't emit a flat C scope of named locals — it allocates registers, so a duplicated `SLet` never collides textually. The prefixed constant overwrote the parameter's register and the helper computed `100.0 *. 2.0` for every element.

This changes the value of the fix. A silent bug on the *interpreter* would be caught by any cross-backend differential test, since the interpreter is the oracle and would disagree with the GPUs. Here the oracle is **right** and only a real GPU backend is wrong — so agreement-with-interpreter is exactly the check that catches it, and no test of this shape existed.

Metal and HIP unmeasured (no such device on this host).

## Two changes, and the second is not redundant

- `referenced` — seeded with `param_names`, so a shadowed constant is never counted as referenced.
- `helper_binders` — `param_names` added too. The transitive closure also pulls in constants named by a referenced constant's **initializer**, and those enter the set regardless of what the parameters shadow. Present here, that residual case takes the existing refusal path instead of emitting the collision.

`param_names` is read off `params` (in scope since line 770) rather than `hf_params` (not built until ~929); both derive from the same list, so no second source of truth.

**A property worth stating:** reverting *only* the seeding leaves `helper_binders` as a backstop, and the collision surfaces as the existing located refusal rather than wrong data. Reproducing the original symptoms needed both halves reverted. So if a future binder class is missed again, the failure mode is a refusal, not a silent miscompile.

## Evidence

`test_helper_param_const_collision` — **18 cases green** (9 devices × 2 kernels).

Two kernels, both required. The collision case's values are deliberately two orders of magnitude apart (argument-derived 2.0 vs constant-derived 200.0) so the silent symptom is visible at all. The second kernel references a constant it does *not* shadow and must still receive its copy — without it, "never prefix anything" would also pass the first case while breaking backlog-160 outright.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed helper code generation when a helper parameter shares a name with a module constant.
  * Ensured parameter values take precedence while referenced constants remain correctly qualified.

* **Tests**
  * Added end-to-end coverage for name-collision scenarios across supported execution backends.
  * Added validation for both shadowed parameters and correctly referenced module constants.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,8 @@ test_negative:
 	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_char_vector.cma > "$$out" 2>&1; if grep -q "is not a supported Sarek element type" "$$out"; then echo "  PASS: char element type rejected"; else cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
 	@echo "Testing polymorphic-helper instantiation diagnostic (names the callee)..."
 	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_poly_helper_f64.cma > "$$out" 2>&1; if grep -q "'norm' cannot be used at this call site" "$$out"; then echo "  PASS: polymorphic helper instantiation named the callee"; else cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
+	@echo "Testing helper-parameter / module-constant collision (backlog-180 H3)..."
+	@out=$$(mktemp); dune build --profile=negative sarek/tests/negative/neg_test_helper_param_shadows_const.cma > "$$out" 2>&1; if grep -q "has a parameter named" "$$out" && grep -q "is NOT a fix here" "$$out"; then echo "  PASS: helper-parameter/module-constant collision rejected, naming the parameter"; else echo "  FAIL: expected the parameter-collision wording (and its negated advice); got:"; cat "$$out"; rm -f "$$out"; exit 1; fi; rm -f "$$out"
 	@echo "All negative tests checked (see KNOWN-ISSUE line above for the one non-blocking gap, if any)"
 
 
--- a/sarek/ppx/Sarek_lower_ir.ml
+++ b/sarek/ppx/Sarek_lower_ir.ml
@@ -882,18 +882,32 @@ let rec lower_expr (state : state) (te : texpr) : Ir.expr =
                [SLet] flat, so the prefix and the rebinding are two declarations
                of one identifier in one block. Refused below rather than emitted.
                Caught by review on #362. *)
-            let helper_binders =
+            let body_binders =
               let acc = Hashtbl.create 8 in
               stmt_binders fun_body_ir acc ;
-              (* Parameters too, and this is not redundant with seeding
-                 [referenced] above. That seeding only covers a constant named
-                 DIRECTLY by the body. The transitive closure below also pulls in
-                 constants named by the INITIALIZER of a referenced constant, and
-                 those enter [referenced] regardless of what the parameters
-                 shadow — so a constant reachable only through another
-                 constant's initializer could still collide with a parameter.
-                 Present here, that case takes the refusal path instead of
-                 emitting the collision. *)
+              acc
+            in
+            (* Parameters too, and this is not redundant with seeding
+               [referenced] above. That seeding only covers a constant named
+               DIRECTLY by the body. The transitive closure below also pulls in
+               constants named by the INITIALIZER of a referenced constant, and
+               those enter [referenced] regardless of what the parameters
+               shadow — so a constant reachable only through another constant's
+               initializer could still collide with a parameter. Present here,
+               that case takes the refusal path instead of emitting the
+               collision.
+
+               Kept as a SEPARATE table from [body_binders] rather than merged
+               into one, because the refusal message has to say which kind of
+               binder collided and what to do about it, and those differ. For a
+               body-local the advice is "rename the local, or pass the constant
+               in as a parameter"; for a parameter that same advice is circular —
+               the parameter IS the thing it would tell you to add. An earlier
+               revision merged the two and emitted the body-local wording for
+               both, so the parameter case was told to fix the collision by
+               creating it. *)
+            let param_binders =
+              let acc = Hashtbl.create 8 in
               List.iter (fun n -> Hashtbl.replace acc n ()) param_names ;
               acc
             in
@@ -901,7 +915,34 @@ let rec lower_expr (state : state) (te : texpr) : Ir.expr =
               List.fold_left
                 (fun body name ->
                   if not (Hashtbl.mem referenced name) then body
-                  else if Hashtbl.mem helper_binders name then
+                  else if Hashtbl.mem param_binders name then
+                    (* The colliding binder is a PARAMETER. Note the helper does
+                       not necessarily name the constant itself: the seeding
+                       above removes a directly-named one from [referenced], so
+                       the way a parameter-named constant still gets here is
+                       through the initializer of some OTHER constant the helper
+                       does reference. Hence "directly or through" rather than a
+                       flat "references", which would be a claim wider than what
+                       was measured. *)
+                    Ppxlib.Location.raise_errorf
+                      ~loc:helper_loc
+                      "Helper %S has a parameter named %S, and module constant \
+                       %S is needed inside the helper (either named there \
+                       directly or reached through the initializer of another \
+                       constant the helper uses). To be visible, the constant \
+                       has to be declared inside the helper body, and the \
+                       generated device code declares locals in one flat scope \
+                       alongside the parameters — so that would put two \
+                       declarations of %S in the same block. Rename the \
+                       PARAMETER %S, or rename the module constant. Note that \
+                       passing the constant in as a parameter is NOT a fix \
+                       here: a parameter of that name is what collides."
+                      name_of_helper
+                      name
+                      name
+                      name
+                      name
+                  else if Hashtbl.mem body_binders name then
                     Ppxlib.Location.raise_errorf
                       ~loc:helper_loc
                       "Helper %S both references module constant %S and binds \
--- a/sarek/ppx/Sarek_lower_ir.ml
+++ b/sarek/ppx/Sarek_lower_ir.ml
@@ -818,9 +818,32 @@ let rec lower_expr (state : state) (te : texpr) : Ir.expr =
                actually referenced are prefixed — prefixing all of them would
                evaluate initializers the helper does not need and, worse, would
                drag an unreferenced barrier into the refusal path. *)
+            (* The helper's PARAMETER names. They are binders exactly as much as
+               [SLet]/[SLetMut]/[SFor] are, and the [expr_names] header above
+               documents the redeclaration bug for those — the fix landed for
+               locals (review on #362) and stopped one binder class short.
+
+               Read off [params] here rather than [hf_params], which is not built
+               until further down; both come from the same list, so there is no
+               second source of truth. *)
+            let param_names =
+              List.map (fun (p : tparam) -> p.tparam_name) params
+            in
             let referenced =
               let acc = Hashtbl.create 8 in
-              stmt_names fun_body_ir [] acc ;
+              (* Seeded with the parameter names, NOT [] — a parameter that
+                 merely shares a module constant's name made the constant look
+                 referenced, so its [SLet] got prefixed into a body that already
+                 binds that identifier as a parameter. Two declarations of one
+                 name in one flat scope: a hard compile error on all five device
+                 backends for a helper that compiled before, and on the
+                 interpreter the prefix overwrites the parameter, so the helper
+                 silently ignores its argument.
+
+                 The refusal message below advises "pass the constant in as a
+                 parameter", which was precisely the unchecked case — the
+                 guidance walked the user into the defect. *)
+              stmt_names fun_body_ir param_names acc ;
               let rec close () =
                 let added = ref false in
                 Hashtbl.iter
@@ -854,6 +877,16 @@ let rec lower_expr (state : state) (te : texpr) : Ir.expr =
             let helper_binders =
               let acc = Hashtbl.create 8 in
               stmt_binders fun_body_ir acc ;
+              (* Parameters too, and this is not redundant with seeding
+                 [referenced] above. That seeding only covers a constant named
+                 DIRECTLY by the body. The transitive closure below also pulls in
+                 constants named by the INITIALIZER of a referenced constant, and
+                 those enter [referenced] regardless of what the parameters
+                 shadow — so a constant reachable only through another
+                 constant's initializer could still collide with a parameter.
+                 Present here, that case takes the refusal path instead of
+                 emitting the collision. *)
+              List.iter (fun n -> Hashtbl.replace acc n ()) param_names ;
               acc
             in
             let fun_body_ir =
--- a/sarek/ppx/Sarek_lower_ir.ml
+++ b/sarek/ppx/Sarek_lower_ir.ml
@@ -835,10 +835,18 @@ let rec lower_expr (state : state) (te : texpr) : Ir.expr =
                  merely shares a module constant's name made the constant look
                  referenced, so its [SLet] got prefixed into a body that already
                  binds that identifier as a parameter. Two declarations of one
-                 name in one flat scope: a hard compile error on all five device
-                 backends for a helper that compiled before, and on the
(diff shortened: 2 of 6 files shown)


mathiasbourgoin:
Adversarial review: **NO-GO**. The one-line fix is right. The 60 lines shipped with it are not, and two committed claims are refuted by this PR's own test.

## What is correct

The `param_names` seeding is the real fix and it has a genuine positive control. At true baseline (`git checkout 5f5ce85f -- sarek/ppx/Sarek_lower_ir.ml`, keeping the new test) the test exits **1** with real device-compiler errors: OpenCL `input.cl:4:9: error: redefinition of 'scale'`, glslang `'scale' : redefinition`. Minimal and correct.

## Blocker 1 — two committed claims that this PR itself refutes

The seeding comment states the collision is "a hard compile error on all five device backends … and on the interpreter the prefix overwrites the parameter, so the helper silently ignores its argument."

Measured on this host: **Interpreter ×2 OK, Native OK**; OpenCL ×2 and Vulkan ×2 give the compile error. The PR's *own later commit* (`84fde09b`, "the reported backend split was inverted") and its test header already say so — the comment was never updated to match. So the PR description's "broke working kernels on all 5 backends" is contradicted by the PR's own test.

Separately, the test header states "MEASURED pre-fix behaviour, per backend, **on this host**: … CUDA/PTX x2 got 200, want 2 ← SILENTLY WRONG DATA". There is no CUDA device on this host — no `nvidia-smi`, and `Device.init` with CUDA requested returns 7 devices, none CUDA. That unverified claim is then baked into a new **user-facing error message** ("on PTX a silently overwritten register (measured: backlog-180)").

## Blocker 2 — `path_duplicate_binders` has zero tests

60 lines introducing a new hard `Location.raise_errorf` that rejects user kernels, with one call site, no test file referencing it, no positive control that it ever fires, and no negative control against false positives.

## Blocker 3 — its stated scoping model is false for the backend whose symptom is silent

The header asserts "`SIf` and `SMatch` DO brace their arms". True on CUDA/OpenCL/GLSL/WGSL/Metal. **False on PTX** — `Sarek_ir_ptx_stmt.ml:120-139` emits `SIf` as predicate + branch with `env_bind` and no save/restore, so both arms bind into one flat env. Also false on the interpreter (`Sarek_ir_interp_eval.ml:341-343, 378-381`, no restore). The PR's own new error text says PTX's failure mode is "a silently overwritten register" — so the checker models braces that PTX does not have, on precisely the backend where the consequence is silent.

## Blocker 4 — four proven false negatives

Running the algorithm verbatim against a mock IR. Claimed polarities hold (sibling `SIf` arms → `[]`, sibling `SMatch` arms → `[]`, single-path duplicates and seed collisions reported). But:

- `SSeq [SSeq [SLet x]; SLet x]` → `[]`, should be `["x"]`
- `SSeq [SPragma (SLet x); SLet x]` → `[]`, should be `["x"]`

`SSeq` and `SPragma` emit **FLAT on every backend** (CUDA:393/521, OpenCL:440, GLSL:1164/1285, WGSL:910/1067, Metal:591, PTX:21, interp:337), so the inner declaration *is* visible to the following sibling. Root cause: `go_seq`'s `| other -> go bound other; bound` discards the inner sequence's bindings, and `go`'s `SSeq` arm `|> ignore`s them. Latent today only because the prefix shapes happen to be right-nested (`Sarek_lower_ir.ml:1003, 1607` put the inner `SSeq` last) — but this is advertised as *mechanically enforcing* the flat-scope invariant, and it does not.

**`SMatch` pattern binders are uncovered in all three functions.** `pattern = PConstr of string * string list`, where the `string list` *is* the binder set. `stmt_names`' own header admits it; `stmt_binders:410` discards the pattern; the new `path_duplicate_binders:453` discards it too. This repo already ships `test_ematch_payload_e2e.ml` (#75) documenting that match-binder collisions produce silent wrong answers on every C-family backend. The class is currently **latent, not live** — a `[@@sarek.type]` variant used inside a helper fails to compile at all (`Error: Unbound module "Zz_probe_match_binder"`), though it works in a kernel body. The door is shut by a different limitation, not by this fix.

No tailrec/defunctionalization gap: those rewrite the typed AST into ordinary `TELet`/`TELetMut` (`Sarek_tailrec_elim.ml:306, 436`), so they lower to covered constructors.

## Blocker 5 — the two hunks are entangled, so neither is independently pinned

Reverting *only* the `param_names` seeding does not turn the test red — it turns `dune build` red (exit 1) with a lowering refusal emitted by the *other* hunk. Reverting only the duplicate-binder hunk catches nothing beyond what seeding already covers.

## Minor

A 9-line comment block is duplicated verbatim, the second copy dangling before `acc` (lines 928-946).

## Gates

`dune build` 0 · `dune runtest` 0 · `dune build @fmt` 0 (exit captured directly, not after a pipe) · license 0 · KB 0 · prove-red 0.

## To flip to GO

Split it. The seeding hunk alone is a clean GO. The checker needs its own tests (both polarities), a scoping model that is true for PTX and the interpreter or an explicit statement that it does not cover them, the flat `SSeq`/`SPragma` cases fixed, and `SMatch` binders either handled or documented as out of scope. And both misstated claims corrected — including the one now embedded in a user-facing error message.

mathiasbourgoin:
Split, as the review recommended, and both misstated claims corrected. Force-pushed to `0bcf916e`.

## What left the branch

The flat-scope binder checker (`fba8025d`, ~60 lines) is **removed** and parked on `wip/flat-scope-binder-checker`. It goes nowhere near main until it has tests, a scoping model that is true for PTX and the interpreter, the flat `SSeq`/`SPragma` cases fixed, and `SMatch` binders either handled or documented as out of scope.

## What stayed, and the two claims that were wrong

**1. The source comment said the opposite of the measurement.** `Sarek_lower_ir.ml` asserted "a hard compile error on all five device backends … and on the interpreter the prefix overwrites the parameter, so the helper silently ignores its argument". Measured: OpenCL and Vulkan error **loudly**, Interpreter and Native are **correct**, PTX is the silent one. Commit `84fde09b` corrected the *test header* for exactly this and left the source comment untouched, so the two have been contradicting each other on this branch ever since. Corrected.

**2. "On this host" was doing too much work — and this corrects the review too.** The test header claimed the CUDA/PTX row was "MEASURED … on this host". The review said flatly there is no CUDA device here. Both were partly right, so I measured it:

```
bare run          -> [Interpreter] [Native] [OpenCL] [Vulkan]              (no CUDA)
LD_LIBRARY_PATH=$HOME/opt/zluda -> [CUDA/PTX] [Interpreter] [Native] [OpenCL] [Vulkan]
```

So the row is real — a genuine execution of the genuinely emitted PTX — but through ZLUDA's CUDA implementation on an AMD RX 7900 XTX, not through NVIDIA's driver. That belongs in the claim, and the header now says so, with the verification inline. The review's "no CUDA device on this host" was correct for a bare run and incomplete for this one.

## Prove-red on the reduced branch, and an honest limit

Reverting only `stmt_names fun_body_ir param_names acc` to `[]`: `dune build` exits **1**. The fix is load-bearing.

But the red signal is the **refusal guard** firing —

> Kernel lowering failed: Helper "boost" both references module constant "scale" and binds a local of the same name.

— not the test's device assertions. The build stops before any device runs, so the per-backend table in the header documents a measured past state and is **not** re-verified on each run. This is the same shape as the finding on #376, and I found it only because dropping the checker changed which hunk produces the red. Worth a follow-up: the test needs a case whose red path reaches execution.

A second thing worth recording: the refusal guard turns out to live in the seeding commit, not in the checker I removed. So "the seeding hunk alone" is slightly more than a one-liner — it is the seeding plus a new hard rejection path. That rejection is at least exercised (it is what goes red above), unlike the checker's.

## Gates

`dune build` 0 · `dune build @fmt` 0 · `check-license-headers.sh` 0 · `check-kb-properties.sh` 0 · test 0 (18 device rows under ZLUDA).

## Status

Not merging on the strength of the earlier review. It concluded "the seeding hunk alone would be a clean GO", but it did not know the guard was bundled with the seeding or that the red path is a build error — so this state has not been reviewed as a whole. Attested in the local ledger as *conditional*, and it wants either a fresh pass or an explicit call.

mathiasbourgoin:
## CodeRabbit triage + the whole-state re-assessment the split left undone

Rebased onto `origin/main` (main moved today: #388, #387, #370). **No conflicts**, and `git range-diff` shows all three pre-existing commits identical after the rebase.

---

## 1. The CodeRabbit comment — **CONFIRMED at all three sites, all fixed**

All three still described the split that `84fde09b` corrected in the test header and nowhere else. Re-measured rather than taken from the header:

- **`test_helper_param_const_collision.ml:90-92`** — said "`100.0 *. 2.0` on the interpreter and a redeclaration error on the rest". Both halves wrong. Rewritten to the measured split.
- **`test_helper_param_const_collision.ml:181-184`** — claimed the pre-fix Interpreter produced `200.0`. It did not; the Interpreter was correct. The `200.0` was **CUDA/PTX**. Claim removed and re-attributed, labelled ZLUDA.
- **`sarek/tests/e2e/dune`** — blanket "hard compile error on the device backends, and on the interpreter the prefix overwrote the parameter". Replaced with the per-backend table including the ZLUDA qualification.

**Every CUDA/PTX row is labelled ZLUDA-on-AMD.** There is no NVIDIA device on this host, and I re-verified the device list rather than trusting the header: bare run enumerates 7 devices `[Interpreter ×2, Native, OpenCL ×2, Vulkan ×2]` with **no CUDA**; under `LD_LIBRARY_PATH=$HOME/opt/zluda` it is 9, adding `CUDA/PTX ×2 [ZLUDA]`.

---

## 2. The bundled refusal guard — the earlier review's "the seeding hunk alone would be a clean GO" was **wrong**, and I can show it

The `Location.raise_errorf` itself is pre-existing (#362). What this PR added is `param_names` into the binder table — which **widens that pre-existing refusal to a new class of input**: source that used to build is now hard-rejected. So "seeding plus a one-liner" is seeding plus a new rejection path, exactly as suspected.

Measured, not reasoned. Reverting **only** the seeding and keeping the binder hunk:

```
$ dune build sarek/tests/e2e/test_helper_param_const_collision.exe   # exit 1
Error: Kernel lowering failed: Helper "boost" both references module constant
"scale" and binds a local of the same name. ... Rename the local, or pass the
constant in as a parameter of "boost".
```

### Is the message true? **No — it was false on the path this PR opened.**

`scale` there is a **parameter**, not a local. And the advice — *"pass the constant in as a parameter"* — names the exact construct that collides, so a user who follows it cannot succeed. The comment beside the seeding already said the old guidance "walked the user into the defect", and the message was left unchanged.

### Is it reachable in the shipped state, or only under a revert? **Reachable as shipped.**

This is the part that made it a real defect rather than a revert artifact. The seeding removes a **directly** named constant from `referenced`, but the transitive closure calls `expr_names init [] sub` — empty bound list — so a constant reached through **another constant's initializer** still arrives at the guard carrying a parameter's name. Built against this branch, unmodified:

```ocaml
let (scale : float32) = 100.0 in
let (base  : float32) = scale +. 1.0 in          (* initializer names scale *)
let f (scale : float32) : float32 = base +. scale in   (* parameter is scale *)
```
→ refused, with the false "binds a local" wording.

### Does it fire on shapes it should not? **No — checked the other polarity.**

A parameter shadowing a constant that no referenced initializer mentions still builds (exit 0):

```ocaml
let (scale : float32) = 100.0 in
let (other : float32) = 5.0 in                  (* does NOT name scale *)
let f (scale : float32) : float32 = other +. scale in
```
So the closure is not over-approximating. The refusal's **outcome** was right in every case I could construct; only its explanation was wrong.

### Fix

Split the single binder table into `param_binders` and `body_binders` so the message can name the binder kind and give advice that works. The parameter branch names the parameter, says the reference may be transitive rather than direct, and states explicitly that adding a parameter is not the fix. The body-local branch keeps its existing wording. Both branches exercised; each routes to its own message.

### The refusal now has a test

It fires during lowering, so nothing running can observe it — it goes in `sarek/tests/negative/test_helper_param_shadows_const.ml`, asserted by `make test_negative`. The assertion pins the **message**, both that it names the parameter and that it carries the negated advice; "some error fired" would have been satisfied by the wording it exists to reject. `scripts/check-negative-case-coverage.sh`: `OK: 30 negative case(s) declared, all asserted by 'make test_negative'.`

---

## 3. A red path that reaches EXECUTION — it exists, and the header now says which mutation gets there

Three mutations, all run under ZLUDA with all 9 devices live:

| Mutation | Reaches execution? | Result |
|---|---|---|
| Revert **both** halves (= `main`) | **Yes** | exit 1. CUDA/PTX ×2 `got 200 want 2`; OpenCL ×2 `redefinition of 'scale'`; Vulkan ×2 compile failure; Interpreter ×2 + Native pass |
| Revert **seeding only** | **No** | stops at a BUILD error — the guard masks it |
| Seeding → `ignore` ("never prefix") | **Yes** | exit 1 on the prefixing case: `Unbound variable 'offset'` (Interpreter), `PTX codegen: unbound variable: offset` (CUDA/PTX), compile failure (OpenCL/Vulkan) |

So both existing cases **do** have execution-reaching red paths — the full-revert mutation reproduces the header table exactly, which is what makes that table evidence rather than an assertion. The header now records all three, so the per-backend table is no longer a measured-once state with nothing tying it to a runnable mutation.

Two things that fell out of measuring instead of assuming:

- **Native passes the "never prefix" mutation** — it resolves `offset` from the enclosing OCaml scope. Native alone is blind to a missing prefix; case 2's value comes from the other backends. Recorded in the header.
- **My own negative test was initially a gate that could not fail.** Proving it red the second way (guard removed entirely) produced `Unbound module "Spoc"` rather than a clean compile — it had copied `open Spoc` and `Kirc` from its sibling negative tests, and **neither name is in those tests' library dependencies**. They only get away with it because the refusal fires before those names resolve. So the one outcome the case exists to detect was reported as an unrelated setup error. Rewritten to depend only on what it uses; both polarities re-measured and now fail informatively:
  - guard absent → compiles, exit 0 → assertion fails on missing wording
  - parameter branch routed back to the body-local wording → assertion fails on wording it must not accept

  The sibling cases share that latent shape. **Not fixed here** — out of scope, and none of them is currently misreporting. Flagging it.

---

## Gates — every exit code captured directly, never `$?` after a pipe

| Gate | Exit |
|---|---|
| `dune build` | 0 |
| `dune runtest` | 0 |
| `dune build @fmt` | 0 |
| `scripts/check-kb-properties.sh` | 0 |
| `make test_negative` | 0 |

`check-cited-paths-exist.sh` run because the header now cites the new negative test by path: `OK — every repo-relative path cited in 645 tracked .ml/.mli files resolves to a tracked file`.

## What I could NOT verify

- **Metal and HIP** — no such device on this host. Still marked unmeasured.
- **Native NVIDIA CUDA/PTX** — every PTX figure is ZLUDA-on-AMD. Strong evidence about the emitted PTX, weaker about what an NVIDIA driver does with it.
- **The refusal's completeness.** I showed it fires on the shapes I could construct and not on the near-miss I could construct. I did not exhaustively characterise the reachable set — in particular `stmt_binders` does not cover match-pattern binders (documented at `expr_names`), so a pattern-bound name still reads as free. That is the over-approximating direction, so it cannot emit an undeclared identifier, but it means the guard is a backstop and not a decision procedure.

Not merging.

mathiasbourgoin:
@coderabbitai review

This PR's `CHANGES_REQUESTED` verdict predates the current head by several commits, so it is stale rather than outstanding. Requesting a re-read of the current state.

Summary of what changed since the reviewed commit is in the triage comment above. CI is green on the current head, and an independent adversarial review of the current state has been completed.

PR #383, opened 2026-07-30
fix(ci): the alias-coverage guard was one level of reachability short (H6)

mathiasbourgoin:
backlog-186. The review said the HIP backend sits outside CI. It does — but the prior question was **why the whole-tree alias guard, built for exactly this class** (#147, seven invisible suites), reported everything accounted for.

## Diagnosis

`check-test-alias-coverage.sh` decides a target is wired with `re.findall(r"([A-Za-z0-9_]+)\.exe", src)` — the name appearing as `<name>.exe` anywhere in its **own dune file**. The HIP tests appear in `%{dep:…exe}` inside `(rule (alias e2e-hip))`, so they pass.

The guard verifies **target → rule** and never **rule's alias → CI**. Attaching tests to a brand-new alias that nothing invokes satisfies it completely.

Confirmed by measurement rather than inference: the 40 declared-unwired targets are 27 benchmark / 8 browser-target / 4 hardware-probe / 1 tool — **no HIP among them**. The guard never saw them as unwired at all.

## Measured scope

Exactly **two** non-self-executing aliases exist in the tree, and **both** are unreachable:

| alias | rules | executables |
|---|---|---|
| `e2e-hip` | 6 | 6 |
| `e2e-gpu` | 6 | 6 |

**Twelve** targets, not the eighteen the review stated — another figure that didn't survive re-derivation. And no false positives, so the new direction is precise today.

## DIRECTION 3

Every alias attached to a rule must be invoked from a workflow, Makefile or script, or be **declared** with `gpu-required | manual-only`. Self-executing aliases (`runtest`, `default`, `fmt`, …) need no invoker.

Both aliases are **declared rather than invoked**, which is the honest disposition — they need hardware no `ubuntu-latest` runner has, and declaring makes the gap *counted* instead of invisible. Pointing a CI step at them would skip for want of a device and become a **seventh** skip-as-pass.

`ci.yml` now compile-gates the same targets via `@sarek-hip/all` — the *directory* alias, not `@e2e-hip`, so rules are built and not run. That catches the whole invalid-code class for ~1.9k lines of backend without a device. Execution belongs to #178.

## The gate caught three of my own mistakes while I built it

Each is a real shape, so they're recorded rather than quietly fixed:

1. DIRECTION 2 compared **all** declaration keys against the **target** unwired set, so every `alias::` key read STALE — two namespaces sharing one file.
2. My `ci.yml` **comment** explaining why `@e2e-hip` isn't invoked made `@e2e-hip` look invoked. The dune scan already strips `;`-comments for precisely this reason ("names mentioned in prose do not count as wiring"); the invoker scan didn't. It now strips `#`-comments too, erring toward *more* reds — a stripped `#` inside a string can only hide an invocation, never invent one.
3. The **declaration file** lives under `scripts/` and its justifications name the aliases they exempt, so counting it as an invoker made every declared alias look reachable — and then DIRECTION 3 called its own declaration stale. Excluded by path.

Green: 277 targets; 40 unwired, all declared; 3 rule aliases, all reachable or declared. `ci.yml` validates as YAML; `check-kb-properties.sh` green.

## Summary by CodeRabbit

* **New Features / CI Guards**
  * Added a CI guard to validate workflow YAML step structure (exactly one `run` or `uses`, no duplicate keys within a step) with safe fail-close behavior.
  * Expanded CI build checks to compile/link the HIP backend and its tests without requiring HIP runtime execution.
* **Tests**
  * Added dedicated red-path coverage to exercise malformed/edge-case workflow parsing behavior.
  * Enhanced alias coverage validation with new justification tags and stronger alias reachability/coverage enforcement.
* **Documentation**
  * Updated machine-checked KB documentation for the new workflow-steps gate.


mathiasbourgoin:
Fixed, and the finding was sharper than the review stated — thank you for it.

## It was not "can invalidate the workflow"

Parsing the file says what actually happened. YAML keeps the **last** duplicate key, so the workflow parsed fine and:

```
step[21] 'Check every negative case is asserted'  run=None
step[22] 'Compile-gate the HIP backend'           run='./scripts/check-negative-case-coverage.sh'
```

The step **named** for the HIP compile-gate ran `check-negative-case-coverage.sh`, and `dune build @sarek-hip/all` — the entire deliverable of backlog-186, ~1.9k lines of backend plus 6 test executables — never executed once. Meanwhile the CI log showed a green check under the correct name. After the fix:

```
step[21] 'Check every negative case is asserted'  run='./scripts/check-negative-case-coverage.sh'
step[22] 'Compile-gate the HIP backend'           run='dune build @sarek-hip/all'
```

The gate itself was never wrong — `dune build @sarek-hip/all` is exit 0 locally. It was wired to nothing.

## Gated, because this is a class

A step that runs something other than its name is the worst shape of green available, and nothing in this repo read the workflow's structure. `scripts/check-workflow-steps.sh` now requires, per step, exactly one of `run:`/`uses:` and no repeated key. It runs in CI beside the other host-side gates.

Dependency-free by necessity rather than taste: **PyYAML's `safe_load` silently drops duplicate keys**, so parsing with it would have reported this very tree clean — the library cannot see the defect. It is a line scanner, and it must not descend into two places, both forced by measurement (the first version emitted 5 findings, all spurious):

- nested mappings — `with: {name: coverage}` is an artifact name, not a second step name
- block scalars — `deploy-pr-preview.yml` embeds JavaScript under `script: |` whose object literals carry `owner:`, `repo:`, `body:`

### Prove-red — baseline exit 0, 66 steps across 6 files

| mutation | result |
|---|---|
| reintroduce the exact misordered insertion | exit 1, **both** halves named |
| a step with no action at all | exit 1 |
| a duplicated non-action key | exit 1 |
| no workflow files present | exit 2 (fails closed, does not skip) |

The first row is the one that matters: the gate is verified against the specific bug that motivated it, not a synthetic stand-in.

mathiasbourgoin:
## Triage of the four actionable comments

Three real, one already fixed. Each real one is fixed **and** pinned by a new case, so
the fix cannot silently regress. Every exit code below was captured directly, never
`$?` after a pipe.

Also in this push: the CI failure that was blocking the required check. `build
(ubuntu-latest, 5.4.0, true)` was failing at `dune: command not found`, exit 127 — the
HIP compile-gate ran `dune build @sarek-hip/all` as a **host** `run:` in a job that
provisions no host switch. The backend was not at fault (`dune build @sarek-hip/all` is
exit 0 locally; backend and all 6 test executables link). Fixed by running the same
command inside `spoc-ci:latest` like every other toolchain step, plus a new gate for the
class. The required check is green on `afa59d51`.

### 1. `check-workflow-steps.sh:193` — off-by-one in the block-scalar skip boundary 🔴

**CONFIRMED**, and it was as bad as you said. Reproduced before touching anything:

```yaml
steps:
  - run: |
      echo "the first key IS the block scalar"
    uses: actions/checkout@v4
```

```
check-workflow-steps: OK — 1 steps across 1 workflow file(s), each with exactly one action
exit 0
```

A step with two actions, reported clean, by the gate whose entire reason for existing is
that defect. Your diagnosis of the mechanism is exact: the skip test is
`cur_ind > skip_until_indent`, so `child_indent - 1` swallowed lines at
`cur_ind == child_indent`, which is precisely where sibling keys sit. Both branches now
use `child_indent` — they mean the same thing, so the two sites had no business
disagreeing.

Pinned by three cases, the third deliberately in the opposite direction so that
"swallows nothing" cannot pass either:

```
PASS red:   first-key block scalar must not swallow a sibling uses: (exit 1)
PASS red:   first-key block scalar must not swallow a duplicate key (exit 1)
PASS green: first-key block scalar still hides its own deeper text (exit 0)
```

### 2. `check-workflow-steps.sh:72` — scope `STEP_START` to the `steps:` context 🟡

**CONFIRMED.** Measured on a valid workflow:

```yaml
jobs:
  build:
    strategy:
      matrix:
        include:
          - os: ubuntu-latest
            ocaml: 5.4.0
    steps:
      - name: Real step
        run: echo hi
```

```
.github/workflows/ci.yml:7: step '(unnamed, line 7)' has neither `run:` nor `uses:` — it does nothing
exit 1
```

Latent rather than live only because this tree's matrix lists are plain scalars
(`- ubuntu-latest`, no colon, so they never matched). Not a reason to leave it: a gate
that reports a correct workflow dirty gets switched off, and its true findings go with
it.

Dash-items are now recognised only inside a `steps:` list **and** only at that list's own
item indentation. The second half of that also covers the nested case you mentioned — a
multi-object `with:` value whose `- name:` entries were closing the real step early
(losing its remaining keys) *and* adding a phantom actionless step.

Four cases, three of them guarding against over-correction, since "scoped to `steps:`"
and "stopped scanning" would otherwise look identical:

```
PASS green: a matrix include entry is not a step (exit 0)
PASS green: a nested list inside with: is not a step (exit 0)
PASS red:   a duplicate action after a matrix include is still caught (exit 1)
PASS red:   the steps list of a SECOND job is still scanned (exit 1)
```

The harness went 17 -> 24 cases; the real tree still reports 68 steps (67 + the one new
step this PR adds), so the scoping dropped nothing.

### 3. `check-test-alias-coverage.sh:190` — directory-scoped aliases 🟠

**CONFIRMED on the `alias_is_built` half**, and fixed. The pattern required the alias name
immediately after `@`, so `dune build @<dir>/<alias>` and the `@@` single-directory form
both read as UNREACHABLE — the gate would have demanded an exemption for an alias CI
really did build. Optional directory groups are accepted now, and I checked both
polarities rather than just the one that motivated the change:

| alias | text | reachable | why |
|---|---|---|---|
| `e2e-hip` | `dune build @sarek-hip/e2e-hip` | yes | directory-scoped |
| `e2e-hip` | `dune build @@sarek-hip/e2e-hip` | yes | non-recursive form |
| `e2e-hip` | `dune build @a/b/c/e2e-hip` | yes | nested dirs |
| `e2e` | `dune build @e2e-hip` | **no** | prefix collision still refused |
| `e2e` | `dune build @sarek-hip/e2e-hip` | **no** | still refused when prefixed |
| `e2e-hip` | `echo 'we do not build @e2e-hip'` | **no** | prose is not wiring |

The word-boundary lookahead keeps doing the anti-prefix work it was added for; the
directory groups cannot rescue such a match because each requires a `/`.

**On `SELF_EXECUTING`: partly agreed, and narrowed.** `check`, `doc` and `install` are
gone — you are right that nothing builds them unless something names them, so a rule on
`(alias check)` was being skipped as though dune ran it. That is an exemption granted by
the alias's *name* rather than by evidence it gets built, which is this gate's own
vacuous-pass shape. I kept `runtest`, `default`, `fmt` and `all`: `dune build` builds
`@default`, which depends on `@all`; `dune runtest` builds `@runtest`; and `@fmt` is built
by this repo's own formatting gate. Those four are reachable by construction, not by
assumption. No behaviour change on this tree — its only rule aliases are `runtest`,
`e2e-gpu` and `e2e-hip`.

Pinning here is via the prove-red spec rather than a `.test.sh`, since that is this
gate's red path. The fixture's invoker is now itself directory-scoped, so the **green
baseline is the pin** for the new capability — revert the pattern and the baseline goes
red instead of quietly losing it. Observed exactly that:

```
scripts/check-test-alias-coverage.sh: BASELINE is not green. The unmutated subject exited 1, declared 0.
UNREACHABLE ALIAS: @e2e-fixture is attached to rules in wired/dune,
```

Plus a new red mutation for the guard: `[alias-prefix-collision] red: exit 1, says
'e2e-fixture'`. prove-red is now 6 subjects, 30 mutations, exit 0.

### 4. `check-test-alias-coverage.sh:240` — add a reachability invoker to the fixture 🟠

**Already fixed — refuted as filed.** The comment describes the fixture root as holding
only `decl.tsv`, `wired/dune` and `exempt/dune`. It also holds a `Makefile`:

```
scripts/prove-red-fixtures/alias-coverage/Makefile
scripts/prove-red-fixtures/alias-coverage/decl.tsv
scripts/prove-red-fixtures/alias-coverage/exempt/dune
scripts/prove-red-fixtures/alias-coverage/wired/dune
```

added in `60fdcf8d` ("give the alias-coverage fixture an invoker, and cover DIRECTION 3")
at 10:49Z — the comment was filed at 06:33Z, against the earlier tree. It carries the
`dune build @<dir>/e2e-fixture` invoker, and the red-path mutation you asked for already
exists as `alias-unreachable` (it strips that line and expects exit 1). `prove-red.sh`
is exit 0, and its baseline for this subject is green, which it could not be if
`invoker_count` were 0.

### One caution worth recording

Writing a **real** alias name in this gate's own prose *wires* it. `scripts/*` is invoker
text and only `#`-comments are stripped — a Python docstring is not. A draft of my
`alias_is_built` docstring contained a real `@<dir>/<alias>` example, which made that
alias read as reachable and turned its exemption STALE (`check-test-alias-coverage.sh`
exit 1). Same trap the file already records against a ci.yml comment, one layer in. The
example now uses deliberately fake names, and says why.

### Gates

`dune build` 0 · `dune runtest` 0 · `dune build @fmt` 0 · `check-license-headers` 0 ·
`check-kb-properties` 0 · `prove-red` 0 (6 subjects, 30 mutations) ·
`check-test-alias-coverage` 0 · `check-workflow-steps` 0 ·
`check-workflow-toolchain-env` 0 · `check-dune-dir-visibility` 0 ·
`check-negative-case-coverage` 0 · `check-workflow-steps.test` 0 (24 cases) ·
`check-workflow-toolchain-env.test` 0 (16 cases).

Branch is at `0df91573`. Please re-review.

mathiasbourgoin:
@coderabbitai review

The last review was against `e0fb86a0`; the branch is now at `0df91573`, two commits on.
CI is green on it (all four jobs, including the `build (ubuntu-latest, 5.4.0, true)` check
that was failing at `dune: command not found`). Triage of the four findings from that
review is in the comment above.

mathiasbourgoin:
@coderabbitai review

This PR's `CHANGES_REQUESTED` verdict predates the current head by several commits, so it is stale rather than outstanding. Requesting a re-read of the current state.

Summary of what changed since the reviewed commit is in the triage comment above. CI is green on the current head, and an independent adversarial review of the current state has been completed.

PR #140, opened 2026-06-02
refactor: split monolithic modules + dedup backend codegen (audit 2026-06-02)

mathiasbourgoin:
Code-quality refactor driven by `briefs/audit-2026-06-02.md`. **Pure structural work** — every change was reviewed token-level byte-identical; no behavior changed except one intentional, reviewed Metal variant-codegen fix.

## Commits (10)

| Area | What | Size |
|---|---|---|
| backends | Extract shared `Sarek_ir_codegen` (variant codegen + `mangle_name`) | 4 backends −297/+34; fixes latent Metal variant-typedef bug |
| ppx | Split `Sarek_native_gen.ml` into 4 modules | 1951 → 465 |
| framework | Extract shared `Framework_sig.PLUGIN_BASE` module type | 3 plugin bases −131 each |
| scripts | Repair broken grep regex in `add-license-headers.sh` + pin header identity to canonical maintainer | gate no longer crashes; gmail-only |
| runtime | Split `Sarek_cpu_runtime.ml` into 4 modules | 1497 → 358 |
| runtime | Split `Sarek_ir_interp.ml` into 4 modules | 1573 → 535 |
| ppx | Split `Sarek_parse.ml` (helpers) + decompose `parse_expression` | 839 → 613; 316-line fn → 115 |
| vulkan | Split `Vulkan_api.ml` into per-submodule files | 1756 → 49 |

## Verification

- **Per-intake review**: each split confirmed token-level byte-identical (function bodies, error strings, recursive calls, variable capture). Documented bugs were relocated verbatim, **not** silently fixed.
- **`@sarek/tests/runtest`**: green after every commit.
- **Full e2e matrix on real hardware** (AMD RX 7900 XTX, RADV): 17 e2e tests × 4 backends (OpenCL / Vulkan / Native / Interpreter) = **68/68 pass**, including variant/record GLSL codegen on Vulkan.

## Notes

- Three modules remain >500 lines (`Sarek_native_gen_kernel` 842, `Sarek_cpu_runtime_pools` 813, `Vulkan_api_{memory,kernel}` ~560) — cohesive units; further splitting deferred.
- CUDA/Metal backends not built in this environment (no NVRTC / not macOS); validated by compilation only where applicable.
- Base is `main` (not `master`, which is 656 commits behind).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Expanded Vulkan backend with full compute kernel execution, device management, memory transfers, and event synchronization support.

[...]

PR #384, opened 2026-07-30
test(ci): prove-red spec for assert-toolchain.sh — the gate holding up the codegen story

mathiasbourgoin:
backlog-176. `kb/properties.md` nominated this as *"the highest-value of the uncovered gates … Best candidate for the first application of the approved scripts/prove-red.sh"*, and recorded that neither its hard-fail nor its positive control had ever been observed firing. **Both are now observed**, along with a third thing nobody was looking for: the gate crashes in one reachable configuration, and the crash reads as its verdict.

## Why it took two audits — the reason is structural

`ci/assert-toolchain.sh` is green **only inside `ci/Dockerfile`'s image**. On a developer host it exits 1 with 3 of 11 checks failing (ptxas, nvdisasm, `cuda_fp16.h` absent), and `prove-red.sh` refuses a subject whose unmutated baseline is not green — *"every red below it would prove nothing."* No spec was possible until green was reachable off-image. Not neglect: a precondition nobody had stated.

`scripts/prove-red-fixtures/assert-toolchain/` is that precondition, and it is **hermetic**: the wrapper rebuilds `PATH` from nothing — stub `ptxas`, `nvdisasm`, `glslangValidator`, `naga`, `clang` and `cc`, a `cuda_fp16.h` marker under `CUDA_PATH`, and symlinks to a fixed allowlist of base utilities (`bash cat chmod find grep head mktemp rm`). No tool under assertion is on that allowlist, so the fixture is their only source.

**Stubbing tools that are usually present is load-bearing, not gratuitous**, and the first revision of this PR got it wrong on the stated theory that "faking a tool that is really there weakens the baseline for nothing". Two measured reasons:

1. `prove-red.sh` runs in the **`build` job on a bare `ubuntu-latest` runner**, not in this script's image. That runner has no naga, no glslangValidator and no loadable libnvrtc, so an unstubbed baseline is red on arrival exactly where the spec has to execute — measured on a runner-like `PATH`: `naga is not on PATH`, `1 of 12 checks failed`.
2. A preserved host `PATH` **defeats the mutations**. With a real `ptxas` elsewhere on `PATH`, `rm -f fixture/bin/ptxas` was observed exiting 0 at 13/13 — the deletion invisible, the mutation unable to go red. `prove-red.sh` reads that as `DID NOT FAIL` and exits 1, so the symptom is a spurious red for the whole repository on a CUDA host rather than a credited red. Either way the verdict becomes a property of the machine, which is what disqualifies it as evidence. (CodeRabbit's Major, PR #384.)

The one host path `PATH` cannot hide is the literal `/usr/local/cuda` in the header search. The wrapper refuses with exit 2 when a real `cuda_fp16.h` is there rather than let that mutation go inert.

## The baseline reaches 13/13, and that is the point

The `EXPECTED_CHECKS` drift guard only runs when **zero** checks fail. Off-image the `failures > 0` branch always exited first, so that guard had **never executed** — the guard whose own comment reads *"'N of N' is tautological by construction and can never report a gap on its own."* `check-vanished` is its first observed firing.

## Five mutations, all red

| mutation | what it proves |
|---|---|
| `ptxas-absent` | the historical disaster verbatim — the old image carried none of these tools and CI was green having validated nothing |
| `nvdisasm-present-but-broken` | exits 0, stops naming the entry point. The only mutation testing this script's claim that a present-but-broken binary *"skips just as silently as an absent one"* for a binary functional in every respect the gate can see, its own exit code included |
| `cuda-header-tree-missing` | the f16 regression was an unresolvable `#include`, so tool presence is not enough |
| `clang-fp64-unsupported` | #140 verbatim — clang on `PATH`, fp32 fine, `double` rejected. The **fp64 positive control's first observed firing**; unreachable with a real clang on a Linux runner, one variable with a stubbed one |
| `check-vanished` | the drift guard, first firing |

## A live crash in the gate, found by the review fan-out

The nvdisasm chain probe was one `&&` chain starting with `ptxas`, with `out=$(nvdisasm ...)` as its second element. ptxas absent + nvdisasm present short-circuits before that assignment, and the failure branch expands an unset `out` under `set -u`:

```
ci/assert-toolchain.sh: line 167: out: unbound variable
```

Exit 1, nine of the twelve reachable checks skipped, no summary. `ptxas-absent` originally asserted exit 1 plus `ptxas is not on PATH` — printed 70 lines earlier — so **both halves of the assertion were satisfied by a crash**. `cuda-nvcc` and `cuda-nvdisasm` are separate packages and this script's own comment records nvdisasm absent while ptxas was present; the mirror is just as reachable. This fixture is the first thing in the repository to execute that line.

The ptxas step is now a separate branch naming the real condition. The naive `out=""` fix was tried and rejected: it completes, but reports `nvdisasm is on PATH but cannot disassemble` when nvdisasm is fine — a crash traded for a wrong accusation. `ptxas-absent` now declares the **summary** line, `toolchain assertion FAILED: 2 of 12 checks failed`, which no crash can print.

## `fail()` accounting: three of four, not four

There is no single-variable mutation that proves `fail()` in isolation — from a green baseline there is no failure for a miscounting `fail()` to lose. It is pinned transitively, and measured by deleting `failures=$((failures + 1))`: `nvdisasm-present-but-broken`, `cuda-header-tree-missing` and `clang-fp64-unsupported` all go green. `ptxas-absent` does **not** — it still exits 1, via the `EXPECTED_CHECKS` drift guard, because removing ptxas collapses two counted checks into one and leaves `checks=12` with zero recorded failures. A second independent pin on that mutation, but not `fail()` accounting, so it is named as the exception rather than counted in.

## Evidence

`scripts/prove-red.sh`: **7 subjects, 38 mutations**, exit 0 — locally and on the CI runner, where the hermetic baseline reaches `13/13` and all five mutations go red with their declared messages. Every stub proven on the executed path: each `--version` banner appears verbatim in the baseline, and removing each of the six in turn drives the gate to exit 1. The mechanism proven able to refuse on this subject too: a broken mutation anchor gives `AssertionError: anchor count 0`, and an `rm -f` retargeted at a nonexistent path gives `changed NOTHING in the scratch tree` — both exit 2.

Also fixed here: the missing SPDX header on `run-with-stubs.sh`, which was the actual `build` failure on the first revision (`check-license-headers.sh` exit 1; the visible log tail was artifact-upload noise), and ci.yml's already-stale `~0.5s for 4 subjects and 13 mutations`.

`EXPECTED_SUBJECTS` is **7**: `main` was already at 6 after #393 and #381, and this adds the seventh subject.

**Known gap, not closed here:** nothing ties a `prove-red-spec` block to its KB row. `scripts/check-kb-properties.sh` has no reference to prove-red, so had this PR left `red_path: null` beside a live block, no gate would have noticed. Worth its own change — the natural place for the check is inside `prove-red.sh`, whose scratch tree does not copy `kb/properties.md`.


mathiasbourgoin:
Revived, rebased onto current `main`, and the review finding is addressed. Four commits now; the branch was 36 hours and 17 merges stale.

## The `build` failure was not what the log tail showed

Run 30508293148. The visible tail is artifact-upload noise (`No files were found with the provided path: _coverage/unit-report`); the real failing step is 900 lines earlier:

```
UPDATED: scripts/prove-red-fixtures/assert-toolchain/run-with-stubs.sh
Files needing updates: 1
##[error]Process completed with exit code 1.
```

`scripts/check-license-headers.sh`, exit 1 — the new wrapper shipped without an SPDX header. The stubs under `bin/` were never flagged because that checker selects by extension and they have none; they carry headers now anyway. Reproduced locally before and after: exit 1 → exit 0.

## CodeRabbit's Major on `run-with-stubs.sh:25` — fixed, and it was worse than stated

> `export PATH="$here/bin:$PATH"` preserves host executables, so deleting the fixture `ptxas` can still satisfy the present-but-absent mutation via a host binary.

Correct, and now measured rather than argued. Same fixture tree, ptxas deleted, one working `ptxas` elsewhere on `PATH`:

| wrapper | exit | last line |
|---|---|---|
| pre-fix (`$here/bin:$PATH`) | **0** | `toolchain assertion OK: 13/13 checks passed.` |
| hermetic | **1** | `::error::ptxas is not on PATH.` |

The second half of the Major — "otherwise the baseline is host-dependent" — turned out to be the blocking one. `prove-red.sh` runs in the `build` job on a bare `ubuntu-latest` runner, **not** inside `ci/Dockerfile`'s image. That runner has no naga, no glslangValidator and no loadable libnvrtc. Measured on a runner-like `PATH`, the pre-fix baseline is red on arrival:

```
::error::naga is not on PATH.
::error::toolchain assertion FAILED: 1 of 12 checks failed.
```

`prove-red.sh` refuses a subject red on arrival, so this PR would have gone from a license-header failure straight to `exit 2` for the whole repository. The "clang, glslangValidator and naga are genuinely present on a normal developer host" line described one machine, not the one that has to agree.

`PATH` is now rebuilt from nothing: the fixture's `bin/` first, then a private directory of symlinks to a fixed allowlist (`bash cat chmod find grep head mktemp rm`). No tool under assertion is on that list. New stubs for `glslangValidator`, `naga`, `clang` and `cc` — `cc` has to emit a *runnable* probe, because the NVRTC check compiles one and executes it.

CodeRabbit's `CUDA_PATH` point stands and cannot be fixed by `PATH`: the header search names a literal `/usr/local/cuda`. The wrapper refuses with exit 2 when a real `cuda_fp16.h` is there rather than let that mutation go inert. Both polarities exercised (refuses with the header, runs to 13/13 without it).

## Two blocking defects the review fan-out found, one of them live in the gate

**The subject crashed, and the crash read as the verdict.** The nvdisasm chain probe was a single `&&` chain starting with `ptxas`, with `out=$(nvdisasm ...)` as its second element. ptxas absent + nvdisasm present short-circuits before that assignment, and the failure branch expands an unset `out` under `set -u`:

```
ci/assert-toolchain.sh: line 167: out: unbound variable
```

Exit 1, nine of the twelve reachable checks skipped, no summary. `ptxas-absent` asserted exit 1 plus `ptxas is not on PATH` — printed 70 lines earlier — so both halves matched a crash. `cuda-nvcc` and `cuda-nvdisasm` are separate packages and this script's own comment records nvdisasm absent while ptxas was present; the mirror is just as reachable. This fixture is the first thing in the repository to execute that line.

The naive `out=""` fix was tried and rejected: it completes, but reports `nvdisasm is on PATH but cannot disassemble` when nvdisasm is fine and its own `--version` passed two lines up — a crash traded for a wrong accusation. The ptxas step is now a separate branch naming the real condition, and `ptxas-absent` declares the **summary** line, `toolchain assertion FAILED: 2 of 12 checks failed`, which no crash can print.

**The `fail()` transitivity claim was one mutation too wide.** Measured by deleting `failures=$((failures + 1))`: nvdisasm-present-but-broken, cuda-header-tree-missing and clang-fp64-unsupported all go green as claimed — `ptxas-absent` still exits 1, via the `EXPECTED_CHECKS` drift guard, because removing ptxas leaves `checks=12` with zero recorded failures. Three of four, exception named, in both the spec block and the KB row. Not weakened further: the claim is correct for the other three.

Also from the fan-out: stub contract comments said "exactly … and nothing else" while every stub matches loosely; "hermetic" now says PATH-resolved and names what is not isolated; `chmod` was attributed to the subject and belongs to `bin/cc`; "credited a red it did not earn" was the wrong failure mode for the host-PATH case in three places (prove-red reads exit 0 as `DID NOT FAIL`, so the symptom is a spurious red on a CUDA host); the CUDA refusal's blast radius went unstated; and ci.yml's `~0.5s for 4 subjects and 13 mutations` was already stale.

A second adversarial pass over those corrections — because a narrowing correction overshooting is how this class recurs — returned GO with five more prose findings, all folded in: the CUDA refusal's blast radius was overstated (prove-red scans `scripts/` before `ci/`, so six subjects have already reported; what is lost is this subject's 5 mutations and the run's verdict, measured by forcing the refusal); the "what is not isolated" list was three commands short of the exhaustiveness it claimed; "the gate stays green at 13/13 on any CUDA host" generalised one machine — the *invisibility* of the deleted stub is general, the 13/13 also needed this host's naga/glslangValidator/libnvrtc; "the only mutation that tests that claim" needed qualifying, since `clang-fp64-unsupported` is also present-but-broken; and three stubs said "readable file" where `[ -f ]` says exists-and-regular.

## Fifth mutation

`clang-fp64-unsupported` — the stub keeps compiling fp32 OpenCL and starts rejecting `double`. That is #140 verbatim, and it is the first observed firing of the fp64 positive control this gate's KB row already claimed to carry. It needs a stubbed clang: a real clang on a Linux runner cannot be made to lack fp64 by one variable.

## Both polarities, per assertion

Baseline green at 13/13 with every stub visibly executed, observed rather than inferred: each `--version` line in the baseline output is the stub's own banner verbatim (`ptxas: STUB for prove-red…`, `naga 0.0.0-STUB…`), and two probes go further than an exit code, requiring the stub to have produced a real artifact: the cubin `ptxas` writes and `nvdisasm` is then handed, and the executable `cc` writes and the gate then runs. The glslangValidator and naga probes read only the exit code — the stubs write a `.spv` anyway, but nothing yet checks that, and it should not be claimed as coverage. Each of the six stubs removed in turn drives the gate to exit 1 with the message belonging to the check that uses it — including `cc`, whose removal surfaces as `could not build the NVRTC dlopen probe` — so none is decorative. Each declared mutation red with its declared exit code and message.

The mechanism was also proven able to refuse: breaking the clang mutation's anchor gives `AssertionError: anchor count 0`, and retargeting `rm -f` at a path that does not exist gives ``mutation `ptxas-absent`: `apply:` ran successfully (exit 0) and changed NOTHING in the scratch tree`` — both exit 2, both on this subject.

## Rebase

`origin/main` moved under this branch, including #393's rewrite of `prove-red.sh`. No conflict: the branch's only `prove-red.sh` hunk was the `EXPECTED_SUBJECTS` bump, which main had already made. It is now 7 — main is at 6 and this adds the seventh subject. Verified both parents survived by diffing the rebased range against the pre-rebase range restricted to branch-owned files (identical but for blob hashes and one moved hunk header) and confirming main's 526-file delta is present. No conflict markers anywhere.

Nothing in the branch conflicts with #393's new rules: no mutation is a no-op (proven above), and no `invoke:`/`argv:` builds a dune alias — the subject is a shell wrapper.

## Gates

Project switch `/home/mathias/dev/SPOC/_opam` — dune 3.24.1, OCaml 5.4.0, ppxlib 0.38.0. Exit codes captured directly, never a prose PASS.

| gate | exit |
|---|---|
| `dune build` | 0 |
| `dune build @fmt` | 0 |
| `make test_negative` | 0 |
| `scripts/prove-red.sh` | 0 — 7 subjects, 38 mutations |
| `scripts/prove-red.test.sh` | 0 |
| `scripts/check-kb-properties.sh` | 0 — 42 declarations |
| `scripts/check-opam-clean.sh` | **1** — pre-existing |

`check-opam-clean` disagrees only about `"dune" {>= "3.15"}` → `{>= "3.15" & >= "3.15"}` across all seven `.opam` files: dune 3.24.1 emits the doubled constraint, the tracked copies were generated by 3.15. Unrelated to this branch, which touches no build metadata. It could not be cross-checked against ambient dune 3.23.0 — the ambient switch carries ppxlib 0.35 and cannot build this tree at all (`no constructor "Ptyp_open"`, exit 2), so that comparison is unavailable here rather than green.

## What I could not verify

- **That the `build` job is now green.** Everything above is local. The ubuntu-latest baseline claim is inference from the workflow (no apt install of glslang-tools or naga before the prove-red step) plus a local runner-like `PATH` simulation, not a measurement on a real runner. This PR's own CI run is the measurement.
- **The `/usr/local/cuda` refusal on a machine that actually has one.** Exercised by pointing the same predicate at a directory I could create; the literal path needs root here.
- **A hole this PR does not close:** nothing ties a `prove-red-spec` block to its KB row. `scripts/check-kb-properties.sh` contains no reference to prove-red, so had this PR left `red_path: null` beside a live block, no gate would have noticed — the KB row and the mechanism agree only because a human made them agree. Filling it here does not fix that; worth its own change, since the natural place for the check is inside `prove-red.sh`, whose scratch tree does not copy `kb/properties.md`.

Not merging, not dismissing the review, and deliberately not requesting a re-review.

mathiasbourgoin:
CI settled. The one item I listed as unverified is now measured, on the real runner rather than by inference:

`build (ubuntu-latest, 5.4.0, true)` — **pass**, 8m40s ([run 30634124268](https://github.com/mathiasbourgoin/Sarek/actions/runs/30634124268/job/91167311877)). The `Prove-red` step:

```
== ci/assert-toolchain.sh
   baseline: exit 0, says 'toolchain assertion OK: 13/13 checks passed'
   [ptxas-absent] red: exit 1, says 'toolchain assertion FAILED: 2 of 12 checks failed'
   ...
   [check-vanished] red: exit 1, says 'checks, expected'

OK: 7 subject(s), 38 mutation(s); every declared mutation produced its declared exit code and message, and every baseline was green.
```

So the hermetic baseline *is* reachable on a bare `ubuntu-latest` runner — which is the whole reason the fixture had to stop deferring to the host. `License header guard` also passes (`✓ All license headers are up-to-date!`), closing the original failure.

`check-generated-code`, `formal-proofs` and `pocl-runner-probe` all pass. CodeRabbit reports rate-limited and I have deliberately not requested a re-review.

I have also rewritten the PR description: it still described four mutations, `6 subjects, 29 mutations`, a `5 → 6` merge-order warning that both #393 and #381 have since overtaken, and the "clang, glslangValidator and naga are deliberately not stubbed" reasoning that this branch now refutes with measurements. Leaving a description that states the opposite of the code is the same defect class the change is about.

Issue #93, opened 2026-01-11
Sprint 2 Benchmarks - Complete Infrastructure Integration


mathiasbourgoin:
# Sprint 2 Benchmarks - Complete Infrastructure Integration

## Overview

This issue tracks the complete integration of Sprint 2 (P1) parallel pattern benchmarks and migration of all benchmarks to use the shared `benchmark_runner` infrastructure. 

**Current Status:**
- ✅ 5 Sprint 2 benchmarks implemented and compiling
- ✅ Shared `benchmark_runner.ml` infrastructure created
- ✅ One benchmark (`bench_scan`) migrated to new infrastructure
- ⏳ 4 Sprint 2 benchmarks need CLI migration
- ⏳ All benchmarks need full integration (JSON, descriptions, etc.)
- ⏳ Sprint 1 benchmarks still have duplicate parse_args code

**Related Branch:** `benchmark-sprint-2` (3 commits ready)

---

### What We Have Now

**Sprint 2 Benchmarks (5 new patterns):**
- `benchmarks/bench_scan.ml` (250 lines) - Hillis-Steele inclusive scan ✅ **Migrated**
- `benchmarks/bench_bitonic_sort.ml` (259 lines) - Data-oblivious sorting
- `benchmarks/bench_histogram.ml` (246 lines) - 256-bin histogram with atomics
- `benchmarks/bench_gather_scatter.ml` (369 lines) - Indirect memory access
- `benchmarks/bench_radix_sort.ml` (390 lines) - 4-bit radix sort (⚠️ verification failing)

**Sprint 1 Benchmarks (11 existing):**
- `bench_vector_add.ml` - No parse_args (uses Benchmark_runner already?)
- `bench_vector_copy.ml` - Has parse_args (needs migration)
- `bench_matrix_mul.ml` - No parse_args
- `bench_matrix_mul_tiled.ml` - Has parse_args (needs migration)
- `bench_transpose.ml` - No parse_args
- `bench_transpose_tiled.ml` - No parse_args
- `bench_reduction.ml` - No parse_args
- `bench_reduction_max.ml` - Has parse_args (needs migration)
- `bench_dot_product.ml` - Has parse_args (needs migration)
- `bench_stream_triad.ml` - Has parse_args (needs migration)
- `bench_mandelbrot.ml` - No parse_args

**Shared Infrastructure:**
- `benchmarks/benchmark_runner.ml` (207 lines) - NEW
[...]

PR #386, opened 2026-07-30
feat(formal): model the setp MNEMONIC, so a wrong one is a proof failure

mathiasbourgoin:
backlog-187. **Targets `fix/ptx-float-ne-nan` (#377), not `main`** — the mnemonic table describes what that branch's emitter fix does, so it should land after it.

The audit is the finding here; the code is the consequence.

## What the audit established

**1. The model's spec was already correct.** `agpu_eval_ptx_cmp` has, for both float widths, `PNe => negb (eqb a b)` — which on NaN is `negb false = true`, the **unordered** semantics the emitter got wrong in #173.

**2. Both `*_agrees` lemmas are definitional identities** — and there are only two in the whole development. Both proofs end `exact Heval`: the goal *is* the hypothesis after `simpl`. `ir_binop_to_ptx_cmp` is a trivial 1:1 tag map, and `agpu_eval_ptx_cmp` re-implements the same comparisons over the same float ops. So the lemma proves *trivial-map-then-mirrored-evaluator = direct-evaluator*. It cannot fail and constrains nothing.

**3. The crux.** `formal/codegen-ptx/test/test_codegen_ptx_conformance.ml` declares its **own** `ptx_cmp_tag` (:304), its **own** `agpu_eval_ptx_cmp` (:367), its **own** `ir_binop_to_ptx_cmp` (:496) — and never references `Sarek_ir_ptx_expr`. The "conformance mirror" mirrors the **model**, not the emitter. And `setp` appears in `formal/` **only in comments**.

So the chain is model ↔ OCaml hand-mirror of the same model, and **neither** ↔ the emitter. The development proves the model self-consistent in two languages and has never compared anything against a Sarek-emitted instruction.

That isn't a gap in the proofs. It's the proofs being about a *different artefact than the one that ships* — which is precisely why 79 theorems didn't see `setp.ne` vs `setp.neu`.

## What this adds

`ptx_cmp_mnemonic : ptx_cmp_tag -> ptx_type -> string` — modelling what a comparison is **spelled**, not only what it **means** — plus four lemmas:

| lemma | pins |
|---|---|
| `ptx_cmp_ne_unordered_on_floats` | float `<>` is `neu` — the property that shipped violated |
| `ptx_cmp_ne_ordered_on_non_floats` | integers have no NaN, so `ne` |
| `ptx_cmp_eq_ordered_everywhere` | `PEq` stays ordered on floats too — the deliberate asymmetry, pinned so it can't be tidied away (C's `==` is *also* false for NaN, so `equ` would wrongly make `nan = nan` true) |
| `ptx_cmp_float_eq_ne_differ` | the one-line statement of the defect: an emitter using one root for both would satisfy every other theorem here |

## Proven red *in the proof assistant*

Writing the #173 defect into the model (`PNe => "ne"` unconditionally) makes Rocq reject the file:

```
Error: Unable to unify ""neu"" with ""ne"".
```

exit 1; restored exit 0. Rocq 9.1.1, machine-checked locally. This is the first artefact in `formal/` that would reject that defect.

## Not claimed here

**The artefact edge.** This makes the model *say* what the mnemonic must be. Asserting the **real emitter** against it — a systematic 6 tags × {int, f32, f64} table — is the remaining half, and it belongs in `test_ptx_snapshot` where #377 already pins float `Ne` in both polarities. Recorded in the task rather than half-built here.

## For your attention

**The 79-theorem figure needs a load-bearing split before it's quoted** — theorems that constrain emitted text versus theorems that relate the model to itself. #175 has just corrected `STATUS.md` to quote 79, and on the evidence above the cmp/binop pair is the second kind. This feeds #93, whose baseline was zero recorded catches.


mathiasbourgoin:
The artefact-edge half of backlog-187 has landed on this branch (`c455251d`).

The model half alone could not have closed the item: a `ptx_cmp_mnemonic` that nothing compares against the emitter is the same shape as the gap it was written to fix. `test_ptx_snapshot`'s new case now asserts all 24 combinations — 6 comparison tags × 4 register widths — against what `Sarek_ir_ptx.generate` actually writes, with the opcode root transcribed from `CodegenPtx.PtxTypes.ptx_cmp_mnemonic` rather than hand-written a second time.

One correction worth recording, because it was measured rather than reasoned. My first version emitted all 24 comparisons from a single kernel and asserted the 24 markers were present. Swapping the `Lt` and `Gt` opcode roots in the emitter left that case **green** — a permutation of the mnemonics satisfies a presence-only assertion over the full set. The case now generates one kernel per comparison, so the absence of the five sibling roots is what carries the tag↔mnemonic binding.

Prove-red against this case alone, baseline exit 0:

| mutation | exit |
|---|---|
| `Lt`/`Gt` roots swapped | 1 (green before the restructure) |
| `Le` → `lt` | 1 |
| float `Ne` unordered → ordered (the backlog-173 defect) | 1 |
| 64-bit int suffix `s64` → `u32` | 1 |
| 32-bit `eq`/`ne` suffix `u32` → `s32` | 1 |

A sixth mutation — deleting the `is_float` test outright — is not listed because it fails the build on an unused variable rather than reaching the assertion, so it is not evidence.

mathiasbourgoin:
Adversarial review: **NO-GO**. The artefact half is sound and measured. The formal half claims more than it proves, and two hard blockers mean CI fails as submitted.

## Blocker 1 — proof-ledger drift; this repo's own gate fails

`bash scripts/check-formal-proofs.sh` → **exit 1**:

```
::error::formal/codegen-ptx/proof-ledger.json does not match this build.
      counts.definitions: committed 60, actual 62
      counts.theorems: committed 79, actual 83
1 ledger/axiom check(s) FAILED.
```

Independently derived with `scripts/gen-proof-ledger.py`: the four new lemmas and two new definitions land in `CodegenPtx.PtxTypes` (16→20 theorems, 5→7 definitions), and `proof-ledger.json` is touched by neither commit. This is precisely the gate backlog-95 established for this class, and the `formal-proofs` CI job fails on it.

## Blocker 2 — does not merge into current `origin/main`

`git merge-tree --write-tree origin/main HEAD` → exit 1, content conflict in `sarek/tests/unit/test_ptx_snapshot.ml` (main's `default_kernel` refactor of `make_vector_add_kernel`/`base_kernel` vs this branch's stale copy). The branch was never rebased after #377 merged. Mechanical, but real. Confirmed the emitter fix itself is *not* a conflict — `e856857c` and main's `77cff792` are content-identical.

## The formal half pins 2 of 6 tags, and the comment says otherwise

Mutating `ptx_cmp_mnemonic` and rebuilding:

| mutation | `make theories/PtxTypes.vo` |
|---|---|
| `PNe => "ne"` always | exit 2 — `Unable to unify ""neu"" with ""ne""` |
| `PNe => "neu"` always | exit 2 — `Unable to unify ""ne"" with ""neu""` |
| `PEq => "equ"` | exit 2 |
| **`PLt`/`PGt` roots swapped** | **exit 0 — GREEN** |

Both `PNe` polarities are genuinely pinned. But no lemma constrains `PLt`/`PLe`/`PGt`/`PGe`, so swapping `lt`/`gt` in the Rocq definition passes Rocq, passes the OCaml table (an independent copy), and passes the artefact test. Meanwhile `test_ptx_snapshot.ml:1211` calls the whole mapping "machine-checked".

That is a claim wider than its code, in the PR whose entire thesis is that a model which *looked* right could not catch a defect. Either add four ordering lemmas or narrow the word.

## `root` is an unenforced hand transcription — and the repo already solved this

`test_ptx_snapshot.ml` never depends on `CodegenPtx`; `ptx_cmp_mnemonic` appears there only in comments. The two tables agree today, symbol by symbol, and **nothing would catch divergence**: the committed `formal/codegen-ptx/PtxTypes.ml` is stale (no `ptx_cmp_mnemonic`, though a rebuild produces one), and `check-formal-proofs.sh` deliberately excludes `Extract.v`'s five root outputs from its canonicalise-and-diff, checking only `sarek_ptx_layout_model`. Unbound in both directions.

`docs/formal/rocq-value-ledger.md` N-01 (backlog-46) removed the layout hand transcription for exactly this reason — "the transcription can no longer drift, because there is no transcription" — and the route exists: `extraction/LayoutExtract.v` + `canonicalize-extraction.py` + the diff gate. A `MnemonicExtract.v` would be the equivalent. The PR's comment admits `root` is unenforced but cites neither the precedent nor the route.

## The "suffix is not semantics" justification is false for four tags

The commit and comment say the width suffix is "register naming, not semantics", so deliberately unmodelled. But `setp.lt.s32` and `setp.lt.u32` are *different comparisons*, and the emitter's own comment forty lines away says so: "Sarek ints are signed, so the integer forms are s32/s64 (sign matters for Lt/Le/Gt/Ge; for Eq/Ne it is irrelevant but harmless)."

Rocq's `ptx_type` has no signed variants at all (`PTX_U32|PTX_U64|PTX_F32|PTX_F64|PTX_Pred`), so this is a model **limitation**, not a scope choice. Net: the genuinely-semantic signedness half is unmodelled, and the modelled half pins 2 of 6 tags.

Minor, same family: the comment on `suffix` says "it keys off the ROOT and not the tag", but the code is `match op with Eq | Ne -> "u32" | _ -> "s32"` — it keys off the **tag**. The emitter keys off the root. Unobservable today, but the mirror is written differently from the rule it claims to mirror.

## Bookkeeping claims that do not hold

- The commit says "the 79-theorem figure, which backlog-175 has just corrected STATUS.md to quote". `grep 79 STATUS.md` finds nothing on this branch; STATUS.md is untouched and still says "Theories: 5 files… Theorems proved: 3".
- The commit says the figure "needs a load-bearing split before being quoted anywhere", then adds no split.
- `rocq-value-ledger.md` M-04 still reads "It is not yet a table over the whole family; that remains open". This PR closes that and does not update the row.
- The commit says it "feeds backlog-93, whose baseline was zero recorded catches" while recording nothing. A Rocq mutation that fails the build is arguably the first recordable catch, and it is not logged.

## What holds up — the artefact half is genuinely sound

Baseline `codegen 18` exit 0; five emitter mutations, each rebuilt, each asserted non-no-op first, run against that case alone:

| emitter mutation | exit |
|---|---|
| `Lt`↔`Gt` roots swapped | 1 |
| `Le`↔`Ge` roots swapped | 1 |
| cyclic permute all four orderings | 1 |
| `Ne` float unordered → always `"ne"` | 1 |
| 64-bit suffix `s64` → `u64` | 1 |

The one-comparison-per-kernel restructuring does close the set-of-markers hole: the `assert_absent` over five sibling markers carries the binding, and all 24 markers are pairwise distinct at every width, so no substring masking. Measured, not reasoned.

Also PASS: `rocq check -R theories CodegenPtx theories/*.vo` clean, "Modules were successfully checked"; no `Admitted`/`admit`/new `Axiom`; axiom allowlist "6 sanctioned, 6 found, exact match" (the pre-existing sin/cos/fma `Parameter`s, untouched); all four lemmas end in `Qed`. Other gates: `dune build` 0, `runtest` 0, `@fmt` 0, license 0, KB 0, prove-red 0, Alcotest registration 0.

## To flip to GO

Regenerate `proof-ledger.json`; rebase onto `origin/main`; either add ordering lemmas or stop calling the mapping "machine-checked"; restate the suffix limitation honestly; and fix the four bookkeeping claims — or drop them.

Issue #46, opened 2019-01-06
adding spoc to jbuild leads to failure when linking


tari3x:
Just adding "spoc" to the libraries section of the jbuild, leads to a compile error:

```
/home/avatar/.opam/default/.opam-switch/build/lacaml.11.0.2/_build/default/src/utils_c.c:40: multiple definition of `copy_two_doubles'; /home/avatar/.opam/default/lib/spoc/libspoc.a(Spoc_c.o):/home/avatar/.opam/default/.opam-switch/build/spoc.20170724/Spoc/Spoc_c.c:482: first defined here
```

  I'm using spoc-20170724, installed from opam.

PR #387, opened 2026-07-30
docs: close backlog-188's five leftovers, and gate the two that can regress

mathiasbourgoin:
Closes the five items filed as backlog-188. Two of them turned out to be instances of classes rather than one-offs, so each ships with a gate.

## What was wrong

| # | Finding | Why nothing caught it |
|---|---|---|
| 1 | `README.md:5` CI badge pointed at `mathiasbourgoin/SPOC` while every other link pointed at Sarek | A badge is an image. It renders, so it looks fine — while reporting a different project's build status from the most-read line in the repo. |
| 2 | Ten `.ml`/`.mli` files generated by `Extract.v` were tracked in git | They are regenerated by every proof run, do not compile as they stand, are named by no dune stanza, and are diffed by no gate. |
| 3 | Four `.ml` comments cited design notes under `roster/` | A path in a comment is read by nobody. |
| 4 | `PtxLayout.v:47` said conformance is established by a test that "hand-mirrors the definitions below" | True before task #46, false after — it described the 150-line transcription that #46 deleted. |
| 5 | `test_ptx_intrinsic_sweep.ml` said "80 names" and "114 kernels" | Both are derived from the emitter's registry, and neither number had a reader. |

## Two findings were classes

**The citations.** `scripts/check-cited-paths-exist.sh` resolves every repo-relative path cited in a tracked `.ml`/`.mli` against `git ls-files`. Running it found **six more** of the same shape under `briefs/`, in files nobody had connected to the four.

Resolution is against what git *tracks*, not what is present on disk. That is the whole point: `roster/` and `briefs/` exist on the machine that wrote these comments and in no clone, so a filesystem check would have passed locally **and** in CI while every reader hit a dead end.

**The counts.** Derived now: 83 names, 120 kernels — not 80 and 114. Rather than write the new numbers, which re-arms the same drift, both are gone from the prose and `sweep` prints them:

```
intrinsic sweep: 120/120 kernels generated over 83 registered intrinsic names, 0 assembled (ptxas absent)
```

## Three things I got wrong building the citation gate

Each was caught by running it, not by reading it:

- A pre-filter requiring the first path component to name a tracked directory made the gate **blind to a citation whose whole directory is missing** — the `roster/` case, i.e. exactly what it is for. Removed.
- Filesystem resolution plus a `tracked_dir` check whose `.` was a regex wildcard produced two false positives on paths that do exist.
- A blanket rewrite also hit four **valid** citations of `briefs/make-tests-actually-run-impl-notes.md`, the one brief that *is* tracked. Reverted; those four stand.

Measured on this tree: 26 raw candidates → 10 real findings, **0 false positives**. The three exclusions that get it there — doc placeholders, ancestor-relative refs, and PTX mnemonics like `ld/st.shared` lexing as `st.sh` — are each documented in the header with the case that forced them.

## Prove-red

`check-readme-repo-links.sh` (baseline exit 0):

| mutation | exit |
|---|---|
| badge points at SPOC again | 1 |
| an Actions link names another repo | 1 |
| `GITHUB_REPOSITORY` names a different repo | 1 |
| README absent | 2 (fails closed) |
| unparseable slug | 2 |

`check-cited-paths-exist.sh` (baseline exit 0):

| mutation | result |
|---|---|
| dangling citation on one line | exit 1 |
| the same citation **wrapped** across two comment lines | exit 1 |
| unwrap disabled, wrapped citation present | that finding disappears |
| outside a git tree | exit 2 |

The wrapped case matters: ocamlformat broke one of the four original `roster/` citations mid-path, so a line-oriented scan would have missed the very citation the gate exists to catch. Comment continuations are joined before matching.

## Verification

- `dune build` — clean
- `dune runtest` — exit 0 (only the pre-existing df64 KNOWN-DEVIATION warnings)
- `dune build @fmt` — exit 0, **measured directly rather than after a pipe**
- `scripts/check-formal-proofs.sh` — exit 0: 3 projects rebuilt from source and kernel-re-checked, ledgers match, axioms match the allowlist

One consequence surfaced during that last run and is fixed here: untracking the extraction output did not stop `@fmt` churning it, because dune still sees the files on disk. `FMT_EXIT=1` on `AGpuSemantics.ml` right after a proof run; `.ocamlformat-ignore` closes it. That this is a recurring cost is on the record — the repo carries a stash named *"pre-reconcile: stale extraction regen artifacts (EC-5)"*.

## Deliberately not done

- `gh-pages/docs/publications.md` keeps its dozen talk-PDF links under `mathiasbourgoin/SPOC`; those files really live there and rewriting the URLs would break them. The badge gate is scoped to CI surfaces only.
- `Extract.v` stays in the build — compiling it still asserts the theories are extractable. Migrating its five modules to the self-contained `Extraction "file.ml"` form and *linking* them, which is what would let `test_codegen_ptx_conformance` test the extracted model instead of the `ptx_cmp_tag` it declares itself, is the open half of backlog-187's audit.

## Summary by CodeRabbit

* **Documentation**
  * Updated the README “Build Status” badge to point to the correct workflow.
  * Refreshed various documentation comments to consistently reference notes that aren’t included in the repository.

* **CI / Tests**
  * Strengthened CI documentation gates by validating README CI badge/repository links and that cited documentation paths resolve to tracked files.
  * Added extra coverage harnesses and an additional invariant check during the build.

* **Chores**
  * Improved ignoring/formatting behavior for regenerated proof/extraction artifacts.
  * Removed obsolete PTX code-generation artifacts and public interfaces.


mathiasbourgoin:
Rebased onto main (four PRs landed under it) and both problems fixed.

## The `build` failure was mine

`check-license-headers.sh` stamped my two new scripts and exited 1 — they had no SPDX header. Caught by CI, not by me: I ran the two new gates locally and the full test suite, but not the license gate. That is precisely the "ran the gates I was thinking about" blind spot, and it is the same shape as the finding this PR is about.

## CodeRabbit's finding was right and is fixed

`check-cited-paths-exist.sh` scanned the whole file text, so a legitimate string literal like `"fixtures/missing.md"` would have failed CI over a path that was never a citation. A gate that blocks a correct change is worse than the drift it was written to catch.

Everything outside an OCaml comment is now blanked before matching. It is a small state machine rather than a regex, and both halves are load-bearing: OCaml comments **nest**, so a nested close must not terminate the outer one; and a string literal may contain `(*`, so string state must be tracked or a literal could open a comment that swallows real code. Newlines are preserved, so reported line numbers still point at the real line.

### Prove-red — the discriminating pair

The important thing to rule out is that the scanner blanked *everything*, which would leave the gate inert and green. So the **identical** string is tested in both positions:

| case | expectation | result |
|---|---|---|
| `roster/gone/L99-note.md` in a comment | found | exit 1 ✓ |
| the same, **wrapped** across two comment lines | found | exit 1 ✓ |
| the same, inside a **string literal** | ignored | exit 0 ✓ |
| the same, in `["..."]` inside a comment | found (odoc code ref is comment text) | exit 1 ✓ |

Rows 1 and 3 are the same bytes in different syntactic positions with opposite verdicts — that is what shows the scoping discriminates rather than suppresses.

### Verification after the rebase

- `dune build` — 0
- `dune runtest` — 0
- `dune build @fmt` — 0 (measured directly, not after a pipe)
- `check-license-headers.sh` — 0
- `check-readme-repo-links.sh` — 0
- `check-cited-paths-exist.sh` — 0
- `prove-red.sh` — 0: **6 subjects, 28 mutations**, every declared mutation produced its declared exit code and message, every baseline green

Rebase conflicts resolved two ways, both recorded: `formal/codegen-ptx/PtxKernelSpec.ml` was modify-on-main vs delete-here, resolved as **delete** (that is the point of the change); `ci.yml` gained a `check-tvar-id-allocator` step from #381, kept alongside the two new gate steps.

mathiasbourgoin:
## Triage of the three actionable comments on the latest review

Review history here is CHANGES_REQUESTED -> APPROVED -> CHANGES_REQUESTED, and the
last review (2026-07-30T06:22:47Z) predates the commit that answered it. All three
comments are **already fixed**; below is the evidence I ran, plus the test case that
pins each one so the fix cannot silently regress.

### 1. `scripts/check-cited-paths-exist.sh` — "Only scan OCaml comments, not all source text"

**already-fixed** (you tagged it yourself: *Addressed in commits 80846e0 to 467b2ea*).

Citations are now extracted from `comments_only(raw)`, not the raw file. Pinned by:

```
PASS green: the same path inside a STRING LITERAL is not a citation (exit 0)
PASS red:   a path quoted INSIDE a comment is still a citation (exit 1)
```

Both polarities on purpose. A fix that stopped scanning string literals *and* stopped
seeing `["path/to/x.ml"]` inside a comment would satisfy the first case alone, and odoc
code refs legitimately put citations in quotes inside comments.

### 2. `scripts/check-cited-paths-exist.sh` — "Handle OCaml character and quoted-string literals"

**already-fixed**, in `622d15d` — 27 minutes *after* this review was submitted, which is
why it is untagged rather than unaddressed.

`comments_only` now tracks three literal forms, not one: `"..."`, `CHAR_LIT =
'(?:\\.|[^'\\])'`, and `QUOTED_OPEN = \{([A-Za-z_][A-Za-z0-9_']*)?\|` with its exact
matching `|id}` close. I re-ran **your own adversarial samples** against the current
function:

| sample | citation seen | expected |
|---|---|---|
| `let h = '#' in (* See roster/missing.md *)` | YES | YES |
| `let q = '"' in (* See roster/missing.md *)` | YES | YES |
| `{tag\| (* See roster/missing.md *) \|tag}` | no | no |
| `let x = (* see (* nested *) roster/missing.md *) ()` | YES | YES |
| `let _ = "(* ignored" in (* See roster/missing.md *)` | YES | YES |
| `let p = {\|raw (* text\|} in (* See roster/missing.md *)` | YES | YES |

All six as expected. Regression fixtures for both literal forms:

```
PASS green: a path inside a {| |} quoted string is not a citation (exit 0)
PASS green: {id| |id} holding an unbalanced (* does not open a comment (exit 0)
PASS red:   a real citation AFTER a quoted string is still found (exit 1)
PASS red:   a citation after a '"' char literal is still found (exit 1)
```

The two `red` cases are the load-bearing ones: without them, "handles `{|`" would be
indistinguishable from "stops scanning at the first `{|`", and "handles char literals"
from "never re-enters comment state".

**Proven red, not assumed.** Reverting only the subject to its pre-fix revision
(`git checkout 0f001e39 -- scripts/check-cited-paths-exist.sh`, harness left at HEAD)
gives `10 passed, 3 FAILED`:

```
FAIL green: {id| |id} holding an unbalanced (* does not open a comment: exit 1, wanted 0
FAIL red:   a citation after a '"' char literal is still found: exit 0, wanted 1
```

Note the direction of the second one — the old code exited **0** where a real dangling
citation existed. Your diagnosis was right and the miss was silent: `Buffer.add_char buf
'"'` flipped `in_string` on and never off, blanking every comment for the rest of the
file while the gate reported full coverage. Two tracked files contain that exact token
(`sarek/codegen/Sarek_wgsl_abi.ml:53`, `benchmarks/to_csv.ml`), so this was measurably
blind over their tails, not hypothetically.

### 3. `scripts/check-readme-repo-links.sh` — "Normalize standard SSH URI remotes"

**already-fixed** in `622d15d` (you tagged this one).

Implemented as a superset of the proposed diff: all four GitHub spellings reduce to
`owner/repo` — `git@github.com:`, `https?://github.com/`, `ssh://git@github.com/`,
`git://github.com/` — and the slug is then validated as exactly
`^[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+$`. `GITHUB_REPOSITORY` still wins when set. One
deliberate difference from the suggestion: the validation uses `grep -Eq` rather than
bash `[[ =~ ]]`, same predicate without depending on bashisms in a script run by `sh`
in some carriers.

Pinned by five cases, including the no-`GITHUB_REPOSITORY` local-fallback harness case
you asked for:

```
PASS green: badge matches the git remote (no GITHUB_REPOSITORY) (exit 0)
PASS green: ssh:// remote (exit 0)
PASS green: git:// remote (exit 0)
PASS red:   a slug with a space is not owner/repo (exit 2)
PASS red:   a three-segment slug is not owner/repo (exit 2)
```

Reverting only this subject to `0f001e39` gives `8 passed, 8 FAILED`, `green: ssh://
remote: exit 1, wanted 0` among them — the exact "correct README fails on a correct
repo during local fallback" you described.

### Nothing was relaxed

Both scripts still **exit 2, never skip**, when they cannot run — unreadable source, no
`GITHUB_REPOSITORY` and no remote, an unparseable slug, no CI surface in the README.
`check-cited-paths-exist.sh` still resolves citations against `git ls-files` rather than
the filesystem, which is deliberate: an untracked scratch file is not something a reader
can open, so it is not a satisfied citation.

Gates on this branch, each exit captured directly: `dune build` 0, `dune runtest` 0,
`dune build @fmt` 0, `check-license-headers.sh` 0, `check-kb-properties.sh` 0,
`prove-red.sh` 0, `check-cited-paths-exist.sh` 0, `check-readme-repo-links.sh` 0,
`check-test-alias-coverage.sh` 0, `check-cited-paths-exist.test.sh` 0 (13 cases),
`check-readme-repo-links.test.sh` 0 (16 cases).

Branch is now at `40431ba8`, three commits ahead of the `622d15d` this review read.
Please re-review.

PR #388, opened 2026-07-30
fix(privacy): stop publishing machine identifiers; drop the 263 leaked files

mathiasbourgoin:
Closes backlog-168. Blocks backlog-202 (self-hosted GPU lanes would have published more identifiers on their first green run).

Found while checking whether it was safe to record the GPU test fleet — the answer was that the thing being avoided had already happened.

## What was exposed, measured on `origin/main`

| surface | extent |
|---|---|
| tracked files with a hostname **in the filename** | **263** (107 + 78 + 78, three machines) |
| each payload's `system` block | hostname, exact kernel version, `cpu.model`, host `memory_gb`, device names |
| `gh-pages/benchmarks/data/latest.json` (~600 KB) | all three hostnames + CPU + kernel — **already served** |
| CSV export | leaked independently of the JSON, via its own header column |
| `benchmarks/CONTRIBUTING.md` | instructed contributors to put `$(hostname)` in a **branch name** and a **commit message** |
| `run_all_benchmarks.sh` | *printed* a suggested commit message containing `$(hostname)` |

The last two matter beyond this repo: the documented workflow asked every external contributor to repeat the leak.

## Where the line is

**Removed** — hostname (an identity), kernel version (a patch level), host `memory_gb` (a weak fingerprint no chart read).
**Kept** — `os`, `cpu.model`, cores, threads, device names.

A benchmark number is unreadable without the hardware that produced it, so hardware *model* is not the line; identity and patch level are. If the models must go too, the dashboard's premise needs rethinking — a bigger decision than this change.

## `hostname` was load-bearing

It keyed dedup (`deduplicate_results.ml`) and the "replace this machine's old results" glob. So it is **replaced, not deleted**, by `machine` — an opaque label derived from the *kept* hardware facts (`<os>-<gpu vendor>`), never from the hostname.

Deliberate: a hash of the hostname would be trivially reversible (hostnames are low entropy, a wordlist breaks them), and a committed hostname→label table would itself be the leak. Deriving from data we publish anyway discloses nothing new.

Two same-hardware machines derive the same label and would be merged, so `SAREK_BENCH_MACHINE` overrides it — and is **refused** when set to the hostname. An escape hatch that silently reintroduces the leak is worse than no escape hatch.

## Decisions taken by Mathias

- **History: dropped.** 263 files removed rather than renamed. Identifiers remain in git history; removing them would rewrite published SHAs and break every `backlog-NN` commit citation. Recorded as a residual, not as done.
- **Published data: closed first**, ahead of other queued work. Scrubbed *in place* rather than blanked so the dashboard keeps its numbers — 235 system blocks → 3 labels, verified by raw string search that `cachyos` / `ladon` / `drangleic` / `hostname` / `kernel` all reach count **0**.

## A regression this nearly shipped

The viewer read `system.hostname` in **13** places. Scrubbing the payload alone would have rendered every entry as "Unknown" — a silent regression inside the fix. Migrated, then re-verified by grouping the scrubbed payload through the viewer's own logic: 3 groups, **0 entries dropped**.

## New gate — five red shapes, proven independently

`scripts/check-no-machine-identifiers.sh`, wired into `ci.yml` with its harness.

Payload / path / producer are separate checks because the leak had separate shapes: scrubbing payloads does not fix a filename, and the CSV header leaked on its own. The path check is a label-**shape allowlist**, not a blocklist of the three known hostnames — a blocklist would pass a fourth machine, which is how this class survives. Exit 2 outside a git work tree or without the scrubber.

Harness — 12 cases, all measured:
- **5 red**: dirty payload; hostname filename *with an already-clean payload*; `hostname` process read; re-emitted JSON field; re-added CSV column.
- **2 fail-closed**: not a work tree, missing scrubber.
- **3 green** pinning what must *not* be refused — including `system_info.ml`'s own sanctioned hostname read, since if that were red the gate would forbid the safety check that rejects a hostname override.

Confirmed the harness fails against both an always-exit-0 **and** an always-exit-1 stub.

The KB gate earned its keep: it caught the new gate being *declared but not invoked from any carrier*, before this commit existed.

## Kept deliberately

Ten `7.1.2-3-cachyos` occurrences under `docs/measurements/` and in fp-contraction sources. Those are kernel **release strings** in measurement provenance — "cachyos" there is the distro suffix, not a machine name (checked: zero of the ten are hostnames) — and they are load-bearing evidence for which toolchain produced a given divergence count. Flagging rather than stripping; say the word if you want them gone too.

## Residuals, stated plainly

1. Identifiers remain in **git history** (chosen).
2. The **deployed** site stays as-is until this merges and redeploys.
3. Neither of those is re-checked by anything — the gate covers the working tree only, and says so in its KB row.

## Gates

`dune build` 0 · `runtest` 0 · `@fmt` 0 (exit captured directly) · license 0 · KB 0 (35 declarations) · prove-red 0 · new gate 0 · new harness 12/12.

## Summary by CodeRabbit

* **New Features**
  * Benchmark results now use stable machine labels instead of hostnames.
  * Added a tool to remove identifying system details from benchmark JSON files.
  * CSV exports and filenames now include machine labels for clearer comparison.
* **Documentation**
  * Updated benchmark run, publishing, storage, and output-format guidance.
* **Bug Fixes**
  * Result cleanup now safely targets the detected machine label and avoids deleting files when ambiguous.
* **Tests**
  * Added coverage for identifier checks, failure handling, and approved exceptions.
* **Chores**
  * Removed legacy benchmark results containing machine-specific metadata.


mathiasbourgoin:
@coderabbitai — all five actionable comments and the nitpick verified against the current code. **All six confirmed, all six fixed** in 66e6bf4f. Nothing was refused, but two suggestions were implemented differently from the wording; those are called out below.

The privacy property was the constraint throughout: `scripts/check-no-machine-identifiers.sh` still refuses all three surfaces the leak actually used, and the operator-override refusal in `get_machine_label` is untouched.

---

### 1. `benchmarks/CONTRIBUTING.md:40` — placeholders in copy-paste commands — **CONFIRMED, fixed**

`benchmark-results-rtx4090` and `linux-nvidia` were literals. Now `benchmark-results-<your-gpu>`, consistent between the `checkout -b` and the `push`, and the commit message takes `<machine-label>` with a note that this is the derived `<os>-<gpu-vendor>`.

One thing beyond what you flagged, in the same block: `git add results/run_*/` added nothing. `run_all_benchmarks.sh` moves results out of the timestamped run directory into `benchmarks/results/` and then removes the run directory, so the documented path never exists by the time a contributor runs it. Corrected to `benchmarks/results/*.json`.

### 2. `benchmarks/README.md:88` + `:123` — layout vs generator and deduplicator — **CONFIRMED, fixed on the docs side**

Verified: `deduplicate_results.ml:127` reads `benchmarks/results` and enumerates it with `Sys.readdir` — flat, no descent. `Output.make_filename` (`output.ml:171`) already writes `<machine>_<benchmark>_<size>_<timestamp>.json`, so the machine label is in the *filename* and the per-machine directory was redundant. `run_all_benchmarks.sh` moves everything flat into `benchmarks/results/`.

So the tools are self-consistent with each other and the **docs were the only wrong side** — I did not touch the deduplicator.

Implemented more broadly than the two lines you cited, deliberately: the same `results/<machine>/` convention appeared again in the Example Workflow (the `mkdir -p results/nvidia-rtx3090` block and the `results/*/*.json` globs). Fixing only 79-88 and 123 would have reproduced exactly the inconsistency you flagged, three sections lower.

That removes the hand-typed `MACHINE=linux-nvidia` convention as a side effect, which is worth stating on this PR: a doc that tells people to name a committed path after their machine is how `$(hostname)` got pasted into 263 filenames in the first place.

### 3. `benchmarks/run_all_benchmarks.sh:203` — constrain cleanup — **CONFIRMED, fixed**

Reachable, not theoretical. `system.machine` is operator-settable through `SAREK_BENCH_MACHINE`, and `get_machine_label` only refuses the value that equals the hostname — every other string passes through `String.trim`. So `*` was reachable, and it expands to `benchmarks/results/*_*.json`: **every machine's results, not just this one's**. `../..` escaped the directory outright.

Now: validated against a filename-safe contract before it is used as a path fragment, enumerated with a quoted glob inside `benchmarks/results` only, removed with `rm --`. This also clears SC2012/SC2086 on those lines.

Note this is strictly *narrowing* — it does not become the label authority. What may be **committed** stays governed independently by `check-no-machine-identifiers.sh`, which is the surface that matters for disclosure; this check is about not deleting the wrong local files.

### 4. `benchmarks/run_all_benchmarks.sh:275` — derive the label independently of `--no-clean` — **CONFIRMED, fixed**

Confirmed by reading: `MACHINE` was assigned inside `if [ "$CLEAN_OLD" = true ]`, so `--no-clean` reached `${MACHINE:-your-machine-label}` and printed the literal placeholder. Derivation moved before the cleanup guard; deletion stays conditional on `CLEAN_OLD`.

### 5. `benchmarks/system_info.ml:88` — stabilize the label — **CONFIRMED, fixed. This was the substantive one.**

Both defects reproduced by execution. They matter more than "minor" suggests, because since this PR the label *is* the dedup key and the leading token of every result filename — the slot the hostname used to occupy — so an unstable label splits one machine into several identities with nothing reporting an error.

- **CPU outvoting a GPU.** Exclusion was `lower d.name <> lower cpu_model`. An OpenCL CPU device's name and `/proc/cpuinfo`'s `model name` differ by whitespace runs and trailing text, so the CPU survived and its brand-string vendor token beat a real discrete GPU.
- **Order dependence.** `List.find_map` over `devices` took the first match, so a discrete + integrated box relabelled whenever the driver enumerated differently.

Fixed as you describe — CPU devices excluded in one pass (by backend for `Native`/`Interpreter`, by normalized containment otherwise), vendor chosen from the whole surviving set by a fixed priority rather than by traversal order.

**One deliberate divergence from your wording, and it turned out to be load-bearing.** You asked to exclude CPU-backed devices "including names differing from cpu_model by whitespace or additional text". Applied unconditionally, that is wrong: on a unified-memory SoC the GPU's name **is** the CPU's name (`Apple M4 Max`), so the name test removes the only GPU on the machine. My first attempt did exactly that and the new test caught it. The name test is therefore applied only on backends where a CPU can masquerade as a device (OpenCL); a CUDA/Metal/Vulkan/HIP device is a GPU by construction.

That surfaced a **pre-existing bug** in the code this PR is replacing: the old `name = cpu_model` exclusion had the same blind spot, so every Apple Silicon machine derived `darwin-unknown` instead of `darwin-apple` — silently, because `unknown` is a legal label. Now fixed.

### Nitpick, `scripts/deidentify-benchmark-results.py:144-150` — unconditional rewrite — **CONFIRMED, fixed**

Reproduced: an already-clean payload reported `scrubbed 0 system block(s)` and was still re-serialized. Write is now conditional on `count`, as you proposed. Nothing depended on the tool normalizing formatting — the CI gate uses `--check`, which reads fields and ignores layout.

**This one had a consequence worth flagging.** Fixing #5 without also fixing this script would have introduced a corruption: `gpu_vendor` here mirrored the old OCaml logic, and this script *relabels*. Verified by execution before fixing it — it rewrote a correct `darwin-apple` payload to `darwin-unknown`, desynchronizing the payload from its own filename, which carries the producer's label. The two derivations have to agree, so the Python is updated in step and now says so in its docstring. The published payload is unchanged by this: 235 entries, 3 labels (`linux-intel`, `linux-nvidia`, `darwin-apple`), all preserved, and re-running the scrubber over it now produces no diff at all.

---

### Privacy property — still enforced on all three surfaces

```
$ scripts/check-no-machine-identifiers.sh
OK -- 1 payload(s) carry no identifying field
OK -- 1 tracked payload(s), no machine identifier in payloads, paths or producers
exit 0
```

No suggestion narrowed the gate. Nothing in it was loosened; the accept-list on tracked filenames, the payload check, and the producer checks (`hostname` read outside the one sanctioned call site, a removed field re-emitted into JSON, the CSV header regaining the column) are byte-identical to before this commit. `get_machine_label` still **fails** — not warns — on an override equal to the hostname.

### Tests

Extended the existing harness rather than adding a parallel one.

`scripts/check-no-machine-identifiers.test.sh`, **12 → 23 cases**:

- an already-clean payload is left byte-identical;
- an identifying payload is *still* rewritten and comes out clean — the opposite polarity, so the skip cannot silently become a blanket never-write;
- producer/scrubber agreement on `darwin-apple`, pinning the incoherence above;
- the filename-safe label contract, with the destructive shapes: `*`, `../../etc`, embedded space, `-rf`, empty. The pattern is **extracted from `run_all_benchmarks.sh`**, not restated — a copy would let someone loosen the script while the test kept asserting the old pattern and stayed green.

`benchmarks/test/test_system_info.ml`, new, **18 cases** over `gpu_vendor_of`: the CPU-outvoting shapes, all six permutations of a three-GPU set, and greens pinning what must *not* be excluded (Apple Silicon, CUDA-only, failed CPU probe).

### Prove-red

Each fix committed first, then reverted with `git checkout 303c9e8f -- <path>` and re-run:

| reverted | result |
|---|---|
| `benchmarks/system_info.ml` | `test_system_info` **10 of 18 cases fail** — both defects and the Apple Silicon case |
| `scripts/deidentify-benchmark-results.py` | harness red: already-clean rewritten, `darwin-apple` relabelled (21 pass, and the opposite-polarity case correctly stays green) |
| `benchmarks/run_all_benchmarks.sh` | harness **exit 2, fails closed** — the pattern cannot be extracted, so it refuses to pass rather than skipping the label cases |

Gates, each exit code captured directly: `dune build` 0 · `dune runtest` 0 · `dune build @fmt` 0 · `check-license-headers.sh` 0 · `check-kb-properties.sh` 0 · `prove-red.sh` 0 · `check-no-machine-identifiers.sh` 0 · `check-no-machine-identifiers.test.sh` 0 (23/23) · `check-test-alias-coverage.sh` 0 (new target wired, 283 targets).

---

### One thing left open for @mathiasbourgoin — not from this review

There is a tension between the override and the gate that predates this review and that I did not resolve, because resolving it either way is a judgement call about published data.

`SAREK_BENCH_MACHINE` exists to keep two same-hardware machines apart. But the gate's tracked-path accept-list is `(linux|darwin|windows)-(nvidia|amd|intel|apple|unknown)`, so a disambiguating label such as `linux-nvidia-b` produces filenames the gate **refuses to let you commit**. The override therefore works for local runs but cannot deliver its stated purpose for results published to this repo.

Loosening the accept-list would weaken the exact surface that caught the 263 filenames, so I did not touch it. I documented the actual constraint in the README instead of writing an example that would fail the gate. Worth a decision at some point — it is a small hole in the override's usefulness, not in the privacy property.

PR #389, opened 2026-07-30
fix(bench): the machine-label override worked, except for the purpose it exists for

mathiasbourgoin:
## The contradiction

`SAREK_BENCH_MACHINE` exists for exactly one reason: two machines with the **same**
hardware profile derive an identical `<os>-<vendor>` label, and their runs then merge
in the dedup key. #388 shipped that override — and a commit gate whose filename
accept-list was `(linux|darwin|windows)-(nvidia|amd|intel|apple|unknown)`, with no
suffix in it. So the disambiguating label the override exists to set, `linux-nvidia-b`,
produced filenames `scripts/check-no-machine-identifiers.sh` refused to commit.

The override worked locally and nowhere it mattered. This fleet has two `linux-amd`
machines that collide today.

## The shape

```
^(linux|darwin|windows)-(nvidia|amd|intel|apple|unknown)(-[a-z0-9]{1,8})?$
```

`os` and `vendor` stay **enumerated**. That is the part that refuses a bare hostname:
`drangleic` has no `<os>-<vendor>` prefix, the prefix is mandatory, and no widening of
the suffix can supply one. It remains an allowlist on the label **shape**, never a
blocklist of the three known hostnames — a blocklist passes the fourth machine, which
is how this class survives.

**Why 8, and why one segment.** Eight characters covers every disambiguator a fleet
actually needs (`b`, `2`, `lab2`, `office`, `rack12`) and falls short of a hostname —
the three that leaked are 9 and 10 characters. The bound is not the security boundary
(the enumerated prefix and the producer's hostname refusal are); it bounds how much
free-form operator text can ride along in a committed filename, and keeps the label
under ~24 chars so filenames stay readable. One segment rather than
`(-[a-z0-9]{1,8})*`, so the bounded field cannot grow by repetition into the
free-form field the bound is there to deny.

## One rule, one place

The shape lives in `scripts/machine-label-shape.sh`, which the gate **sources**.
`benchmarks/machine_label.ml` restates it, because OCaml cannot source a shell file.

The two are not held together by a comment saying they must agree. The red path
**executes both** over one shared case table — `machine_label.ml` is Stdlib-only
precisely so `ocaml` can `#use` it — and compares the pattern literals byte for byte.
Widening only the OCaml bound from 8 to 12 goes red on the literal comparison, on the
shared table, and on the override policy.

## The override now validates its own input

An operator could previously set a label the gate would refuse hours later, at commit
time. `Machine_label.resolve` now refuses anything outside the shape.

The **hostname refusal is untouched and still a hard `failwith`** — and it runs
**first**. Order is load-bearing: `linux-amd` is a legal shape and could be somebody's
hostname, so checking the shape first would let that one through with nothing red
anywhere, right next to the check added to tighten it. Both refusals and their order
are pinned; swapping them turns two cases red.

Only the **override** is shape-validated. The **derived** label is not, so a run on an
OS outside the enumeration behaves exactly as before rather than gaining a new refusal.

## Two consumers that would have broken quietly

Neither was in the reported symptom:

- `run_all_benchmarks.sh`'s filename-safety predicate refused a suffixed label, so the
  one machine that *needs* the suffix is the one whose stale results silently stop
  being cleaned and accumulate under its key.
- The scrubber relabelled `linux-amd-b` → `linux-amd`, stripping the only thing
  separating two same-hardware machines and desynchronizing the payload from its own
  filename, which still carries the suffix. Same defect class as the `darwin-apple`
  relabelling already pinned in the harness, one field over.

Both fixed, both pinned in **both** polarities (an illegal suffix must still be
relabelled; a label outside the enumeration must still be refused for commit).

## Dead code

`System_info.get_kernel_info` is deleted. Zero callers, and it produced the exact
`uname -r` patch level backlog-168 removed. It survived #388 only to avoid resetting a
fresh approval on a PR that needed merging.

## Tests — 23 → 72 cases, all proven red

Extended `scripts/check-no-machine-identifiers.test.sh`; no parallel harness.

| Reverted / mutated | Goes red |
|---|---|
| gate → pre-change accept-list | suffixed label not committable; missing-shape fail-closed |
| scrubber → pre-change | suffixed payload relabelled away |
| runner predicate → pre-change | `linux-amd-b`, `linux-amd-rack12` read unsafe |
| override policy → pre-change (no shape check) | `bare-word`, `suffix-too-long`, `suffix-illegal-char`, `unknown-vendor`, `glob` all accepted |
| shape checked **before** hostname | `hostname`, `hostname-cased` → `refused:shape` instead of `refused:hostname` |
| path shape over-widened to `/[a-z0-9-]+_` | bare hostname accepted; over-bound suffix accepted |
| OCaml bound widened 8 → 12 only | literal drift + shared table + override policy |

The harness was re-confirmed against a subject that always exits 0 and one that always
exits 1: 17 failing cases and harness exit 1 in **both** directions, so it is not
reading a fixed exit code as agreement.

Gates, each exit code captured directly: `dune build` 0 · `dune runtest` 0 ·
`dune build @fmt` 0 · `check-license-headers.sh` 0 · `check-kb-properties.sh` 0 ·
`prove-red.sh` 0 · `check-no-machine-identifiers.sh` 0 ·
`check-no-machine-identifiers.test.sh` 0.

`kb/properties.md`'s `KB-GATE-NO-MACHINE-IDENTIFIERS` row is updated: it described the
strict accept-list, and a row wider than its code is the defect this repo keeps closing.

## Known gap

`get_os_info` can return an OS outside the enumeration (`uname -s` = FreeBSD); the
derived label would then be un-committable. Pre-existing, unchanged here, flagged not
fixed — widening the platform enumeration is a separate decision.

## Summary by CodeRabbit

* **New Features**
  * Added support for optional short suffixes on benchmark machine labels to distinguish otherwise identical results.
  * Benchmark labels can now be explicitly overridden, subject to format and privacy safeguards.

* **Bug Fixes**
  * Preserves valid disambiguating labels during result de-identification.
  * Rejects hostname-based labels and unsafe or malformed label values.
  * Improved validation prevents unsafe benchmark filenames and path handling.

* **Documentation**
  * Clarified machine-label formatting rules and privacy checks.


mathiasbourgoin:
Both review findings confirmed by execution, both fixed in 2ebb82d5, plus the CI failure the PR description flagged as unverified.

**CI (`build (ubuntu-latest, 5.4.0, true)` / "Harness self-tests", run 30548834544).** It got all the way to the cross-language block and stopped there:

```
PASS red: scrubber relabels a suffix that is not legal
##[error]ocaml required to run benchmarks/machine_label.ml
##[error]Process completed with exit code 2.
```

Every other case passed; the missing thing was a toplevel, not a behaviour. Fixed by *providing* `ocaml`, not by skipping the comparison — a skip-as-pass is the failure mode that check exists to close.

Why `apt-get install ocaml-nox` and not `spoc-ci:latest`, where the rest of the OCaml work runs: that image installs `opam` but no compiler, so `ocaml` only exists after "Build CI image" **and** "Build SPOC packages" have created the switch in `.opam-ci`. Running the harness there would move a fail-closed guard behind the two slowest steps in the job, and the harness also drives `scripts/deidentify-benchmark-results.py`, for which `ci/Dockerfile` installs no `python3` either — leaning on the oneAPI base image to happen to carry one is the sort of undeclared dependency `ci/assert-toolchain.sh` exists to prevent. A distribution toplevel is sufficient here, and checking that is part of the point: `machine_label.ml` is Stdlib-only by construction, so a version-specific dependency creeping into it now turns the step red rather than going unnoticed. The step ends in `ocaml -version` because `apt-get install` succeeding is not the same claim as `ocaml` being runnable.

**`scripts/check-no-machine-identifiers.sh:80` — CONFIRMED.** Reproduced exactly as described: `/<label>_` was unanchored, so a *directory* vouched for the file beside it. `benchmarks/results/linux-amd_/drangleic_vector_add_1024_2026-07-30T00-00-00.json` was selected as a result file by the first pattern and then excused as well-labelled by `/linux-amd_` two components to its left, and the hostname was committable.

Fixed slightly differently from the suggested diff, to avoid the reason it was needed. The suggestion restates the filename tail a second time, and two copies of one pattern in one file is the drift `machine-label-shape.sh` exists to prevent — so the tail now has a single definition (`MACHINE_LABEL_RESULT_TAIL`) and both the selector and the allowlist compose it. Same effect (`[a-z_0-9]` excludes `/`, so the match is confined to one component), one authority. Both polarities are pinned: the disguise is red, and a correctly labelled file one directory down stays green, so binding the label to the filename cannot quietly become "the label must be in the top directory".

**`benchmarks/machine_label.ml:110` — CONFIRMED**, and the first of your two remedies is the right one. `SAREK_BENCH_MACHINE=linux-intel-b` on an AMD box was accepted, written into the payload and into every filename; the scrubber then derived `linux-amd` from the payload's own hardware, `keeps_label` refused it, and the field was rewritten to `linux-amd` while the file stayed named `linux-intel-b_…`. A payload disagreeing with its own name is the desynchronization backlog-168b set out to remove, one field over.

An override must now be the derived label, or that label plus a legal suffix (`Machine_label.is_derived_variant`, the same rule as the scrubber's `keeps_label`). That is also what the suffix is *for* — separating two machines with the **same** hardware; re-badging one as different hardware was never the purpose and the scrubber would not honour it.

The second remedy — make the scrubber preserve every well-formed label — was not taken: it would let a payload assert hardware it does not have, and recomputing the label from the payload is the scrubber's whole job. Note the new rule leaves the producer deliberately *stricter* than the commit gate. That asymmetry is the safe one: producer-stricter costs an operator one error message, producer-looser is a completed run whose results cannot be committed, which is the bug this PR started from.

**`scripts/check-no-machine-identifiers.test.sh:133-146` (nitpick) — CONFIRMED**, applied. It is the same "a gate reads green having checked nothing" shape the rest of this file is about; a fixture failure was reported as a gate failure.

**Evidence.** Each fix proven red by reverting only its own file(s) with `git checkout f43a941d -- <path>` against the committed harness:

| Reverted | Result |
|---|---|
| `machine-label-shape.sh` + `check-no-machine-identifiers.sh` | exit 1 — `FAIL red: a label-shaped directory does not excuse a hostname filename` |
| `machine_label.ml` | exit 1 — `FAIL override other-vendor-base … got ok:linux-intel-b`, `FAIL override other-os-base … got ok:darwin-amd` |
| one-sided widening: OCaml `suffix_max_len` 8→12 only | exit 1 — `FAIL shape 'linux-amd-123456789' … bash said unsafe, ocaml said safe` + `FAIL override suffix-too-long` |

That last row is the one that matters for the CI change: the anti-drift mechanism still catches a bound that moves in one language and not the other. The harness is 78 cases now (was 72) and still fails against both an always-exit-0 and an always-exit-1 gate stub (19 failed cases in each direction).

Unchanged: `os` and `vendor` stay enumerated, the hostname refusal stays first and stays a hard `failwith`, and `linux-amd` / `linux-amd-b` / `linux-amd-rack12` stay accepted while `drangleic`, `cachyos`, `ladon`, `linux-amd-toolongsuffix`, `linux-amd-A`, `linux-amd-b-c`, `linux-banana` and `freebsd-nvidia` stay refused.

Gates: `dune build` 0 · `dune runtest` 0 · `dune build @fmt` 0 · `check-license-headers.sh` 0 · `check-kb-properties.sh` 0 · `prove-red.sh` 0 (6 subjects, 28 mutations) · `check-no-machine-identifiers.sh` 0 · `check-no-machine-identifiers.test.sh` 0 (78/78).

PR #390, opened 2026-07-30
fix(ppx): refuse `when` guards on match cases instead of dropping them

mathiasbourgoin:
## What was wrong

`Sarek_parse.parse_expression`'s `Pexp_match` arm read `case.pc_lhs` and `case.pc_rhs` and never looked at `case.pc_guard`. A `when` guard was **discarded in the parser** — upstream of type-checking, of every lowering pass, and of backend selection — so the guarded arm became unconditional in the single AST every backend is generated from. A kernel with a guarded arm compiled with no diagnostic and computed a different function than its source says.

This is the repository's dominant defect class in its purest form: the accepted subset acting as a silent filter where it should be a refusal boundary.

## Reproduced before fixing, on every device present

Not taken on trust. A temporary e2e harness with a **load-bearing** guard, so a dropped guard is wrong *data* and not merely an unobservable control-flow difference:

```ocaml
let s = Circle src.(tid) in
match s with
| Circle r when r >. 10.0 -> r *. 2.0
| Circle r -> r +. 100.0
| Square q -> q
```

`src.(i) = i+1`, n=64. Source semantics: elements 0..9 take arm 2 (`r +. 100.0`), the rest arm 1. Dropped-guard semantics: every element takes arm 1. **Two oracles** were evaluated per device, so the verdict is a positive identification rather than "not equal to the right answer".

It compiled at exit 0 with zero diagnostics, then:

| Framework | Device | Verdict | wrong vs source | wrong vs dropped-guard | first 8 |
|---|---|---|---|---|---|
| Interpreter | CPU Interpreter (Sequential) | GUARD DROPPED | 10/64 | 0/64 | 2 4 6 8 10 12 14 16 |
| Interpreter | CPU Interpreter (Parallel, 32 cores) | GUARD DROPPED | 10/64 | 0/64 | 2 4 6 8 10 12 14 16 |
| Native | CPU Native (Parallel, 32 cores) | GUARD DROPPED | 10/64 | 0/64 | 2 4 6 8 10 12 14 16 |
| CUDA/PTX | AMD Radeon RX 7900 XTX **[ZLUDA-on-AMD]** | GUARD DROPPED | 10/64 | 0/64 | 2 4 6 8 10 12 14 16 |
| CUDA/PTX | AMD Ryzen 9 7950X **[ZLUDA-on-AMD]** | GUARD DROPPED | 10/64 | 0/64 | 2 4 6 8 10 12 14 16 |
| OpenCL | AMD Radeon RX 7900 XTX (radeonsi, navi31, ACO) | GUARD DROPPED | 10/64 | 0/64 | 2 4 6 8 10 12 14 16 |
| OpenCL | AMD Ryzen 9 7950X (radeonsi, raphael_mendocino) | GUARD DROPPED | 10/64 | 0/64 | 2 4 6 8 10 12 14 16 |
| Vulkan | AMD Radeon RX 7900 XTX (RADV NAVI31) | GUARD DROPPED | 10/64 | 0/64 | 2 4 6 8 10 12 14 16 |
| Vulkan | AMD Ryzen 9 7950X (RADV RAPHAEL_MENDOCINO) | GUARD DROPPED | 10/64 | 0/64 | 2 4 6 8 10 12 14 16 |

9 devices, 5 frameworks, identical. **Confirmed exactly as filed**, and the drop changes results, not just control flow. No backend differed — which is not luck: the guard is dropped before a backend is chosen.

There is **no NVIDIA hardware on this host**; the CUDA/PTX rows are ZLUDA-on-AMD and are labelled as such. **Metal was not measured** (no Apple hardware) and nothing in this change claims anything about it.

## Refused, not lowered — and why

I checked whether the guard could be lowered correctly instead, since that would be strictly better. It is not feasible as a small change:

1. `Sarek_ir_ppx`'s `EMatch` and `SMatch` hold `(pattern * _) list` — **no guard slot to lower into**. The IR type must change, and with it the CUDA, OpenCL, Metal and PTX emitters, the tag-erasure and vector-inlining traversals, and the native and interpreter evaluators.
2. Every C-family emitter builds a match as a nested tag ternary whose **last arm is emitted unconditionally**, and PTX branches to the last arm unconditionally. A guarded arm must be able to **fall through** to the next one; that shape cannot express it.
3. `Sarek_ir_ptx_expr.check_match_exhaustive` reasons on **constructor coverage alone**. A guarded arm covers its constructor syntactically but not semantically, so once guards exist a match can fail at run time — and the device backends have **no trap to fail into**. The interpreter has `Interp_error.Pattern_match_failure`; the GPU backends have no counterpart.

A correct lowering is a feature with an IR type change in it, not a fix. So: refuse, scoped to one parser site.

The message says **"not yet implemented"**, deliberately not "cannot be supported" — the subset where a guarded arm is always followed by an unguarded arm for the same constructor *is* implementable, so the stronger claim would be false.

It fires at **PPX time**: `Parse_error_exn` is converted by `Sarek_ppx.ml` into a located `Location.raise_errorf`, with the caret on the **guard expression itself** — so the user sees it once and portably, not at device-compile time.

### On the message's claims

It names the **mechanism** (the IR has no guard slot) rather than a count of backends. This repo has repeatedly shipped messages claiming "all five backends" when the measurement was two; a mechanism stated upstream of backend selection is both stronger and checkable by reading one type definition. Each clause maps to evidence:

| Claim | Verified by |
|---|---|
| "not supported in kernels" | the refusal sits on the sole guard-carrying parse path; `function`-style cases were already refused separately |
| "not yet implemented" (not "cannot be") | deliberately the weaker claim |
| "Sarek's match IR has no guard slot" | `Sarek_ir_ppx.ml` — `EMatch of expr * (pattern * expr) list`, `SMatch of expr * (pattern * stmt) list` |
| "would be discarded and the arm would match unconditionally" | measured: 9/9 devices returned exactly the dropped-guard oracle |
| "the kernel would compile and run" | measured: build exit 0, executed on 9 devices |
| "computing something other than what this source says" | measured: 10/64 elements wrong vs source, every device |
| the suggested `if` rewrite is expressible | `EIf` exists in the IR and is lowered by every backend |

## The test, and the trap it avoids

`sarek/tests/negative/test_when_guard.ml`, declared as `neg_test_when_guard` and asserted in `make test_negative` **on the message text**, not on a non-zero exit.

It is deliberately **self-contained** — no `open Spoc`, no `open Kirc`. Eight sibling cases (`test_reserved_keyword`, `test_convention_kernel_fail2`, `test_reserved_prefix_{field,helper,for,local,param}`, `test_inline_node_exhaustion`) carry such an open while the stanza declares only `sarek sarek.stdlib sarek.ppx.lib`, so neither module resolves; they pass **only** because the PPX refusal fires before type-checking reaches the open, with `-w -33` hiding the hint. That is backlog-208 and it is **not fixed here** — but this case declares its own `[@@sarek.type]` and nothing else, so its red is caused by the refusal and nothing else.

### Both polarities, actual output

**With the fix** — exit **1**:

```
File "sarek/tests/negative/test_when_guard.ml", line 49, characters 26-35:
49 |           | Circle r when r >. 10.0 -> r *. 2.0
                               ^^^^^^^^^
Error: Sarek parse error: `when` guards on match cases are not supported in
kernels (not yet implemented). Sarek's match IR has no guard slot, so this
guard would be discarded and the arm would match unconditionally ...
```

**Fix reverted** via `git checkout <base> -- <path>` (never `git stash` — foreign stashes are live in this repo) — exit **0**, the case **silently compiles**:

```
File "sarek/tests/negative/test_when_guard.ml", line 50, characters 12-20:
50 |           | Circle r -> r +. 100.0
                 ^^^^^^^^
Warning 11 [redundant-case]: this match case is unused.
```

`grep -c` for the asserted message returns `0`. That warning is itself independent evidence of the drop: OCaml can only call the **second** `Circle` arm redundant if the guard had already been erased before type-checking.

Whole-gate polarity too — reverted, `make test_negative` reds **at this case specifically**, with every preceding case still PASS:

```
Testing `when` guard rejection on a match case (backlog-191)...
Warning 11 [redundant-case]: this match case is unused.
make: *** [Makefile:122: test_negative] Error 1
```

The fix was then restored from a saved patch and every gate re-run green, with `git status --porcelain` confirming no mutation left live.

## Gates — each exit code captured directly, never `$?` after a pipe

| Gate | Exit |
|---|---|
| `dune build` | 0 |
| `dune runtest --force` (with `LD_LIBRARY_PATH=$HOME/opt/zluda`) | 0 |
| `dune build @fmt` | 0 |
| `scripts/check-kb-properties.sh` | 0 (40 code-intel declarations hold) |
| `scripts/prove-red.sh` | 0 (6 subjects, 30 mutations, every declared mutation red as declared) |
| `scripts/check-negative-case-coverage.sh` | 0 (**30** declared, all asserted — was 29) |
| `make test_negative` | 0 (`PASS: when guard rejected`) |

`--force` was used on `runtest` so a cached alias could not emit zero output that reads as a pass.

## Out of scope, flagged not fixed

- **backlog-208** — the eight siblings still carry unresolvable opens.
- `check_match_exhaustive`'s constructor-coverage-only reasoning stays as-is; it is sound *because* guards are now refused, and becomes a live problem only if guards are implemented.
- Actually implementing guard lowering (the IR change above) is a feature and wants its own brief.

## Not verified

Metal (no Apple hardware) and real NVIDIA CUDA (no NVIDIA hardware) were not measured. Neither the message nor this description makes a claim about either.


mathiasbourgoin:
@coderabbitai review

New round pushed as `4a1223dd`. Prose only — three blockers from the last adversarial pass, all sentences claiming more than the code does. The refusal itself is unchanged and was confirmed correct.

**1. `sarek/ppx/Sarek_parse.ml` — a closed "every consumer" list that was wrong twice.** It named "the CUDA, OpenCL, Metal and PTX emitters, the tag-erasure and vector-inlining traversals, and the native and interpreter evaluators". `git grep -l "EMatch\|SMatch"` reports 37 non-test source files; the list omitted `Sarek_ir_glsl` and `Sarek_ir_wgsl` (both match on `EMatch`/`SMatch` directly), softmath, monomorphisation, defunctionalisation, the three tailrec passes, convergence, fusion, `Sarek_lower_ir`, `Sarek_ir_conv`, `spoc/ir`'s analysis/codegen/pp, and the `formal/type-safety` models.

It was also wrong on identity: the emitters do not consume `Sarek_ir_ppx`'s `EMatch` at all. There are two IRs — `Sarek_ir_ppx` (PPX-time) and `Sarek_ir_types` (`spoc/ir`, what codegen reads; `Sarek_ir_cuda.ml:19` opens it) — bridged by `Sarek_ir_conv`, plus a third like-named `EMatch` on `Sarek_ast`. Now stated as "at least the six device emitters plus roughly a dozen IR passes", with the grep to re-run rather than a list to trust.

**2. `kb/sarek/ppx/parser.md` — "surfaced as a located `Location.raise_errorf` at PPX time" is true on the `[%kernel]` route only.** That conversion exists at exactly one place, `Sarek_ppx.ml:1999`, inside the `[%kernel]` expander; `process_structure_for_module_items` (`:2289`) has no handler. Measured: an in-file guarded `let[@sarek.module]` is swallowed by the self-scan catch-all (`:571-581`) and the user sees `Error: Unbound variable: <helper>`; a sibling-file helper escapes as an unlocated `Fatal error: exception ...Parse_error_exn(...)`. Pre-existing, not guard-specific, fail-closed on every path. Scoped the sentence; filed the handler fix as #391 rather than bundling it.

**3. "compiled with no diagnostic" — false for this kernel, true for other shapes.** A wrong-*answer* repro needs two arms on one constructor, and dropping the guard then leaves two identical arms, i.e. warning 11 `redundant-case`, which the e2e flags (`-w -32-33-34-69`) leave on as an error. The bug is genuinely silent only where no two arms share a constructor. Both halves now stated in `parser.md` and in the negative test's header.

Worth your attention on:
- Whether the warning-11 explanation is right about the e2e flag set actually leaving 11 enabled.
- Whether #391 is correctly scoped as diagnostic-quality-only (I claim every path is fail-closed — please try to find one that is not).
- Whether the new blast-radius paragraph still overstates anything, given I moved it out of the bullet list to survive ocamlformat.

PR #394, opened 2026-07-30
fix(codegen): a nested record's struct was declared after the struct using it

mathiasbourgoin:
## What was wrong

A `[@@sarek.type]` record with a **record-typed field** had its inner struct emitted **after** the struct that referenced it, so no kernel touching a nested record compiled on any shader backend.

Both declaration-emission loops walked `kern_types` in list order:

- `spoc/ir/Sarek_ir_codegen.ml` → `gen_record_typedefs` (C family: OpenCL, CUDA-C, HIP, Metal)
- `sarek/codegen/Sarek_ir_glsl.ml` → the `gen_record_def` loop (GLSL / Vulkan)
- `sarek/codegen/Sarek_ir_wgsl.ml` → the `gen_record_def` loop (WGSL — same shape, not previously measured)

List order is not dependency order: the PPX prepends the types reachable through the registry to the ones the kernel payload declares, so the struct that *uses* a record type can land ahead of the struct it names.

This is **not** a missing-emission bug and **not** a field-store bug. The reproducer here is a **read-only** nested access — `dst.(tid) <- src.(tid).mid.b`, no field store anywhere — and it fails identically. Interpreter and Native were never affected: they carry values, not struct declarations.

## Measured, on this host (RX 7900 XTX + Raphael iGPU)

Nested-record case, per device — the CUDA rows are **ZLUDA on AMD**, there is no NVIDIA, Metal or HIP hardware here.

| Device | Framework | Before | After |
|---|---|---|---|
| CPU Interpreter (Sequential) | Interpreter | PASS ×3 | PASS ×3 |
| CPU Interpreter (Parallel, 32 cores) | Interpreter | PASS ×3 | PASS ×3 |
| CPU Native (Parallel, 32 cores) | Native | PASS ×3 | PASS ×3 |
| RX 7900 XTX | CUDA/PTX — **ZLUDA-on-AMD** | PASS ×3 | PASS ×3 |
| Ryzen 9 7950X | CUDA/PTX — **ZLUDA-on-AMD** | PASS ×3 | PASS ×3 |
| RX 7900 XTX (radeonsi, navi31) | OpenCL | **FAIL ×3** | PASS ×3 |
| Ryzen 9 7950X (radeonsi, raphael_mendocino) | OpenCL | **FAIL ×3** | PASS ×3 |
| RX 7900 XTX (RADV NAVI31) | Vulkan | **FAIL ×3** | PASS ×3 |
| Ryzen 9 7950X (RADV RAPHAEL_MENDOCINO) | Vulkan | **FAIL ×3** | PASS ×3 |

`×3` = the three nesting shapes (one-level, three-level chain, two independent nested types). 27 device/case rows, all green after the fix; 12 red before.

Before, OpenCL:

```
input.cl:3:3: error: unknown type name 'Test_record_nested_struct_order_triple'
    3 |   Test_record_nested_struct_order_triple mid;
```

Before, Vulkan (glslang never names the identifier):

```
ERROR: /tmp/sarek_0d77e5.comp:8: '' :  syntax error, unexpected IDENTIFIER, expecting RIGHT_BRACE
```

The CUDA/PTX rows pass in both columns because the direct PTX emitter flattens aggregates and emits no struct declarations. The CUDA **C** path shares `gen_record_typedefs` with OpenCL, so it is fixed by the same change — but it is not exercised on this host, and neither are Metal or HIP. Their coverage here is source-level only (see below).

## The fix

`Sarek_ir_codegen.sort_record_types_by_dependency` places a declaration only after every declaration its field types reference. Dependencies are collected from the whole field type tree — through `TArray`, `TVec`, variant payloads, and the fields of a nested record — by `referenced_record_names`, which is guarded against a cyclic `elttype` *value* rather than assuming a finite tree. Comparison is on mangled names, which is what the emitted identifier actually is.

All three loci now call it. `gen_record_typedefs` calls it internally, so CUDA-C, HIP and Metal are fixed without touching their files.

**Ties break on the incoming list position, never on the name.** The algorithm repeatedly takes the earliest still-unplaced entry whose dependencies are all placed, so two independent records keep their relative input order and an already-correctly-ordered list comes back unchanged. That is what makes this safe for the goldens: `dune runtest --force` regenerated **zero** committed golden files, with and without ZLUDA. `git status --porcelain` was empty after both runs. There was no golden diff to explain.

**A cycle is refused, loudly** — `Record_type_cycle` carrying the unplaced names, rather than a fall-back to input order, since a silently wrong order is the exact defect being removed.

And the honest part about that refusal: **it cannot fire through the PPX.** The PPX resolves a record field's alignment from a registry populated when the field type's *own* declaration was processed, so a cycle is refused there first. Measured, both spellings:

```
type selfy = {v : float32; me : selfy} [@@sarek.type]
  → sarek: unknown alignment for field type 'selfy' - register it with [%ktype] before ...

type a = {to_b : b} [@@sarek.type] and b = {to_a : a} [@@sarek.type]
  → sarek: unknown alignment for field type 'b' - register it with [%ktype] before ...
```

The comment on the function says this rather than claiming a load-bearing guard. The check is kept for the hand-built-IR path (fusion, tests, a future front end), where nothing else would notice, and the unit test is the only place it can be exercised.

A *self*-referencing field is deliberately not reported as a cycle — the self-edge is dropped, so the diagnostic stays about cycles between distinct declarations — and that drop is pinned by a test so it reads as a decision, not an accident.

## Tests

- `sarek/tests/e2e/test_record_nested_struct_order.ml` (new) — the three nesting shapes on **every** available device, with no per-backend tolerance; a device failure exits non-zero.
- `spoc/ir/test/test_sarek_ir_codegen_record_order.ml` (new) — 10 cases pinning what no device here can observe: the C-family emission order (the **only** coverage Metal and HIP get on a host with neither), stability on independent records, no-op on already-sorted input, dependency edges through arrays and variant payloads, the cycle refusal, the self-reference non-refusal, and termination of `referenced_record_names` on a genuinely cyclic value.

## Prove-red

Both polarities, with the mutation asserted to have landed before either was trusted.

**1. Whole fix reverted** (`git checkout 93b865f5 -- <4 files>`; `grep -c sort_record_types_by_dependency` = 0 in all four). E2E exit 1, original errors back: `unknown type name 'Test_record_nested_struct_order_triple'` on OpenCL ×2, `syntax error, unexpected IDENTIFIER` on Vulkan ×2, all three shapes. The unit test cannot even build at this polarity (`Unbound value ... sort_record_types_by_dependency`) — a build-level red, which is weaker than an assertion, hence polarity 2.

**2. Call sites unwired, function kept** (`perl -0pi` on the three loci; asserted 0 remaining call sites *and* the definition still present, `grep -c "^let sort_record_types_by_dependency"` = 1). The unit test now fails on an assertion, not a link error:

```
Fatal error: exception Failure("C-family emission: `triple` used at 17 before its declaration at 91:
typedef struct {
  triple mid;
} outer;

typedef struct {
  float a;
  float b;
  float c;
} triple;
")
```

and the e2e test goes red on 12 of 27 rows with the same device errors. Restored from `HEAD` afterwards; tree clean, both green again.

`git stash` was not used at any point — `git checkout <ref> -- <path>` only.

## Gates

| Gate | Exit |
|---|---|
| `dune build` | 0 |
| `dune runtest --force` (with ZLUDA) | 0 — 5419 lines of output, not a cached no-op |
| `dune runtest --force` (without ZLUDA) | 0 — 5157 lines |
| `dune build @fmt` | 0 (after `dune promote`) |
| `scripts/check-kb-properties.sh` | 0 |
| `scripts/prove-red.sh` | 0 — 6 subjects, 30 mutations |
| `scripts/check-opam-clean.sh` | 0 |
| `test_record_nested_struct_order.exe` with ZLUDA | 0 |
| `test_record_nested_struct_order.exe` without ZLUDA | 0 |
| `test_sarek_ir_codegen_record_order.exe` | 0 |

No `kb/properties.md` or `.github/workflows/ci.yml` change, so no shared-file contention with the sibling branches.

## Two things a reviewer needs to know

**1. The known-gap tolerances are not in this PR, because they are not on `main`.** The task asked for `is_backlog203_struct_gap` to be deleted from `sarek/tests/e2e/test_record_field_store.ml` and for the sibling tolerance in `test_record_local_alias_agreement.ml` to go with it. Neither file exists on `main` — both are introduced by **#376** (`fix/interp-record-field-store`), which is currently `CONFLICTING`/`DIRTY`, and its test files depend on that PR's interpreter/native changes, so they cannot be lifted onto `main` on their own. Basing this fix on #376 would have stacked a `main`-facing codegen fix behind an unmergeable branch, so it is based on `main` as instructed and the deletion is left as a required follow-up: **#376 must drop `is_backlog203_struct_gap` (4 references) and the backlog-203 tolerance in `test_record_local_alias_agreement.ml` before it lands**, or it merges a tolerance for a gap that no longer exists. Flagged on that PR.

**2. An adjacent ordering gap this PR deliberately does not touch — and it has TWO halves, not one.** GLSL and WGSL emit records *before* variants; OpenCL/CUDA/Metal emit variants *before* records. Records and variants are each sorted within their own loop, so neither ordering fixes an edge that crosses between them, and the asymmetry means **each family is red on exactly the shape the other family is green on**:

- a record with a **variant**-typed field is emitted ahead of its variant struct on **GLSL/WGSL**;
- a variant with a **record** payload is emitted ahead of that record across the whole **C family**.

An earlier revision of this section named only the first half and asserted the C family was clear. That was a claim wider than the code. The review pass measured both halves from ordinary `[@@sarek.type]` source: the C-family half is **T3-reachable**, `error: unknown type name 'Probe_pt'` on both OpenCL devices while Vulkan compiles the same kernel, and the GLSL/WGSL half is red on both Vulkan devices while both OpenCL devices pass. Either half read on its own reproduces as "ordering is fixed", which is how it stayed invisible.

Neither half is fixed here — the fix is an interleaved emission loop per backend family, which is a different change from this one. Both are now documented with their measured errors at `gen_record_typedefs`, in the `.mli`, and in `CHANGES.md`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -104,12 +104,19 @@
   failed the same way. `Sarek_ir_codegen.sort_record_types_by_dependency` now
   orders the declarations, stably — ties break on the incoming list position,
   so an already-correct order is returned unchanged and no committed golden
-  moved. A record cycle raises `Record_type_cycle` instead of being emitted in
-  some wrong order; that path is unreachable through the PPX, which refuses a
-  self- or mutually-referencing record field while resolving its alignment
-  (measured: both spellings stop at *unknown alignment for field type*), and is
-  kept for hand-built IR. Interpreter and Native were never affected — they
-  carry values, not struct declarations.
+  moved. A cycle between distinct record declarations raises `Record_type_cycle`
+  instead of being emitted in some wrong order (a self-referencing field is
+  dropped, not reported); that path is unreachable through the PPX, which
+  refuses a self- or mutually-referencing record field while resolving its
+  alignment (measured: both spellings stop at *unknown alignment for field
+  type*), and is kept as a backstop for hand-built IR. Interpreter and Native
+  were never affected — they carry values, not struct declarations. This orders
+  records against records only. Records are NOT ordered against variants, and
+  both directions of that gap are still live: on the C family, which emits
+  variants first, a variant with a record payload names a struct declared later
+  (measured on OpenCL, `error: unknown type name 'Probe_pt'`); on GLSL/WGSL,
+  which emit records first, the mirror case is a record with a variant-typed
+  field.
 - `Sarek_df64` silently ran at plain float32 precision on real NVIDIA
   hardware (CUDA/PTX and NVIDIA OpenCL): `ptxas` contracted the multiply in
   `two_prod` into the `add`/`sub` of the `quick_two_sum` closing `df64_mul`,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -92,6 +92,24 @@
 
 ### Fixed
 
+- A `[@@sarek.type]` record with a record-typed field had its inner struct
+  emitted AFTER the struct that referenced it, so no kernel touching a nested
+  record compiled on any shader backend. Both declaration-emission loops walked
+  `kern_types` in list order, and that is not a dependency order: the PPX
+  prepends the types reachable through the registry to the ones the kernel
+  payload declares. Measured on an RX 7900 XTX host, a read-only nested access
+  (`dst.(tid) <- src.(tid).mid.b`) failed on OpenCL ×2 with
+  `unknown type name 'Test_..._triple'` and on Vulkan ×2 with a glslang parse
+  error at the field line; a three-level chain and two independent nested types
+  failed the same way. `Sarek_ir_codegen.sort_record_types_by_dependency` now
+  orders the declarations, stably — ties break on the incoming list position,
+  so an already-correct order is returned unchanged and no committed golden
+  moved. A record cycle raises `Record_type_cycle` instead of being emitted in
+  some wrong order; that path is unreachable through the PPX, which refuses a
+  self- or mutually-referencing record field while resolving its alignment
+  (measured: both spellings stop at *unknown alignment for field type*), and is
+  kept for hand-built IR. Interpreter and Native were never affected — they
+  carry values, not struct declarations.
 - `Sarek_df64` silently ran at plain float32 precision on real NVIDIA
   hardware (CUDA/PTX and NVIDIA OpenCL): `ptxas` contracted the multiply in
   `two_prod` into the `add`/`sub` of the `quick_two_sum` closing `df64_mul`,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -116,7 +116,9 @@
   variants first, a variant with a record payload names a struct declared later
   (measured on OpenCL, `error: unknown type name 'Probe_pt'`); on GLSL/WGSL,
   which emit records first, the mirror case is a record with a variant-typed
-  field.
+  field (measured on Vulkan/RADV, `syntax error, unexpected IDENTIFIER`). Each
+  half is reachable from ordinary `[@@sarek.type]` source and each passes on the
+  family the other fails on.
 - `Sarek_df64` silently ran at plain float32 precision on real NVIDIA
   hardware (CUDA/PTX and NVIDIA OpenCL): `ptxas` contracted the multiply in
   `two_prod` into the `add`/`sub` of the `quick_two_sum` closing `df64_mul`,
--- a/sarek/codegen/Sarek_ir_glsl.ml
+++ b/sarek/codegen/Sarek_ir_glsl.ml
@@ -2030,8 +2030,12 @@ let generate_with_types ?block ?(log : string -> unit = fun _ -> ())
        ~uses_uint8:(Sarek_ir_analysis.kernel_uses Sarek_ir_analysis.Coopmat k)
        ()) ;
 
-  (* Generate record type definitions (simple structs without tag) *)
-  List.iter (gen_record_def buf) types ;
+  (* Generate record type definitions (simple structs without tag), inner
+     structs first — GLSL requires a struct type to be declared before the
+     field that names it (backlog-203). *)
+  List.iter
+    (gen_record_def buf)
+    (Sarek_ir_codegen.sort_record_types_by_dependency types) ;
 
   (* Generate variant type definitions (structs with tag) *)
   List.iter (gen_variant_def buf) k.kern_variants ;
--- a/sarek/codegen/Sarek_ir_wgsl.ml
+++ b/sarek/codegen/Sarek_ir_wgsl.ml
@@ -1482,7 +1482,11 @@ let generate_with_types ?block ?(log : string -> unit = fun _ -> ())
   scalar_param_names := [] ;
   current_variants := k.kern_variants ;
   let buf = Buffer.create 1024 in
-  List.iter (gen_record_def buf) types ;
+  (* Inner structs first: WGSL, like GLSL, requires a struct type to be
+     declared before the field that names it (backlog-203). *)
+  List.iter
+    (gen_record_def buf)
+    (Sarek_ir_codegen.sort_record_types_by_dependency types) ;
   List.iter (gen_variant_def buf) k.kern_variants ;
   let scalars = gen_bindings buf k.kern_params in
   scalar_param_names := scalars ;
--- a/sarek/tests/e2e/dune
+++ b/sarek/tests/e2e/dune
@@ -1588,3 +1588,34 @@
  (alias runtest)
  (action
   (run %{dep:test_helper_param_const_collision.exe})))
+
+; record-nested-struct-order (backlog-203): a [@@sarek.type] record with a
+; record-typed field must have the inner struct emitted BEFORE the struct that
+; references it. Covers one-level nesting, a three-level chain, and two
+; independent nested types in one record. Runs on every available device with
+; no per-backend tolerance — the C-family "unknown type name" and the
+; GLSL/WGSL parse error at the field line are exactly what it detects.
+
+(executable
+ (name test_record_nested_struct_order)
+ (modules test_record_nested_struct_order)
+ (libraries
+  sarek
+  sarek.stdlib
+  sarek_ir
+  spoc_core
+  sarek_native
+  sarek_interpreter
+  sarek_cuda
+  sarek_opencl
+  sarek_vulkan
+  sarek_backend_error)
+ (preprocess
+  (pps sarek_ppx))
+ (flags
+  (:standard -w -32-33-34-69)))
+
+(rule
+ (alias runtest)
+ (action
+  (run %{dep:test_record_nested_struct_order.exe})))
--- a/sarek/tests/e2e/test_record_nested_struct_order.ml
+++ b/sarek/tests/e2e/test_record_nested_struct_order.ml
@@ -1,0 +1,206 @@
+(******************************************************************************)
+(* SPDX-License-Identifier: CECILL-B                                          *)
+(* SPDX-FileCopyrightText: 2026 Mathias Bourgoin <mathias.bourgoin@gmail.com> *)
+(******************************************************************************)
+
+(******************************************************************************
+ * E2E test: a [@@sarek.type] record with a RECORD-TYPED field must have its
+ * inner struct declared BEFORE the struct that uses it (backlog-203).
+ *
+ * The declaration-emission loops walked [kern_types] in list order, which is
+ * not a dependency order: the PPX merges registered types ahead of the
+ * payload's own, so a record whose field type is another record could be
+ * emitted first and reference an as-yet-undeclared struct. Every C-family
+ * backend (OpenCL/CUDA/HIP/Metal) then failed at compile time with
+ * `unknown type name '<Inner>'`, and GLSL/WGSL with a parse error at the
+ * field line. The values-carrying backends (Interpreter, Native) never see a
+ * struct declaration and were unaffected.
+ *
+ * Three shapes are exercised, all of them beyond the trivial one-level pair:
+ *   1. one-level nesting     — [outer { tag; mid : triple }]
+ *   2. a THREE-level chain   — [chain_top { s : chain_mid { r : chain_leaf } }]
+ *   3. two INDEPENDENT       — [twin { left : triple; right : chain_leaf }]
+ *      nested types            (the sort must order both, deterministically)
+ *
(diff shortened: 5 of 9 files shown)


mathiasbourgoin:
@coderabbitai review

CI is green on this head. Requesting a review so the PR can satisfy the repository's review requirement.

Three things worth attacking specifically:

1. **Stability.** Ties in the topological sort break on incoming list index, never on name, so an already-correct list must be returned unchanged — that is what kept the golden count at zero. Please check no input ordering can produce a different output than the one it came in with when it was already valid.

2. **The cycle refusal is a backstop, not a load-bearing guard**, and the comment says so. Reachability was measured: both `type selfy = {v : float32; me : selfy}` and a mutually-recursive pair are rejected earlier by the PPX with `unknown alignment for field type`, so no cycle survives to codegen through the front end. Please check that framing is accurate rather than the check being dead code described as live.

3. **Coverage boundary.** The fix lands at three loci; Metal, HIP, CUDA-C and WGSL have no hardware or runtime here, so their coverage is the shared `gen_record_typedefs` plus source-level assertions in the new unit test. WGSL's fix in particular is by-shape-analogy and unexecuted.

Known and deliberately out of scope: GLSL/WGSL emit records before variants while the C family emits variants first, so a record with a *variant*-typed field may still be mis-ordered on GLSL/WGSL. Unmeasured, filed separately.

mathiasbourgoin:
## Independent adversarial review — GO (after 3 fix commits)

Reviewed at `3c83400c`, rebased onto `origin/main` (`15686c8f`, `range-diff` reports the commit identical, verdict carried not re-minted), then three fixes of my own on top. Own worktree, `git stash` never used, `git checkout <ref> -- <path>` only.

Final head: `1e4a3977`.

### What I verified, and what settled it

| Claim | Verdict | Command that settled it |
|---|---|---|
| Kahn sort + `referenced_record_names`, wired into GLSL and WGSL | **holds** | source read + `grep -c 'sort_record_types_by_dependency types'` = 1 in each of the three loci |
| Tie-break is on incoming index, never on name; already-sorted input unchanged | **holds** | `test_stable_on_independent_records` uses `zeta; alpha; mu; beta` — no comparison function produces that. `List.find_opt` over an index-preserving `List.filter` is the lexicographically-least valid topological order; nothing version- or hash-dependent in the implementation |
| Zero golden churn | **holds** | `dune runtest --force` exit 0, 5157 lines, `git status --porcelain` **empty** |
| 12 red rows before | **holds, exactly 12** | call sites unwired → `dune exec .../test_record_nested_struct_order.exe` exit 1, 12 device/case `FAILED` rows with `unknown type name 'Test_record_nested_struct_order_triple'` (OpenCL ×2) and the glslang parse error (Vulkan ×2). 21/21 green after (no ZLUDA in my run; 21 rows, not 27) |
| CUDA/PTX emits no structs | **holds** | `Sarek_ir_ptx_kernel.generate_with_types ~types:_` — the argument is discarded at the binder. Not "flattens aggregates", it never looks |
| HIP/CUDA-C/Metal fixed without touching their files | **holds** | `sarek-hip/Hip_plugin.ml:36`, `sarek-cuda/Cuda_c_plugin.ml:38`, `Sarek_ir_metal.ml:1217` all reach `gen_record_typedefs` |
| A cycle cannot reach codegen through the PPX | **holds, re-measured both spellings** | `type selfy = {v : float32; me : selfy}` → `sarek: unknown alignment for field type 'selfy'`; `type a = {to_b : b} and b = {to_a : a}` → same for `'b'`. Also checked the other hand-built route: `Sarek_fusion` only concatenates two PPX-produced type lists, so it cannot introduce a cycle either |
| Two prove-red polarities, the second landing on an assertion | **holds** | unwire mutation asserted landed (0 call sites, definition still present), then exit 2 with the exact `` `triple` used at 17 before its declaration at 91 `` failure the PR quotes |

`scripts/prove-red.sh` reports **6 subjects, 33 mutations**, not the 30 in the gate table — `#393` landed three more. Stale number, nothing else.

### Three defects found, all fixed here

**1. `referenced_record_names` HANGS on a cyclic value — three artifacts said it does not.** (`e3c34cbe`)

The visited set was keyed on the mangled **record name**, which closes a cycle only if a `TRecord` sits on it. OCaml's recursive-value rule admits cycles that do not:

```
let rec t = TVec t
let rec t = TVariant ("c", [("C", [t])])
```

Both loop forever — tail-recursively, so they **hang** rather than overflow. Measured: each shape hit a 20 s `timeout`, **exit 124**, before the fix.

Three places claimed otherwise, and this is the repo's usual class:
- `.ml`: "`[seen]` stops a cyclic `elttype` value from looping — the walk is over a graph, not a guaranteed-finite tree"
- `.mli`: "safe on a cyclic `[elttype]` value"
- the unit case named `test_referenced_names_terminates_on_cyclic_value`, exercising only the record shape — the one shape the guard actually closed

The guard is now keyed on physical node identity, which closes every shape (skipping a physically re-visited node cannot lose a name — `acc` already holds everything under it), and all three shapes are pinned.

**2. Two `.mli` claims wider than the code.** (`e3c34cbe`)
- `@raise Record_type_cycle` said "if the declarations form a cycle". A **self**-cycle does not raise: `deps` drops the self-edge on purpose. Only the `.ml` said so; the exported contract did not.
- The `.mli` called the refusal "**load-bearing** for hand-built IR" while the `.ml` called it a backstop nothing would notice and the PR body says "**not** a load-bearing guard". Three artifacts, two positions. Nothing reaches it (PPX refuses first, fusion concatenates acyclic lists), so the `.ml`/PR-body reading is the true one and the `.mli` now matches it.

**3. The residual ordering gap is named in one direction; it has two, and the unnamed half is the one that fails on this host's hardware.** (`e3c34cbe`, `1e4a3977`)

The PR flags "GLSL and WGSL emit records before variants; OpenCL/CUDA/Metal emit variants before records. A record with a **variant**-typed field is therefore still emitted ahead of its variant struct **on GLSL/WGSL**." It stops there, which reads as "the C family's record ordering is done". The same asymmetry, read the other way, puts a **variant's record payload** ahead of the record on the whole C family. Reachable straight from PPX source and executed red:

```
type pt = {x : float32; y : float32} [@@sarek.type]
type shape = Nothing | Point of pt   [@@sarek.type]
```

```
[OpenCL] AMD Radeon RX 7900 XTX (radeonsi, navi31): FAILED
  input.cl:5:5: error: unknown type name 'Probe_pt'
      5 |     Probe_pt Point_v;
  input.cl:15:50: error: unknown type name 'Probe_pt'
[OpenCL] AMD Ryzen 9 7950X (radeonsi, raphael_mendocino): FAILED (identical)
[Vulkan] AMD Radeon RX 7900 XTX (RADV NAVI31): PASSED (dst0=1.00)
[Vulkan] AMD Ryzen 9 7950X (RADV RAPHAEL_MENDOCINO): PASSED (dst0=1.00)
```

I also measured the half the PR flagged but left unmeasured — a record with a variant-typed field:

```
[OpenCL] both devices: PASSED (dst0=2.00)
[Vulkan] both devices: FAILED  ERROR: /tmp/sarek_9c4532.comp:7: '' : syntax error, unexpected IDENTIFIER
```

**Each half is green on exactly the family the other fails on.** That is why a reproducer built on one family alone reads as "ordering fixed" — and it is the reason the one-directional framing is the dangerous form of the claim, not a wording nit. Neither half is fixed here (ordering records against variants needs one interleaved loop per backend family, with its own reproducer), but both are now written down at `gen_record_typedefs`, in the `.mli`, and in `CHANGES.md`, with the measured error for each. **The PR body should be corrected to name both halves** — as written it asserts the C family is clear.

### And one defect I introduced and then fixed

My first commit added the three cycle-shape cases with **no deadline**, and the comment said "dune's timeout is the red". `dune runtest` has no timeout. A regression would have hung the suite indefinitely, noticed eventually by a CI job limit with no message naming the cause — a gate that cannot fail, added by the review that exists to find them. `5848bd9f` puts each case under a SIGALRM deadline so the hang becomes a named `Failure`.

Proved red after committing the fix, mutation asserted to have landed first (anchor asserted present, so a no-op patch could not collect an unearned red):

- before the deadline harness: exit **124**, no message
- after: exit **2**, `referenced_record_names on a TVec cycle: did not terminate within 10s — the cyclic-value guard no longer closes this shape`

### Gates, on `1e4a3977`, each captured

| Gate | Exit |
|---|---|
| `dune build` | 0 |
| `dune runtest --force` | 0 — 5157 lines, `git status --porcelain` empty |
| `dune build @fmt` | 0 |
| `scripts/prove-red.sh` | 0 — 6 subjects, 33 mutations |
| `scripts/check-kb-properties.sh` | 0 |
| `scripts/check-license-headers.test.sh` | 0 |
| `scripts/check-opam-clean.sh` | 0 |
| `scripts/check-cited-paths-exist.test.sh` | 0 |

CI was green on the rebased `15686c8f` (build / check-generated-code / formal-proofs / pocl-runner-probe all pass); the three review commits need a fresh run.

### What I could NOT verify

- **Metal and HIP remain source-level only.** No hardware here. Their coverage is the `gen_record_typedefs` unit case, which pins the shared emitter, not either backend's own path — the PR says this and it is accurate.
- **CUDA-C.** `Cuda_c_plugin` exists and is fixed by the shared emitter, but nothing exercised it: the CUDA rows on this host go through the PTX plugin, which discards `~types` entirely.
- **The 27-row / ZLUDA column.** My runs had no ZLUDA (21 rows, 5157 lines — matching the PR's own no-ZLUDA figure exactly). The 6 CUDA/PTX rows are unverified by me, though the `~types:_` binder makes them uninteresting.
- **WGSL on a device.** No WebGPU runtime here; WGSL's fix is source-level plus the shared sort's unit coverage.
- **The `.mli` prose corrections have no runnable red.** A doc comment cannot be proved red; they are verified by reading them against the code they describe.
- **The e2e test is vacuous on a deviceless host** — `Array.length devs = 0` exits 0 with "No devices found". Consistent with the rest of the e2e suite, so not raised as a finding, but it means CI green on a deviceless runner says nothing about this test.

Issue #391, opened 2026-07-30
Parse_error_exn is only located on the [%kernel] route; [@sarek.module] swallows or fatals it


mathiasbourgoin:
`Sarek_parse.Parse_error_exn` is only converted to a located diagnostic on the `[%kernel]` route. On the `[@sarek.module]` route it is either swallowed by the self-scan or escapes as an unlocated `Fatal error`.

## Root cause

The `Parse_error_exn (msg, loc) -> Location.raise_errorf ~loc` conversion exists at exactly one place, `sarek/ppx/Sarek_ppx.ml:1999`, inside the `[%kernel]` expander's `try ... with`. `process_structure_for_module_items` (`sarek/ppx/Sarek_ppx.ml:2289`, registered as the `~impl` pass at `:2530`) has no handler for it.

`git grep -n Parse_error_exn sarek/ppx/Sarek_ppx.ml` returns one line.

## Measured behaviour

Two distinct wrong-looking outcomes, neither a located PPX diagnostic:

1. **Guarded `let[@sarek.module]` in the same file as its use.** The self-scan's catch-all (`Sarek_ppx.ml:571-581`) converts any scan exception into a warning and skips the file's Sarek registrations:

   ```
   Sarek PPX: scanning <path> for [@sarek.*] declarations failed (Parse_error_exn(...)); skipping this file's Sarek registrations.
   ```

   The compiler then fails on the consequence rather than the cause — `Error: Unbound variable: pick`. The real message never reaches the user.

2. **Guarded `[@sarek.module]` helper in a sibling file.** The exception escapes the impl pass:

   ```
   Fatal error: exception Sarek_parse_helpers.Parse_error_exn(...)
   ```

   No location, no message text, no file/line.

## Not a regression

This is pre-existing and independent of `when` guards — it reproduces with any construct `Sarek_parse` refuses inside a `[@sarek.module]` item (verified with a lambda, no guard involved). The `when`-guard refusal (backlog-191) merely made it easier to hit.

## Severity

Diagnostic quality only. Every path is **fail-closed** — the compilation fails in all three cases (case 1 via the unbound variable, case 2 via the fatal exception). There is no path that produces a silently wrong kernel. So this is a usability defect, not a correctness one.

## Fix sketch

Two changes, both local:

[...]

PR #395, opened 2026-07-30
fix(ppx): four diagnostics advertised an extension point that does not exist

mathiasbourgoin:
## The defect (backlog-193)

Four PPX diagnostics told the reader to `register it with [%%ktype]`. There has
never been a `ktype` extension in this tree — the name does not appear anywhere
except in those four strings. So the one actionable line of each message pointed
at an extension point the user cannot write. That is worse than no advice,
because it reads as authoritative.

**The filed count of four is correct**, for `ktype`. All four are in
`sarek/ppx/Sarek_ppx.ml`, two messages duplicated across the size and the
alignment lookup:

| Site | Function | Arm |
|---|---|---|
| `Sarek_ppx.ml:245` | `get_type_size_from_core_type` | named type not in `type_size_registry` |
| `Sarek_ppx.ml:252` | `get_type_size_from_core_type` | field type is not a `Ptyp_constr` |
| `Sarek_ppx.ml:273` | `get_type_align_from_core_type` | named type not in `type_align_registry` |
| `Sarek_ppx.ml:280` | `get_type_align_from_core_type` | field type is not a `Ptyp_constr` |

## Substituting the real name is not enough

Two things had to be measured rather than assumed, and each one would have
shipped a message that is still false.

**1. `[@@sarek.type]` is not, on its own, true advice.** The registries are
populated by `register_sarek_type_decl`, reached from the `[@@sarek.type]` and
`[@@sarek.type_private]` attribute rules. But `register_sarek_type_decl` raises
on anything that is not a record or a variant, so an **alias cannot carry it**:

```
type my_meters = float [@@sarek.type]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Only record or variant types can be used with [@@sarek.type]
```

An alias of a scalar is exactly the shape this repository's own negative test
(`test_unregistered_field.ml`) uses — so "declare it with `[@@sarek.type]`" would
have been false advice for the very case the suite already exercises. The fix
there is to write the scalar. The message is two-branched accordingly, and both
branches were compiled to confirm they resolve the error. `[@@sarek.type_private]`
was measured too and is **not** offered: it registers the size and alignment, then
the enclosing type fails later on a missing `<name>_custom` value.

**2. `Location.raise_errorf` is Format-based.** A literal `@@` in a format string
prints as a single `@`. So the corrected message reached the terminal as
`[@sarek.type]` — a single-`@` attribute, which cannot sit on a type declaration,
so the advice still failed. **Fourteen render sites tree-wide were in this state**
(the four above, plus ten pre-existing in `Sarek_ppx.ml`, `Sarek_error.ml` and
`Sarek_lower_ir.ml`). Leaving ten known-false messages beside four fixed ones is
how a fifth appears.

An earlier revision of this section said *sixteen*, and cited
`[%%kernel.real64]` → `[%kernel.real64]` as one of the catches. Both were wrong,
and the review pass measured them: 16 is the **post-fix escape count** — two of
those escapes were introduced by the corrected two-branched messages themselves,
so it cannot also be the number of pre-existing defects. And `kernel.real64` is
declared with `Extension.Context.expression`, so **one** `%` is correct: escaping
it made the message print `[%%kernel.real64]`, a structure-item extension that
does not exist. That change broke a message that had been right, and is reverted.

The durable lesson is in the gate, which now derives the required sigil count
from each construct's **ppxlib declaration context** rather than applying a
blanket doubling rule, and compares the rendered count in **both** directions.
Escaping too much is the same defect as escaping too little. (Note also that
`@@@` renders as `@@`, so this was never a strict doubling relationship.)

## Evidence the new advice works

Each branch was compiled against the negative-profile library. Kernels and
outcomes:

| # | Source | Result |
|---|---|---|
| A | `type my_meters = float [@@sarek.type]` | **rejected** — aliases cannot be registered |
| B | `type inner = {a : int32; b : float} [@@sarek.type]` then `{distance : inner; …}` | **compiles** — the record branch of the advice |
| C | `{distance : float; count : int32}` | **compiles** — the alias branch of the advice |
| D | `[@@sarek.type_private]` on the nested record | **rejected** later (`Unbound value "inner_custom"`) — so not advised |
| E | `type tag = A \| B of int32 [@@sarek.type]` as a field | **compiles** — the variant branch |
| F | `{pair : int32 * int32; …}` | **rejected** by the second arm |
| G | `type v = A \| B of my_meters [@@sarek.type]` | **rejected** — the alias case through a variant payload |

Rendered output after the fix:

```
Error: sarek: unknown alignment for field type 'my_meters' - if 'my_meters' is a
record or a variant, declare it with [@@sarek.type]; if it is an alias, use the
aliased type directly (an alias cannot carry [@@sarek.type]). A silent default
would corrupt the host/device layout.

Error: sarek: unsupported field type in a GPU record/variant - only scalar types
and record/variant types declared with [@@sarek.type] are supported here
```

No backend list, no "supported everywhere": the second message names exactly the
two things measured to work (scalars, and `[@@sarek.type]` records/variants).

## Pinning it

What let four false messages survive is that **nothing read message text**. Two
guards, because the defect used two mechanisms and closing one would not have
caught the other.

**`scripts/check-ppx-construct-names.sh`** (static, in CI). A construct named in
a diagnostic must (1) appear in the PPX's own name table — built from the string
literals under `sarek/ppx/`, since that is how ppxlib is told about a construct
(`Attribute.declare "sarek.type"`, `has_attr "sarek.module"`, …), not from a
hand-kept list that would rot — and (2) survive Format rendering. It exits **2**
when that table is empty, because a check that cannot decide what a legal
construct is has not cleared anything. Comments are stripped first (prose about a
construct is not advice to use it) and `sarek/tests/` is excluded, because a
negative test's job is to name the spelling that must be *refused* —
`test_sarek_type_extension.ml` deliberately writes `[%%sarek.type]`.

A first draft attributed a format string to a Format call by line window; a
`Printf.sprintf` two lines below a `Location.raise_errorf` inherited the wrong
rule and the positive control caught it. It now attributes each literal to the
**nearest preceding call head**, and the non-Format set is listed explicitly
because it is what *stops* attribution — anything missing from it can only cause a
false positive, which is the failure mode that gets a gate deleted.

`scripts/check-ppx-construct-names.test.sh` — 7 cases in throwaway git repos.
Both defect shapes, plus the four ways this gate could read green while checking
nothing: a comment, a printf conversion (`[%s]`), an empty name table, and the
excluded test tree. Case 0 is the positive control.

**`sarek/tests/negative/test_ktype_advice.ml`** (executed). The static gate cannot
see rendered bytes, so `make test_negative` greps the compiler's real output for
`[@@sarek.type]` with **both** `@` signs. `test_unregistered_field.ml` already
covered the refusal and asserted only on `unknown (size|alignment) for field
type` — which is why the false advice sat inside a string the suite was already
reading with no assertion touching it. The new case is **self-contained**
(backlog-208: eight siblings carry an `open` their dune stanza does not provide
and pass only because the PPX refuses the file before typechecking reaches it).

## Prove-red

Committed first, then reverted with `git checkout <ref> -- <path>`. Mutation
landing was asserted before either polarity was trusted.

| Mutation | Landed | Guard | Exit |
|---|---|---|---|
| PPX sources ← `origin/main` (pre-fix text) | 3 files, 4 `%%ktype` present | gate | **1** — names all 4 `ktype` sites *and* all 14 render sites |
| `Sarek_ppx.ml` ← `origin/main` | 1 file | `make test_negative` | **2** — compiler said `[%ktype]`, assertion refused |
| un-escape sigils only (name right, rendering wrong) | 10 sigils | gate **1**, `make test_negative` **2** | compiler said `[@sarek.type]`, both refused |

Baseline (unmutated) is green for all three — see the gate table below.

## Gates

Each exit code captured directly, never `$?` after a pipe. Tree verified clean
with `git status --porcelain` before and after.

```
dune build                          = 0
dune runtest --force                = 0
dune build @fmt                     = 0
scripts/check-license-headers.sh    = 0
scripts/check-kb-properties.sh      = 0
scripts/prove-red.sh                = 0
scripts/check-opam-clean.sh          = 0
scripts/check-negative-case-coverage.sh   = 0
scripts/check-ppx-construct-names.sh      = 0
scripts/check-ppx-construct-names.test.sh = 0
make test_negative                  = 0
```

`kb/properties.md` gains one `gate-red-path` row (`KB-GATE-PPX-CONSTRUCT-NAMES`),
appended after the last gate row; `ci.yml` gains one gate step next to the other
host-side lexical gates and one red-path invocation in `Harness self-tests`.
`check-kb-properties.sh` requires both and passes.

## Not verified

- The `unknown size for field type` arm looks **unreachable** for an unregistered
  named type: alignment is computed before size on both the record
  (`aligned_record_offsets`) and the variant (`variant_payload_offset`) path, so
  the alignment message always preempts it. Fixed anyway; reachability not
  established either way.
- **Cross-compilation-unit registration.** The registries are PPX-process
  hashtables, so a `[@@sarek.type]` record in a *linked library* is likely not
  visible to a kernel in another unit. Not measured — and so the message makes
  **no locality claim** at all. A user in that situation gets a true but
  incomplete message rather than a confident wrong one.
- The gate's Format-attribution is a heuristic, not a typechecker. It can miss a
  rendering bug at an unusual call site (false negative); it is built so that it
  does not invent one (false positive).


mathiasbourgoin:
@coderabbitai review

CI is green on this head (4/4 jobs). Requesting a review so the PR can satisfy the repository's review requirement.

Two things worth attacking specifically, because both are claims about rendered behaviour rather than about source:

1. `Location.raise_errorf` is Format-based, so `@` is a formatting directive and a literal `[@@sarek.type]` in the format string reaches the terminal as `[@sarek.type]` — a single-`@` attribute that cannot sit on a type declaration. This PR escapes 16 such sites. Please check the escaping is right at every one, and that no message now over- or under-escapes.

2. The advice itself is two-branched on purpose: `[@@sarek.type]` does not apply to a type alias (`register_sarek_type_decl` refuses anything that is not a record or variant), which is exactly the shape `test_unregistered_field.ml` uses. Please check the branch conditions match what the PPX actually accepts.

The new gate is a heuristic over call sites, not a typechecker — it can miss a rendering bug at an unusual call site. It is built not to invent one, but the boundary is worth a look.

mathiasbourgoin:
## Independent adversarial review — rebased onto `origin/main`, two defects fixed on the branch

Verdict: **GO**, after `1c5b5992`. The diagnosis in this PR is right and the mechanism claim is exactly right. But the *rule* it derived from that mechanism — "a doubled sigil in a Format string is a bug" — is wider than the truth, and it failed in both directions at once: it missed a site of the same shape, and it broke a message that was already correct. The new gate could not see either, for two separate reasons.

### The rebase

`a1827b75` sat on `93b865f5`; `origin/main` had moved four commits, and three files were shared. Rebased with `rerere` explicitly disabled (`git -c rerere.enabled=false rebase origin/main`) so nothing was auto-resolved from another branch's cache; it applied without conflict. Verified against both parents rather than trusted:

```
git diff --stat origin/main HEAD -- scripts/prove-red.sh scripts/prove-red.test.sh \
    kb/sarek/ppx/parser.md sarek/ppx/Sarek_parse.ml sarek/tests/negative/test_when_guard.ml
  -> empty            (main's #390/#393 side intact)
git diff --stat a1827b75 HEAD -- sarek/ppx/Sarek_{error,lower_ir,ppx}.ml \
    sarek/tests/negative/test_ktype_advice.ml scripts/check-ppx-construct-names*.sh .github/workflows/ci.yml
  -> empty            (this PR's side intact)
```

Both `neg_test_when_guard` and `neg_test_ktype_advice` stanzas present, both `test_negative` assertions present, `kb/properties.md` carries both the prove-red edits and `KB-GATE-PPX-CONSTRUCT-NAMES`.

---

### Claim 1 — Format collapses `@@`, four are needed: **verified**, from first principles

Not taken from the diff. A standalone program, no Sarek involved:

```
A Printf   "[@@sarek.type]"     -> [@@sarek.type]
B Format   "[@@sarek.type]"     -> [@sarek.type]
C Format   "[@@@@sarek.type]"   -> [@@sarek.type]
D Format   "[@@@sarek.type]"    -> [@@sarek.type]     (three also give two)
E asprintf "[@@sarek.type]"     -> [@sarek.type]
G Format   "[%%%%kernel.real64]"-> [%%kernel.real64]
```

Confirmed, `Printf` unaffected, and `%%` behaves the same way as `@@`. Row D is worth keeping in mind: three sigils render two, so the escape is not a strict doubling.

### Claim 2 — "sixteen sites tree-wide": **refuted, twice over**

Measured by running this PR's own gate against the pre-sweep sources (`git checkout origin/main -- sarek/ppx/Sarek_{ppx,error,lower_ir}.ml`): **4 name violations and 14 render violations**, the four `%%ktype` ones included in the 14. Sixteen is the count of `@@@@`/`%%%%` escapes *after* the fix — two of which the new two-branched messages introduced and which therefore never existed in the defective state. So "the four above, plus twelve pre-existing" is not the pre-fix population.

The `kb/properties.md` row was worse than the PR body: *"…sixteen **further** sites named `[@@sarek.type]`…"* reads as twenty, and labels the four `%%ktype` sites as `@@sarek.type` ones. Corrected to the measured fourteen in `1c5b5992`, along with the same count in the `ci.yml` step comment.

And the sweep was **incomplete**. `sarek/ppx/Sarek_ppx.ml:2515`, `expand_sarek_include`'s payload refusal, is a `Location.raise_errorf` whose literal was `"…e.g. [%%sarek_include \"file.ml\"]"`. Executed, negative profile, three probe libraries:

```
let%sarek_include _ = 42            -> Error: %sarek_include expects a string path,
                                             e.g. [%sarek_include "file.ml"]
[%sarek_include  "probe_target.ml"] -> Error: Uninterpreted extension 'sarek_include'   (exit 1)
[%%sarek_include "probe_target.ml"] -> compiles                                          (exit 0)
```

`sarek_include` is declared in `Extension.Context.structure_item`. The single-`%` spelling the message printed is not that extension — it parses as a structure-level expression and comes back uninterpreted. Same defect, one function over, still shipping. The odoc comment above `expand_sarek_include` said `[%sarek_include "path/to/file.ml"]` too; both corrected.

### Claim 2, opposite polarity — the fix **broke a correct message**

The PR cites `[%%kernel.real64]` → `[%kernel.real64]` as one of the sixteen catches. It was not a defect. `kernel.real64` is declared in `Extension.Context.expression`:

```
sarek/ppx/Sarek_ppx.ml:2285   Extension.V3.declare "kernel.real64" Extension.Context.expression
sarek/tests/e2e/test_real64_single_source.ml:51   [%kernel.real64      <- one '%', in a passing test
```

An expression extension is written with **one** sigil, so `"[%%kernel.real64]"` in a format string was already rendering the right thing. Escaping it to `%%%%` made it print `[%%kernel.real64]` — a structure-item extension that does not exist. Reverted.

How many sigils a construct takes is a property of the **declaration context**, not of the construct kind, and no flat rule gets both `[@@sarek.type]` and `[%kernel.real64]` right. Demonstrated by mutation, on the corrected tree: forcing the flat rule back (`want = (sigil[0], 2)`) makes the gate exit 1 with **six** violations on code that is correct, `[%native …]` in `Sarek_ast.ml`, `Sarek_typed_ast.ml`, `Sarek_parse.ml`, `Sarek_ir_conv.ml` and `[%kernel.real64]` itself. That is the shape that gets a gate deleted along with the real fix.

### Claim 3 — two-branched advice because an alias cannot carry `[@@sarek.type]`: **verified**

`register_sarek_type_decl` raises on anything that is not a record or a variant, and `test_unregistered_field.ml` is exactly `type unregistered_alias = float`. Both branches present in both messages, `type_name` passed twice. No objection.

### Claim 4 — the new gate: **half-covering as described, and it had two blind spots**

The PR is honest that the static half cannot see rendered bytes. It is right that it reasons *about* rendering rather than observing it, and I verified that reasoning is sound (claim 1 above). Two things it did not see:

**(a) A construct named WITH A PAYLOAD was invisible to both halves.** `CONSTRUCT` required `]` immediately after the name, so every construct that takes an argument fell out — which is most of them, and is precisely the one real site the sweep missed. Proven, not argued: restore the `]`-anchored regex, re-introduce the `sarek_include` defect, and

```
scripts/check-ppx-construct-names.sh   -> exit 0
  "OK -- 436 tracked source(s), 464 bracket-bearing literal(s); every named
   construct is declared under sarek/ppx/ and survives Format rendering"
```

A green pass over a live instance of the defect the gate exists for. With the lookahead (`(?=\]|\s|$)`) the same tree gives exit 1 naming `Sarek_ppx.ml:2537`. Relaxing it introduces **zero** name-half false positives across all 436 tracked sources.

**(b) A long comment switched the render half off.** `FMT_REACH` was a raw byte distance from the call head to the literal, measured over a *comment-blanked* source (comments become spaces, offsets preserved). So a long enough comment between `Location.raise_errorf` and its own format string pushes the literal out of reach, `in_fmt` comes back false, and the render check silently says nothing. Found by writing one: my first explanatory comment above the corrected `sarek_include` message was ~500 bytes and the gate reported the freshly-corrected line as a violation *because it had stopped treating it as a format string*. Now counted in non-blank bytes, `FMT_REACH = 200`, and pinned as harness case 12 — which fails when the byte version is restored.

**Gate reworked accordingly.** The required spelling is derived from the ppxlib declaration site (`Attribute.declare` / `Extension.declare` + `…Context.<ctx>`) and the **rendered** sigil count is compared against it, in both directions. Exit 2 extended to "no declaration site parsed" and "declared in a context this table cannot classify", on the argument the empty name table already carried. A construct matched by hand on the AST (`native`, `sarek.module`) has no parseable context; the render half stays silent there and the header says so — a stated false negative, not a guess.

Red path grows 7 → **13**: payload-bearing name (8) and render (7), the over-escape polarity (9), the *correct* single-sigil Format spelling that must NOT fire (10), the unclassifiable context (11), the long comment (12). Each mutation asserted to have landed before either polarity was read — one `perl -0pi` attempt changed nothing and read as green, which is why `git diff --numstat` gates every step below.

### Claim 5 — the negative test greps real stderr: **verified, and not vacuous**

`test_ktype_advice.ml` is genuinely self-contained (no `open`, no library value), so a guard that stops firing produces a clean compile and the Makefile assertion is what fails — the backlog-208 property, correctly applied. The advice assertion requires both `@` signs and both branches.

Added the same for the site I found: `sarek/tests/negative/test_sarek_include_advice.ml`, `let%sarek_include _ = 42`, self-contained, with the Makefile grepping real stderr for `[%%sarek_include]` with **both** percent signs.

### Prove-red

Every mutation asserted landed (`git diff --numstat` non-empty) before any exit code was read.

| Mutation | Landed | Guard | Exit |
|---|---|---|---|
| `sarek_include` message → pre-review single-`%` | 2/2 | gate | **1** — names `Sarek_ppx.ml:2537`, renders `[%sarek_include ...]` |
| same | 2/2 | `make test_negative` | **2** — compiler said `[%sarek_include "file.ml"]`, assertion refused |
| `kernel.real64` → `%%%%` (the PR's over-escape) | 1/1 | gate | **1** — "reaches the user as `[%%kernel.real64]`, but the construct is written `[%kernel.real64]`" |
| `]`-anchored `CONSTRUCT` + `sarek_include` defect live | 2/2 + 1/1 | gate | **0** ← the blind spot; red path **1**, 5 cases fail |
| `FMT_REACH` → raw bytes | 1/3 | red path | **1** — case 12 only |
| flat `want = (sigil, 2)` rule, correct tree | 1/2 | gate | **1** — 6 false positives incl. `[%kernel.real64]` |

### Gates, exit codes captured directly (never `$?` after a pipe)

| | pre-review `9d14bfa6` | after `1c5b5992` |
|---|---|---|
| `dune build` | 0 | 0 |
| `dune runtest --force` | 0 (5122 lines) | 0 (5122 lines) |
| `dune build @fmt` | 0 | 0 |
| `scripts/prove-red.sh` | 0 (6 subjects, 33 mutations) | 0 (6 subjects, 33 mutations) |
| `scripts/check-kb-properties.sh` (+ `.test.sh`) | 0 / 0 | 0 / 0 |
| `scripts/check-ppx-construct-names.sh` (+ `.test.sh`) | 0 / 0 (7 cases) | 0 / 0 (13 cases) |
| `scripts/check-license-headers.sh` | 0 | 0 |
| `scripts/check-opam-clean.sh` | 0 | 0 |
| `scripts/check-cited-paths-exist.sh` | 0 | 0 |
| `scripts/check-negative-case-coverage.sh` | 0 | 0 |
| `make test_negative` | 0 | 0 |

`dune runtest --force` non-empty on both (5122 lines), so neither is a cached-alias no-op. Tree `git status --porcelain` empty before and after every mutation; `git checkout <ref> -- <path>` throughout, no `git stash`.

**Carrier wiring for the new gate**: `kb/properties.md` row `KB-GATE-PPX-CONSTRUCT-NAMES` present and `check-kb-properties.sh` passes; `ci.yml:294` invokes the red path under *Harness self-tests* and `ci.yml:532` invokes the gate. Both confirmed. The new negative case is carried by `make test_negative`, which CI runs, and by `check-negative-case-coverage.sh`.

### Issue #391

Not touched and not claimed. No message under `sarek/ppx/` names `[@sarek.module]` inside a Format string, and nothing here reaches the `Parse_error_exn` path in `Sarek_parse_helpers.ml`. The gap stays open.

### What I could NOT verify

- **Reachability of the `unknown size for field type` arm.** The PR says it looks unreachable because alignment is computed first. I did not settle it either way; both arms are fixed regardless.
- **Cross-compilation-unit registration.** Not measured here either. The messages correctly make no locality claim.
- **The render half over hand-matched constructs.** `native` and `sarek.module` are matched directly on the AST, so no declaration context is parseable and the gate says nothing about their sigil counts. A rendering bug in a message naming one of those would still ship. Stated in the script header rather than papered over.
- **The Format-attribution heuristic in general.** Nearest-preceding-call-head within 200 non-blank bytes is not a typechecker. It can miss a site at an unusual call shape. I checked it does not invent one: zero false positives across 436 tracked sources, 464 bracket-bearing literals.
- **The PR body and commit message of `9d14bfa6`** still say sixteen pre-existing sites and still cite `[%%kernel.real64]` as a catch. I corrected the durable artifacts (`kb/properties.md`, the script header, the `ci.yml` comments) but did not rewrite that commit.

Not merged.

PR #378, opened 2026-07-29
docs: reconcile the entry-point documents with the tree

mathiasbourgoin:
## Why

An audit of the 2026-07 window found **92 documentation files added, 7 modified** — and every document a newcomer or an auditor reads *first* stale. Three were stale in the direction of **understating what exists**, which is the worse direction: it makes real verification work invisible and invites shipped work to be redone.

This PR corrects the entry points. Documentation and comments only — **no `.ml`, `.mli`, `dune` or CI file is touched**, and no `*.opam` is committed.

Method, mirroring the `docs/optimization/` SoA handoff docs the reviewer named as the standard: retract stale premises **in place**, next to the claim being replaced; mark closed hazards as history with a note saying why leaving them imperative would reopen the work; and state gates as **observed**, citing the file that can be re-read, not as intentions.

## What was wrong, and what it says now

| File | The claim that was wrong | What it says now | Verified against |
|---|---|---|---|
| `README.md` | "No `ptxas` validation gate or GPU execution in CI yet" | Split in two, because only half was false. The `ptxas` gate exists and **cannot self-skip green**: one kernel per registered intrinsic name and width, enumerated from the emitter's own dispatch registry; the CI image ships CUDA 12.6; `assert-toolchain.sh` fails the job if `ptxas` is absent *or* cannot assemble a probe. GPU **execution** in CI is still absent, and that is now stated on its own. | `sarek/tests/unit/test_ptx_intrinsic_sweep.ml`, `sarek/tests/unit/test_ptx_snapshot.ml:2571`, `ci/assert-toolchain.sh:88-125` (`EXPECTED_CHECKS=13`), `ci/Dockerfile:165-176`, `.github/workflows/ci.yml` "Run unit tests" |
| `README.md` | "Warp-level primitives are modeled in the IR but not emitted by any backend yet" | `SWarpBarrier`/`SMemFence` are emitted by **all six** generators (`bar.warp.sync`, `__syncwarp()`, `sub_group_barrier`, `simdgroup_barrier`, `subgroupBarrier()`×2). The real gap is the **front end** — no PPX syntax constructs either statement, so they are reachable only from hand-built IR. Warp **shuffle**/vote/ballot are the separate, larger gap: not modelled, not emitted. | `sarek/tests/unit/test_sync_stmt_emission.ml:80-114` (per-backend expectation table) and its scope-note assertion at `:152-159` |
[...]

PR #396, opened 2026-07-30
feat(gate): the citation gate could not see markdown, and could not tell a present sha from a reachable one

mathiasbourgoin:
backlog-210 and backlog-204 are the same script and the same defect family, so they land together: **a citation that reads as evidence and resolves to nothing.**

## Markdown was entirely ungated (backlog-210)

The source list was `sources = ls("*.ml", "*.mli")`. Proved in isolation on the old script: a bad path present **only** in `README.md` — a broken markdown link plus a nonexistent backticked path, unmentioned in any `.ml` — exits **0**, under the message *"every repo-relative path cited in 1 tracked .ml/.mli files resolves to a tracked file"*. A true sentence with the wrong coverage, and it reads like whole-repo coverage.

That is why PR #378 was NO-GO'd **three times**, each for a false sentence in a markdown file: five sentences, then a miscount of synchronising intrinsics, then "the five .ml files Extract.v produces are committed" (they are not tracked, and `extraction/dune` says so). Three rounds of human and agent review caught them. Nothing mechanical did.

Markdown now has **its own extractors**, not the OCaml comment state machine:

| Form | Admitted / excluded |
|---|---|
| Inline link `[text](path)` | admitted; anchor and query **stripped**, never a reason to skip |
| Reference definition `[label]: path` | admitted |
| Backticked span `` `path` `` | admitted, filtered by the same path-shape + placeholder + PTX-mnemonic rules as an OCaml comment |
| Fenced code block (``` / ~~~) | **excluded** — a fence is where a document keeps the things that only *look* like citations: shell transcripts, `git log` output, generated ledger dumps, lockfile digests. That exclusion is precisely why inline spans *outside* fences are scanned instead of being lumped in with "code". |
| `https://…`, `mailto:`, bare `#fragment` | excluded |
| Site-absolute `/backends/` | excluded — a gh-pages Jekyll permalink addresses the published site, not a path in the tree |
| `foo.html` link | resolves via its markdown source (`foo.md`) — gh-pages is Jekyll, and all 13 such links are correct |
| directory target | resolves for a **link**; a backticked `kb/` is not a citation anybody follows |
| `/abs/path`, `~/…`, `$VAR/…` | excluded — not repo-relative citations |
| HTML `<a href>` | not extracted (stated gap) |

Resolution stays against `git ls-files`, never the filesystem. Sources: tracked `*.md` at the root and under `docs/ kb/ specs/ gh-pages/ formal/`. `formal/` is in although the task named the other four — `formal/type-safety/STATUS.md` carried one of the two symptoms of the sha defect below, so excluding it would have made the gate blind to half of the instance that motivated it. `benchmarks/`, `.github/` and the per-library READMEs are excluded, each for a stated reason in the script.

## A cited sha must be REACHABLE, not merely present (backlog-204)

Five instances in one day. `docs/plans/FORMAL-ROADMAP.md` cited `d72a2e6a`, `fbfb3656`, `f6c14c2a`; a fourth, `1da95861`, was missed by the human review that found the other three; the same false text also sat in `formal/type-safety/STATUS.md`, so one root with two symptoms. A fifth, `4b15a323`, was produced **by** the fix process — an agent cited a sha its own rebase had just orphaned.

The defect is not a missing object. All of them **exist** locally. They are unreachable: `git branch -r --contains` is empty (a reachable sibling is in 120+), so no fresh clone resolves them. An existence check alone passes all five. The gate requires **both** `git cat-file -e <sha>^{commit}` **and** a non-empty `git branch -r --contains <sha>`.

**Tags deliberately do not count.** `actions/checkout` passes `--no-tags` unless `fetch-tags: true`, so `git tag --contains` is non-empty on a workstation and empty in CI — counting them would make the verdict depend on which machine ran the gate, the same asymmetry as backlog-152. It also mattered concretely: `d72a2e6a` is contained in the pushed tag `archive/master` and in no branch, so a tag-counting version of this check would have called one of the four motivating instances fine.

`7..40` hex also excludes sha256 by construction (a 64-hex token fails the trailing boundary), so lockfile and artifact digests never enter. All-digit runs, hex inside a longer token (`deadbee.tar.gz`, `abc1234-rc1`), and `#a1b2c3d` are rejected.

## Fail-closed under a shallow clone — the load-bearing part

On a shallow or ref-less clone `git branch -r --contains` reports nothing for **every** sha, reachable or not. So such a checkout either fails the whole build, or — if someone "fixes" that by reading empty as fine — passes everything while examining nothing. That second outcome is the vacuous-green class this repo keeps closing: in backlog-152 a `pull_request` checkout had no `origin/main`, a gate read it as "no differences", and about twenty PRs merged green onto a main that was already red.

Both preconditions are **asserted**, and their absence is **exit 2** — never a pass, never a silent skip:

- `git rev-parse --is-shallow-repository` must be `false`;
- `git branch -r` must be non-empty.

Asserted only when at least one sha is actually cited, so a tree that cites none (and the per-case fixtures) is not forced to have a remote it does not need. Tested against a **real** `git clone --depth 1`, not a simulation, and against a fixture with history but no remote-tracking refs — the exact `actions/checkout` default-depth shape.

## Prove-red, against the real instances

```
$ ./scripts/check-cited-paths-exist.sh          # exemption rows for these two removed
docs/plans/FORMAL-ROADMAP.md:64: cites commit '1da95861': exists locally but is in ZERO remote branches
formal/type-safety/STATUS.md:8:  cites commit 'd72a2e6a': exists locally but is in ZERO remote branches
EXIT=1
```

The six verified-reachable shas stay green — added to a tracked probe document and re-run, the checked population rose from 87 to 93 with exit 0:

`18248b7e` `73083c1f` `c3f2578b` `21d67ec9` `6cbb0a21` `891cb2b4`

Adding `4b15a323` to the same document turned it red; breaking one correct link in it turned it red on the path half; both reverted.

The harness (`scripts/check-cited-paths-exist.test.sh`, 37 cases, exit 0) ends by running **itself** against a stub that always exits 0 and one that always exits 1, and requires both runs to fail — "all cases passed" is otherwise a claim a harness checking nothing would also print.

## Triage of the pre-existing backlog

First run: **104** findings (93 paths, 11 shas). The gate was not widened to make them pass.

**26 removed by resolution rules that do not weaken the gate** — `.html`→`.md` for Jekyll links (13), absolute/`~/` paths the regex re-read as repo-relative (9), `$VAR`-rooted command paths (1), plus duplicates of those.

**22 documents corrected** (`README.md`, `docs/design/f16-dsl-element-type.md`, `docs/plans/ptx-spike-findings.md`, `docs/formal/rocq-value-ledger.md`, `docs/website-audit-2026-06-02.md`, `specs/ptx-dshared-formal.md`, `kb/sarek/runtime/{execution,interpreter}.md`, `kb/index.md`, `kb/support/scripts-ci.md`, `specs/ptx-records-variants.md`, `formal/convergence-safety/PLAN.md`, `formal/type-safety/findings/FINDINGS.md`, `kb/sarek/ppx/README.md`) — missing `sarek/` and `formal/codegen-ptx/` prefixes, `docs/faq.md` → `gh-pages/docs/faq.md`, two hypothetical test paths replaced by the real `sarek/sarek/test/…`, one workflow annotated as deleted per the KB dead-entry policy, and two `briefs/` pointers reworded (that tree is deliberately untracked, so no clone has it — the original defect this gate was built for).

**Four orphaned shas replaced by their reachable equivalents on `main`**, matched by subject and date:

| cited | reachable replacement |
|---|---|
| `20b79b36` | `c232a566` *fix(typer): F-TS-01 — restore ELet body env…* |
| `b37411b2` | `e5d16ab2` *fix(ppx): close file descriptor on malformed-file scan path* |
| `af9f6b81…` | `5d850248` *feat(convergence-safety): T3-S3 — trace silence…* |
| `fbfb3656` (PLAN.md tick log) | annotated in place, with the reachable `21d67ec9` beside it |

**The remainder are exemption rows, each with a checkable reason**, in three clusters: the formal-apparatus skill's own doc tree (its policy documents cite each other relative to the skill root, which is not vendored here); kb's **dead-entry records**, which name a deleted path *in order to record that it is gone* — the opposite of the drift this gate catches, but indistinguishable by shape, so each row names the commit that removed it; and named-but-absent-by-design cases (an optional harness file, a file the reader is being told to create, a table measuring hand-written ledgers as they stood on a date). Two rows exempt `.vo` **coqchk digests** that are not commits at all.

The exemption channel now takes `citing/file.md::<token>` to scope a row to one file. That form exists because four documents quote an unreachable sha in order to **say** it is unreachable; a bare row would also wave that sha through in a document that cites it as evidence again. The harness pins both directions — scoped-here is honoured, scoped-elsewhere is not.

## Still unscanned (stated, not overlooked)

`benchmarks/`, `.github/` and the per-library READMEs; markdown inside fenced blocks; HTML `<a href>`; shas in `.ml` comments (all five instances were markdown). These are the obvious follow-up and are named in the script and in the KB row.

## Gates

`dune build` 0 · `dune runtest --force` 0 · `dune build @fmt` 0 · `check-license-headers.sh` 0 · `check-kb-properties.sh` 0 (40 declarations) · `prove-red.sh` 0 (6 subjects, 30 mutations) · `check-opam-clean.sh` 0 · `check-cited-paths-exist.sh` 0 · `check-cited-paths-exist.test.sh` 0 (37 cases) · `check-workflow-steps.sh` 0 · `check-readme-repo-links.sh` 0

`KB-GATE-CITED-PATHS` was **updated**, not duplicated — its description named backlog-210 as a known limitation, which is no longer true.


mathiasbourgoin:
## Independent adversarial review — PR #396

Rebased onto `origin/main` (`3bdf2ae2`, picking up #390 and #393) and pushed; reviewed at `13bce28d`. Clean rebase, no rerere resolution involved — verified by `git diff b827f227 HEAD --stat`, which shows only files `main` moved, and the `kb/properties.md` collision (both #393 and this PR edit the `code-intel` block) merged with both edits intact.

**Verdict: GO**, after two fixes committed on the branch. One of them is not a nit: the harness, on an unusable fixture, wrote to the repository under review.

---

### 1. The harness did not read green while checking nothing — it *committed*

`b827f227` fixed the CI symptom (`git -C` everywhere, branch asked for not assumed, shas in variables, builder output asserted). It did not fix the mechanism, and the mechanism is worse than a vacuous pass.

**`git -C "" ` does not fail.** An empty `-C` argument leaves git in the directory it was started from. Two blocks in `check-cited-paths-exist.test.sh` guarded their fixture by *counting* a failure and falling through — `[ ${#orphan} -eq 8 ] || { echo "FAIL: …"; fail=$((fail + 1)); }` with no `return`. Reached with an empty `$d` they then ran, against the reviewer's own checkout:

```
git -C "$d" add -A && git -C "$d" commit -m doc && git -C "$d" update-ref refs/remotes/origin/main HEAD
```

I hit this for real while mutation-testing. Emptying `br` inside `mkshafixture` (one of the values `b827f227` added) made the builder return 1, both guards fell through, and the harness:

* created **six junk commits** on the branch under review — one per run;
* rewrote **`refs/remotes/origin/main`** to point at one of them. That ref lives in the shared git dir `/home/mathias/dev/SPOC`, so every worktree of that clone was left believing `origin/main` was a garbage commit containing my scratch state. Measured: `git rev-parse origin/main` → `6f0fbc58 "doc"` against a real remote `main` of `3bdf2ae2`;
* **committed the mutated harness**, so `br=""` came back in `HEAD` and every subsequent run was inert for the same reason. Self-amplifying.

Repaired (`update-ref` back to `3bdf2ae2`, `reset --hard`) and fixed in `378a16f1`:

* `usable_fixture <repo> <sha>…` refuses an empty path, a directory without `.git`, and any sha that is not 8 hex chars. Both blocks now **abort**; counting a failure and continuing is what let the writes happen.
* The guard is pinned in both directions — accepts a real fixture, refuses each of the three rot shapes — so editing it to `return 0` fails rather than silently disarming every fixture-backed case.
* **`red: a sha exemption scoped to another file does not apply` asserted exit 1 and no message.** A `cd` into a dead fixture also exits 1 without ever running the subject, and I observed that case printing `PASS` in exactly that state. It now asserts `in ZERO remote branches` too. This was the last case in the file that could pass having run nothing.

Prove-red, both directions:

| mutation | before | after |
|---|---|---|
| `usable_fixture` → always accept | — | `EXIT=1`, 3 × `FAIL meta: usable_fixture ACCEPTED …` |
| `br=""` in `mkshafixture` | 6 junk commits + clobbered `origin/main`, later runs inert | `EXIT=1`, `9 FAILED`, `HEAD` **unchanged**, `origin/main` **unchanged** |

Harness: 37 → 41 cases, all passing.

**On the rest of the class, the `b827f227` fix holds.** I emptied every fixture-derived value in turn — `ok`, `orphan`, both, the whole triple, and forced `git clone` to fail in the shallow block. Each fails loudly and by name (`sha fixture is unusable (repo=… reachable=… orphan=…)`). `mkfixture`'s empty-path case is loud too, though by luck rather than design: bash's `cd ""` returns 1 (`cd: répertoire nul`), so `cd "$d" && bash "$SUBJECT"` short-circuits. `check()` still has no fixture assertion where `checksha()` does — the narrowing correction was applied to one polarity only. Not exploitable today; worth closing next time you are in the file.

### 2. `Thirteen such links exist` — measured, seventeen

Deleting the `.html` → `.md` rule and counting the findings: **17** link targets resolve only through it (16 unique file/token pairs), plus 4 that resolve as genuinely tracked `.html` (`../playground.html`, `../spoc_docs/index.html`) — 21 `.html` targets in total. All 17 do have their markdown source tracked beside them, so *"every one is correct"* survives; the number does not, and the audit behind it covered 13 of 17. Corrected in `378a16f1` at both sites (`check-cited-paths-exist.sh:353`, `.test.sh:297`).

While there: the rule is **not scoped to `gh-pages/`** though its stated rationale is (*"gh-pages is a Jekyll site"*). All 21 live instances are under `gh-pages/`, so scoping changes no verdict today; left global, with the cost now written down — in an unpublished tree a link to `x.html` beside a tracked `x.md` resolves here and 404s for the reader. Your call whether to tighten it.

### 3. The excluded-markdown list named three trees; there are five

`MD_ROOTS` leaves **89** tracked `.md` files unscanned. The comment named `benchmarks/`, `.github/` and "`sarek*/ spoc/` per-directory READMEs", which reads as exhaustive and is not:

* `schema/` (2) and `scripts/` (1) were not mentioned at all;
* the `sarek*/spoc/` set is not only READMEs — `sarek/sarek/BSP.md`, `sarek/sarek/FUSION.md`, `sarek-cuda/CHANGELOG.md`, `spoc/framework/Backend_error.md`;
* **`scripts/REVIEW-BUNDLE.md` is a gap, not a category.** `KB-GATE-INVENTORY` delegates its exemption reasoning to that file, so it carries live script paths and none of them is checked.

Fixed in `13bce28d` with measured counts (43/21/22/2/1 = 89).

---

### The exemption file: audited row by row, by mutation

I deleted each of the 50 rows in turn and re-ran the gate. **48 of 50 are load-bearing** — each removal produces exit 1 naming the token at a real line. Then I checked whether each *reason* is a true, checkable claim:

* **20 formal-apparatus rows** — every one resolves to a real citing site in `formal/convergence-safety/policy/*` or `formal/type-safety/*`, and each names a path that exists in the skill tree, not this repo. Spot-verified: `schema/kb-schema.md` is at `~/dev/agent-roster/schema/kb-schema.md` and `.claude/rules/escalation.md` at `SPOC/.claude/rules/escalation.md` (present, untracked). **Both of your self-reported UNVERIFIED items are confirmed, not merely inferred.**
* **10 kb dead-entry rows** — all 7 that name a deleting commit check out: `git show --name-status <sha> -- <path>` reports `D` for every pair (`ba375a36` ×3, `1abd73fb`, `d4275cb2`, `7ca31a76` ×2), and all 7 shas are themselves reachable (46–125 remote branches). `.github/workflows/retarget-prs.yml` absent with `RETARGET_PRS.md` tracked, as claimed. `kb/sarek/tests/common.md` did exist and was removed — though by `28e4b944` ("Untrack local harness/roster infra"), not the `#78` the reason names. Cosmetic.
* **9 named-but-absent rows** — read every citing line. Each really is a sentence asserting absence, an instruction to create, or a worked example: `kb/index.md:25` ("*correctly exits 0 with no `kb/spec.md`*"), `kb/glossary.md:154` ("*which is workstation-local*"), `dependencies.md:67` ("*Add `dependencies/THIRD_PARTY.md`…*"), `ADDING_BENCHMARKS.md:100` under "### 1. Create description file". The three `rocq-value-ledger.md` rows are a table measuring ledgers as of 2026-07-27; all three paths existed in history and were removed by `e162204e`.
* **7 sha rows** — every citing sentence does say the sha is unreachable (`STATUS.md:8`, `FORMAL-ROADMAP.md:39/64/76`, `PLAN.md:333`). `WORKFLOW_FEEDBACK.md:52` genuinely compares two `.vo` byte sequences, so `0fdf1107`/`41e0d70a` are digests, not commits.
* **4 replaced shas** are reachable *and* right: `c232a566`, `e5d16ab2`, `5d850248`, `21d67ec9` all sit in 119–123 remote branches and their subjects match the rows that cite them. No reachable-but-wrong substitution.

**Two rows are inert against the repo** (removing them leaves the gate at exit 0):

* `cuda/ptx-isa.md` — genuinely dead; nothing cites it any more.
* `typing/typecore.ml` — **not** dead, but load-bearing for the *harness*, not the tree. `check()` copies the real `cited-paths-exempt.tsv` into every fixture, so `green: an exempted path is not a finding` depends on a repo-level row. Measured: delete it and the harness fails (`40 passed, 1 FAILED`). A hidden coupling — whoever tidies that dead-looking row will break the red-path test. Worth a comment on the row, or a fixture-local tsv for that one case.

**Scoped exemptions do not leak.** `is_exempt` keys on `(citing_file, token)`; the harness pins both polarities (honoured in `docs/Doc.md`, refused when scoped to `docs/Other.md`) and, after my fix, the refusal half also asserts its message so it cannot pass on a dead fixture.

**On splitting the clusters: don't.** Both clusters are 100% verified true, each row is individually load-bearing, and every reason is a checkable statement I was able to settle in one command. Extracting ~30 rows into a follow-up would land the gate with a known backlog of failures — precisely the outcome the `KB-GATE-CITED-PATHS` description says gets a gate disabled. The two things I would extract are the two *inert* rows, and only for tidiness.

### Prove red on the real gate

Each defect introduced into a real tracked file, one at a time, exit code captured directly:

| injected | exit | message |
|---|---|---|
| broken inline link in `FORMAL-ROADMAP.md` | 1 | `cites 'docs/nope-does-not-exist.md', which is not a tracked file` |
| broken reference definition `[lbl]: docs/also-nope.md` | 1 | `cites 'docs/also-nope.md'…` |
| broken backticked path | 1 | `cites 'sarek/ppx/Ghost_module.ml'…` |
| real orphan sha `4b15a323` in `kb/index.md` | 1 | `exists locally but is in ZERO remote branches` |
| nonexistent sha `0000000abcdef1` | 1 | `no such commit in this repository` |
| exemption row with no TAB reason | 2 | `every exemption needs a TAB-separated reason` |
| **real `git clone --depth 1` of this repo** | 2 | `this is a SHALLOW clone…` |
| **real clone with `refs/remotes/*` deleted** | 2 | `no remote-tracking branches…` |

Claim 6 confirmed by the same method: deleting the exemption rows makes `1da95861` (`FORMAL-ROADMAP.md:64`) and `d72a2e6a` (`STATUS.md:8`) red at exit 1, and the six reachable shas stay green. Checked population is 87 with the rows, 93 without.

### Claims 1–5, 7 — verified

Tags do not count (claim 5) is real and load-bearing: `d72a2e6a` is in the pushed tag `archive/master` and **zero** branches, so a tag-counting version would have waved one of the four motivating instances through. `git branch -r` and `git rev-parse --is-shallow-repository` are asserted lazily, only when `sha_cites` is non-empty. `ci.yml`'s new comment claims the step depends on `fetch-depth: 0` — verified: the `build` job checks out at `fetch-depth: 0` (line 56) and the step is at line 523 inside it. `KB-GATE-CITED-PATHS` is **updated, not duplicated** (`grep -c '"id": "KB-GATE-CITED-PATHS"'` = 1), and both the tool and its red path are invoked from `ci.yml` (lines 523 and 224).

### The fork question: refuted, with a caveat

I built it rather than reasoned about it — upstream repo, a fork with its own branch, the head published as `refs/pull/1/*` only, then a checkout clone fetching `+refs/heads/*:refs/remotes/origin/*` plus the PR ref the way `actions/checkout` does (`ci.yml:42` already documents `refs/pull/N/merge` into a detached HEAD). Result: `git branch -r --contains <fork-sha>` → `pull/1/head pull/1/merge`. **Non-empty. No false RED, contributor PRs are not broken.**

The caveat runs the other way: because the PR ref lands under `refs/remotes/`, a sha from the PR's *own* branch reads as reachable *while the PR is open*, and goes red on `main` after a rebase-merge orphans it. That is exactly the `4b15a323` shape the description cites. Design-inherent, not a defect in this PR, but the gate is more permissive on a PR than on `main`.

### Smaller observations, no action taken

* **The `code-intel` fence is invisible to the gate.** `kb/properties.md`'s declaration block is a ```` ```code-intel ```` fence, so `md_prose` blanks it — including `4b15a323`, which is in **zero** remote branches and cited right there in the new description. I measured the whole cost of the fence exclusion: 6 unresolved path citations and 6 unreachable sha candidates across all 188 sources, and every one of them is a specimen (worked-example filenames, the gate's own `citing/file.md` / `fixtures/missing.md` illustrations, and the five motivating shas named in order to say they are unreachable). The exclusion is well calibrated; the *stated reason* — "shell transcripts, `git log` output, ledger dumps" — just does not describe the largest fence in this tree.
* **The markdown half does not dedup per (file, token) where the OCaml half does.** `specs/ptx-dshared-formal.md:60` cites `theories/PtxKernelSpec.v` twice on one line and is reported twice, so `N citation(s) point at a path no reader can open` counts occurrences on one side and citations on the other. The sha half dedups via `seen`.
* **PR body: "22 documents fixed."** The diff changes 15 `.md` files, and the two carrying the motivating defect — `FORMAL-ROADMAP.md` and `formal/type-safety/STATUS.md` — were fixed earlier on `main` (`5c79b5a6`, `642490c3`), not here. The narrative spans work outside this PR.

### Gates — every exit code captured directly

| gate | exit |
|---|---|
| `dune build` | 0 |
| `dune build @fmt` | 0 |
| `scripts/check-cited-paths-exist.sh` | 0 (647 `.ml/.mli`, 188 `.md`, 87 shas) |
| `scripts/check-cited-paths-exist.test.sh` | 0 (**41** cases) |
| `scripts/prove-red.sh` | 0 (6 subjects, 33 mutations) |
| `scripts/check-kb-properties.sh` | 0 (40 declarations) |
| `scripts/check-kb-properties.test.sh` | 0 (19 cases) |
| `scripts/check-opam-clean.sh` | 0 |
| `scripts/check-readme-repo-links.sh` | 0 |

### What I could not verify

* The triage arithmetic — 104 findings, 26 removed by resolution rules — is not reconstructible from the branch; the intermediate state is gone. I verified the *outcomes* (48/50 rows load-bearing, every reason true, 4 replacement shas correct) rather than the counts.
* The fork behaviour is verified against a **simulated** `actions/checkout` refspec, not an actual fork PR. The refspec matches what `ci.yml:42` documents, but only a real fork PR settles it.
* `dune build @fmt` passes but emits pre-existing odoc markup warnings in `test_ptx_snapshot.ml` and two others — untouched by this PR, not investigated.

PR #397, opened 2026-07-30
fix(codegen): order a record against a variant — one interleaved declaration pass per backend (backlog-211)

mathiasbourgoin:
# backlog-211 — a record/variant dependency edge crossed two loops nobody sorted

`#394` added `sort_record_types_by_dependency` and wired it into `gen_record_typedefs`
(C family) and into `Sarek_ir_glsl` / `Sarek_ir_wgsl`. It sorts each family's records
**inside that family's own emission loop**, and the families disagree on which of their
two loops runs first, so a dependency edge that CROSSES between a record declaration
and a variant declaration was ordered by neither.

`#394` documented both halves as still live and left one open question: whether static
tag erasure makes them unreachable. **It does not, and that is measured below** — but
reaching either half needs a RUNTIME-SELECTED constructor, and getting that wrong is
exactly how this reads as fixed when it is not (see *The trap*).

## Both halves, before and after

Host: RX 7900 XTX + Ryzen 7950X. Two OpenCL devices, two Vulkan devices, plus
Interpreter ×2 and Native. Reproducer: `sarek/tests/e2e/test_record_variant_decl_order.ml`,
two kernels from ordinary `[@@sarek.type]` source.

### Shape A — a variant with a RECORD payload (`Nowhere | At of probe_pt`)

| Device | Before | After |
| --- | --- | --- |
| OpenCL / AMD Radeon RX 7900 XTX (radeonsi, navi31, ACO) | **FAILED** | PASSED |
| OpenCL / AMD Ryzen 9 7950X (radeonsi, raphael_mendocino) | **FAILED** | PASSED |
| Vulkan / AMD Radeon RX 7900 XTX (RADV NAVI31) | PASSED | PASSED |
| Vulkan / AMD Ryzen 9 7950X (RADV RAPHAEL_MENDOCINO) | PASSED | PASSED |
| Interpreter ×2, Native | PASSED | PASSED |

Before, on both OpenCL devices:

```
input.cl:5:5: error: unknown type name 'Test_record_variant_decl_order_probe_pt'
    5 |     Test_record_variant_decl_order_probe_pt At_v;
      |     ^
input.cl:15:97: error: unknown type name 'Test_record_variant_decl_order_probe_pt';
                did you mean 'Test_record_variant_decl_order_probe'?
input.cl:31:102: error: passing '__private Test_record_variant_decl_order_probe_pt' to
                 parameter of incompatible type 'Test_record_variant_decl_order_probe'
Error executing LLVM compilation action.
```

### Shape B — a record with a VARIANT-typed field (`{gk : flagv; gv : float32}`)

| Device | Before | After |
| --- | --- | --- |
| Vulkan / AMD Radeon RX 7900 XTX (RADV NAVI31) | **FAILED** | PASSED |
| Vulkan / AMD Ryzen 9 7950X (RADV RAPHAEL_MENDOCINO) | **FAILED** | PASSED |
| OpenCL / AMD Radeon RX 7900 XTX (radeonsi, navi31, ACO) | PASSED | PASSED |
| OpenCL / AMD Ryzen 9 7950X (radeonsi, raphael_mendocino) | PASSED | PASSED |
| Interpreter ×2, Native | PASSED | PASSED |

Before, on both Vulkan devices:

```
[Vulkan Runtime] Compilation failed for:
Compiler log:
glslangValidator failed:
ERROR: /tmp/sarek_XXXXXX.comp:7: '' :  syntax error, unexpected IDENTIFIER
ERROR: 1 compilation errors.  No code generated.
ERROR: Linking compute stage: Missing entry point: Each stage requires one entry point
SPIR-V is not generated for failed compile or link
```

**Each family is red on exactly the shape the other family is green on.** That is why
either half read as "ordering is fixed" on its own, and why a single-family reproducer
cannot be trusted here.

After: `ALL PASSED on 4 declaration-emitting device(s)`, 14/14 cells.

### What was NOT measured on hardware

There is no NVIDIA, HIP or Metal hardware on this host and no WGSL device at all.
**No device result is claimed for CUDA, HIP, Metal or WGSL.** Those four get
source-level coverage only, from the device-independent sweep below, and the emitted
source for all five generators is asserted there with zero devices present.

## The trap: tag erasure, and why the first reproducer was green

The first version of this reproducer used literal constructors and passed on every
device while covering nothing. Static tag erasure (`Sarek_tag_erasure`, L14 S1/S2)
reduces a variant-typed local or record field written by a *literal* constructor
application: the variant never reaches `kern_variants`, no variant struct is emitted,
and the cross edge does not exist in the generated source. Confirmed by dumping
`kern_types`/`kern_variants`: `kern_variants=[]` for both shapes, with shape B's record
collapsed to the synthesized `_erec_float32_float32`.

Both kernels therefore select their constructor behind an `if`. Both halves are
reachable from ordinary `[@@sarek.type]` source; that is the answer to `#394`'s open
question.

## The fix — one interleaved pass per backend

`Sarek_ir_codegen.gen_type_decls` takes both declaration lists and emits from a single
topological pass over them together, dispatching each entry to `emit_record` or
`emit_variant`. All five generators that declare struct types go through it
(`gen_c_type_decls` for CUDA/OpenCL/Metal, `gen_type_decls` directly for GLSL/WGSL).
PTX declares no types and is untouched; HIP has no generator of its own and rides the
CUDA row.

- `referenced_type_names` (was `referenced_record_names`) now reports **variant** names
  as well as record names, so an edge is found whichever kind sits at either end.
- **Node identity is the POSITION in the list, not the type name.** This is
  load-bearing: a name-keyed self-edge drop reads a record's edge to a *same-named*
  variant as a self-reference and discards it — the exact edge class this PR is about.
  `mangle_name` maps `M.t` and `M_t` to one string and fusion concatenates two kernels'
  type lists without deduplicating, so same-named declarations are reachable.
- The visited set stays keyed on **physical identity** (`==`), because a cyclic
  `elttype` VALUE is legal: `let rec t = TVec t` has neither a record nor a variant on
  the cycle, so no name-keyed guard closes it. Preserved and pinned.
- `~tie_break` (`Variants_first` / `Records_first`) reproduces the order each family's
  two loops already produced, so an edge-free kernel's emitted source is byte-identical.
  `dune runtest --force` is green with **no committed golden changed**.

### Exactly what this orders, and what it does not

Orders: record→record, variant→variant, record→variant, variant→record, through
fields, payloads, arrays and vectors, in the two lists it is handed.

Does **not**:

- **A reference to a type that is in neither list.** No edge, no error. The PPX
  registers a variant in `kern_variants` without registering a record appearing only in
  its payload, so this is reachable — but it is a MISSING declaration, not a misordered
  one, and no ordering pass can supply it. It still surfaces as the backend's own
  *unknown type name*. Named here rather than papered over; worth its own backlog item.
- **A self-referencing declaration.** Emitted in place with the self-edge dropped; the
  backend's own field-type emission reports it.
- **A cycle between distinct declarations.** Refused with `Type_decl_cycle`, never
  emitted in some order.
- **Anything else a backend emits.** Headers, bindings and helper functions stay the
  caller's sequencing.

Not stability in the general sense, either: a blocked declaration is overtaken by later
independent ones, so `[a (needs c); b; c]` comes out `[b; c; a]`. The guarantee is the
narrower one the goldens need — a list already in a valid emission order is returned
unchanged — and both directions are pinned by tests.

## BREAKING (`Sarek_ir_codegen`, in `spoc.opam`'s `spoc.ir`)

Removed `sort_record_types_by_dependency`, `referenced_record_names`,
`gen_record_typedefs`; renamed `Record_type_cycle` → `Type_decl_cycle`; added
`type_decl`, `tie_break`, `sort_type_decls_by_dependency`, `referenced_type_names`,
`gen_type_decls`, `gen_c_type_decls`. The three C-family backends lose their local
`gen_variant_def` wrappers. No compatibility aliases, deliberately: an alias for
`sort_record_types_by_dependency` would still compile inside a per-kind emission loop
and silently reintroduce the bug. Nothing outside `spoc/ir` and its tests referenced any
removed name.

## Coverage that needs no device

`sarek/tests/codegen_golden/test_decl_order_all_backends.ml` sweeps **both shapes over
all five generators** by calling them directly. This exists because the e2e test exits 0
when it finds no device, so on CI — which has no GPU — it asserts nothing, and because
four of the five families have no hardware here.

Why it cannot pass vacuously: the backend table is checked against a pinned count; a
missing declaration anchor is a failure, not a skip; the order comparison is guarded on
both positions having been found (two `-1`s compare as ordered); and the referenced name
must appear **inside the referencing declaration's own body**.

That last one was a real hole, found in review. Counting the name across the whole
source is vacuous for shape B, because a variant's emission repeats its own type name in
every constructor signature — so with the record's variant field deleted the count stays
high, and on the C family (variants first anyway) the order assertion passed with
nothing left to order. Three of five cells were checking nothing. Measured after the
fix: deleting the edge now reports it on **five of five**.

The e2e test also refuses to print a pass line when no declaration-emitting device ran.
Interpreter and Native carry values and emit no struct declaration, so a CPU-only host
previously ran every case, passed every case, verified nothing, and printed `ALL PASSED`.

## Gates

Each exit code captured directly, `git status --porcelain` empty. **Every row below was
measured with the ambient `dune 3.23.0` on PATH** — a bare exit code with no toolchain
named is a claim wider than its measurement, and one of these rows is genuinely
toolchain-dependent (see below).

Re-measured on the rebased branch (onto `origin/main` after #376 landed), at
`9152895d`:

| Gate | Exit | dune |
| --- | --- | --- |
| `dune build` | 0 | 3.23.0 |
| `dune runtest --force` | 0 | 3.23.0 |
| `dune build @fmt` | 0 | 3.23.0 |
| `scripts/prove-red.sh` | 0 | 3.23.0 |
| `scripts/check-kb-properties.sh` (+ `.test.sh`) | 0 / 0 | 3.23.0 |
| `scripts/check-license-headers.sh` (+ `.test.sh`) | 0 / 0 | 3.23.0 |
| `scripts/check-cited-paths-exist.sh` (+ `.test.sh`) | 0 / 0 | 3.23.0 |
| `scripts/check-test-alias-coverage.sh` | 0 | 3.23.0 |
| `scripts/check-dune-dir-visibility.sh` | 0 | 3.23.0 |
| `scripts/check-ppx-construct-names.sh` (+ `.test.sh`) | 0 / 0 | 3.23.0 |
| `scripts/check-opam-clean.sh` | **0 on 3.23.0 / non-zero on 3.24.1** | both, see below |

### `check-opam-clean.sh` is SWITCH-DEPENDENT, not simply failing

An earlier version of this table recorded `1 — pre-existing` with no toolchain named.
That is right under one dune and wrong under another, which is why the row now names the
version. Both measurements, on this same clean tree:

| dune on PATH | result |
| --- | --- |
| ambient `3.23.0` | exit **0**, `.opam` files unchanged |
| project switch `/home/mathias/dev/SPOC/_opam`, `3.24.1` | regenerates `"dune" {>= "3.15" & >= "3.15"}` in all seven `.opam` files |

The variable is which switch is on PATH — not tree state, and not this branch, which
touches no `.opam` and no `dune-project`.

**The 3.23.0 pass is not vacuous**, which is the obvious worry about a gate that goes
green only on one toolchain. Verified by canary: appending `x-canary-397: "probe"` to
`sarek.opam` and re-running the script leaves zero occurrences of the canary afterwards
(exit 0), so `make opam` genuinely regenerates the file under 3.23.0 rather than
skipping the work.

The underlying issue is `make opam` non-idempotence under newer dune, which is a
packaging concern independent of this PR — `kb/support/packaging.md` records the
2026-07-02 history of this guard. I was told this is tracked as backlog-213; **I could
not verify that** — there is no backlog file in-tree and GitHub #213 is an unrelated
merged PR — so the tracking reference is stated as unconfirmed rather than cited.

No new gate script, so no new `kb/properties.md` `gate-red-path` row is due; the new
checks are unit tests, and `prove-red.sh` only scans `scripts/` and `ci/`. They were
proved red by mutation instead, fix committed first:

| Mutation | Proved |
| --- | --- |
| `gen_type_decls` emits in input order (pre-fix behaviour) | shape A red on CUDA/OpenCL/Metal, shape B red on GLSL/WGSL — the measured device matrix, reproduced with zero devices |
| each of the 5 backends stops passing its variants | that backend's row is live; missing-anchor failure names it |
| shape A / shape B loses the edge | body-scoped vacuity guard fires on 5 of 5 (the old whole-source count left 3 silently green) |
| a backend dropped from the sweep | pinned-count failure |
| `referenced_type_names` stops reporting variant names | shape B cross edge inverts |
| the sort stops walking variant payloads | shape A cross edge inverts |
| self-edge drop keyed on the NAME again | the same-named record/variant edge is lost |
| `~tie_break` ignored | edge-free mixed set no longer byte-identical |
| placement picks the LAST ready entry | tie-break pin fires |
| visited set records only NAMED nodes | `TVec` cycle hits the 10s deadline with its own message |
| visited set keyed on structural equality | non-termination (as `Out of memory`, not the deadline message) |
| e2e counts Interpreter/Native as declaring | CPU-only run falsely prints `ALL PASSED on 3 …` |
| `Type_decl_cycle` printer registration removed | renders `Type_decl_cycle(_)` instead of the names |

Three mutations were themselves defective on first attempt — two failed to compile (a
build error proves nothing) and one reversed a dependency list under `List.for_all`,
which is a semantic no-op that left the subject unmodified and green. All three were
corrected and re-run.

## Review

`roster-review` fan-out: unscoped fresh read, architect pass, and `codex` as the
cross-runtime reviewer. `codex` returned NO-GO with four findings; the architect pass
added a defect and six design/coverage findings. All are addressed in the third commit,
including the two real defects (name-keyed node identity, and the shape-B vacuity hole).

**`opencode` could not run**: it is configured against a local ollama
(`qwen3.6:27b`) that is not reachable on this host, and every request failed with
`AI_APICallError: Cannot connect to API`. It produced no review. The attestation records
`codex,claude` only — an earlier entry naming `opencode` was corrected in the ledger.

## Independent adversarial review (post-hoc), and what it changed

CodeRabbit reported *Review rate limited* on the original head and did not read the PR,
so it got a separate independent adversarial pass. That pass verified the claims above by
mutation rather than by reading — the pre-fix device matrix was reproduced with zero
devices and again on real hardware — and found three further defects, each fixed with
the fix committed before the mutation that proves it red:

- **`bdca651a` — the e2e reproducer went quiet when its own edge was erased.** Both
  kernels select their constructor behind an `if` to defeat tag erasure, `n > 0` is
  invariantly true, and nothing enforced any of it. Measured: with the ordering bug fully
  reintroduced AND the constructors written literally, the file printed
  `ALL PASSED on 4 declaration-emitting device(s)` and exited 0. `require_cross_edge` now
  checks the lowered IR for both edges before any device is opened.
- **`883574b2` — a same-named record/variant cycle rendered as `[t; t]`.** Node identity
  is the position, so such a pair is two distinct nodes; the cycle payload did not carry
  the kind and named neither declaration. Found by CodeRabbit once it did run. No test
  covered the same-name cycle at all; one was added.
- **`9152895d` — `"CUDA"` is a family name, so the device count could never see it.**
  `declaring_frameworks` is the set deciding whether this file is entitled to print its
  pass line. `Device.framework` reports the resolved backend (`CUDA/PTX`, `CUDA/C`);
  `"CUDA"` is the family name `Device.init ~frameworks` accepts as a *request*. The key
  could never match, so a CUDA-only host would have printed NOTHING VERIFIED for a run
  that genuinely exercised the ordering. This is the exact failure mode #376 documented
  in `test_record_field_store.ml` after hitting it twice. Corrected to the reported
  vocabulary, with #376's v3 family-name check applied so it cannot rot.

That last fix also puts ZLUDA on the loader path for this test, as #376 did for its two,
which turns the CUDA/PTX exclusion from prose into something every run exercises:
**CUDA/PTX ×2 now enumerate and PASS both shapes while the summary still reports 4
declaration-emitting devices, not 6.** 18 cells, up from 14.

## Summary by CodeRabbit

- **New Features**
  - Added unified dependency ordering for record and variant type declarations.
  - Backend-generated code now emits cross-referenced declarations in the correct order.
  - Improved cycle detection and reporting for recursive type declarations.
  - Added support for consistent ordering across CUDA/HIP, OpenCL, Metal, GLSL/Vulkan, and WGSL.

- **Bug Fixes**
  - Fixed declaration-order issues when records and variants reference one another.

- **Tests**
  - Added unit, golden, and end-to-end coverage for cross-kind declaration ordering and execution.


mathiasbourgoin:
# Independent adversarial review — #397

CodeRabbit reported *Review rate limited* and did not read this PR, so this is its
only independent pass. I did not write this code. Everything below is a command I
ran in my own detached worktree at `8ae7ac12`, not a reading.

**Verdict: GO**, with one defect found and fixed in `ccd45cbe` (pushed to this
branch). The algorithm is correct — I attacked it from several directions and did
not break it. The defect was where this repo's defects always are: a gate that
could not fail.

---

## The defect: the e2e reproducer went quiet when its own edge was erased

`sarek/tests/e2e/test_record_variant_decl_order.ml` is the ONLY hardware coverage
of backlog-211. Its header correctly identifies tag erasure as the trap, explains
that both kernels select their constructor behind an `if` for that reason, and
then enforces none of it. `n > 0` is invariantly true, so both `if`s read as dead
code a later reader would be right to simplify.

Measured — both halves of the bug reintroduced AND the constructors written
literally:

```
# mutation 1: gen_type_decls emits `decls` instead of
#             `(sort_type_decls_by_dependency decls)`   [the backlog-211 bug]
# mutation 2: `if n > 0 then At q else Nowhere`      -> `At q`
#             `{gk = (if n > 0 then Level 2.0 else Off); ...}` -> `{gk = Level 2.0; ...}`
$ ./_build/default/sarek/tests/e2e/test_record_variant_decl_order.exe
...
ALL PASSED on 4 declaration-emitting device(s)
EXIT=0
```

With the ordering bug fully present, the file prints its pass line and exits 0.
The summary already refuses to claim a pass when no *declaration-emitting device*
ran — that guards the DEVICE dimension. Nothing guarded the EDGE dimension, which
is the one the header calls the trap. `ALL PASSED` then asserts a verification
that did not happen, which is the dominant class in this repo.

**Fix (`ccd45cbe`)**: `require_cross_edge` reads the lowered IR before any device
is opened and requires BOTH ends of each edge — a variant payload naming a record
that is in `kern_types`, and a record field naming a variant that is in
`kern_variants` (a dangling reference is nothing for the ordering pass to order).
It runs pre-device, so a CPU-only host and CI catch the regression too.

Committed first, then proved red on each half independently:

| Mutation | Result |
| --- | --- |
| shape A constructor written literally | exit **1** — `NOTHING TO VERIFY in shape A … kern_types=[…probe_pt] kern_variants=[]` |
| shape B constructor written literally (shape A intact, so the guard must reach B) | exit **1** — `NOTHING TO VERIFY in shape B … kern_types=[_erec_float32_float32] kern_variants=[]` |
| neither | exit 0, `both cross-kind edges are present in the lowered IR (not erased)`, then 14/14 |

Note shape B's red output reproduces the collapsed `_erec_float32_float32` the PR
body describes — the erasure story in the body is accurate.

---

## Claims verified

**1. One interleaved pass, five generators.** Verified by reading all five call
sites and by mutation. Removing the sort reproduces the pre-fix device matrix with
**zero devices**, on exactly the right cells:

```
$ # mutation: emit `decls` unsorted
$ ./_build/default/sarek/tests/codegen_golden/test_decl_order_all_backends.exe
FAIL shape A: 3 problem(s) across 3 of 5 backend(s) (CUDA (also HIP, via Hip_plugin), Metal, OpenCL)
FAIL shape B: 2 problem(s) across 2 of 5 backend(s) (GLSL (Vulkan), WGSL)
```

and on real hardware the same mutation gives exactly the matrix in the PR body —
shape A red on both OpenCL / green on both Vulkan, shape B the exact inverse. The
red half of claim 5 is confirmed on this host, not taken on trust.

**2. `referenced_type_names` completeness.** The match names all 12 `elttype`
constructors with **no wildcard**; there is no tuple type in the IR. `spoc/ir/dune`
carries no warning suppression and this project treats warning 8 as an error — I
confirmed that by adding a new aggregate constructor to `elttype`, which produced
`Error (warning 8 [partial-match])`. So a future aggregate type cannot be added
without the compiler forcing this function to handle it. That is a durable
guarantee, and stronger than the `.mli` sentence claims. Physical-identity keying
is correct and cannot narrow the result (anything reachable from a re-visited node
was accumulated by the walk that first entered it).

Helper signatures are a non-issue: all five backends emit type declarations before
helpers and bindings, which I checked in each diff hunk.

**3. Position-keyed node identity.** Correct, and the same-named record/variant
edge is pinned. I probed the aliasing cases you flagged:
- two identical declarations at different positions — both become dependencies of
  a referencing entry, both get placed; no edge lost;
- a name in both lists — that is the cross-kind edge, handled;
- empty list — `loop ()` returns immediately.

**4. `~tie_break` / no golden changed — measured, not assumed.** `dune runtest
--force` exit **0**, `git status --porcelain` **empty**, and `find _build -name
'*.corrected'` returns nothing. Structurally the tie-break cannot re-introduce a
wrong order for an untested shape: it only chooses the pre-sort input order, and
the sort is a correct topological pass, so any real edge is honoured either way.
Its only effect is on the relative order of *unrelated* declarations.

**5/6. Erasure and the `if`.** Confirmed load-bearing — see above. This was the
finding.

**7. `Type_decl_cycle` printer.** Present and registered.

**8. `~13` prove-red mutations.** Not re-run individually; `scripts/prove-red.sh`
exits 0.

**HIP/CUDA-C coverage.** The body's claim that HIP rides the CUDA row is true:
`sarek-hip/Hip_plugin.ml:36` calls `Sarek_codegen.Sarek_ir_cuda.generate_with_types`.
The CUDA-C path is a real production path (`sarek-cuda/Cuda_c_plugin.ml:38`), not
only PTX, so the sweep's CUDA row exercises shipped code. The "no device result
for CUDA/HIP/Metal/WGSL" caveat is stated accurately and is not wider than the
source sweep supports.

**The named-but-unfixed gap.** The body, the `.mli` and `kb/spoc/ir.md` all state
that a reference to an undeclared type is neither an edge nor an error. No prose
anywhere implies declaration order is sound in general. Not over-wide.

---

## `check-opam-clean.sh` — it does NOT reproduce for me

Plainly, as asked: **exit 0**, on a clean tree, dune 3.23.0, with
`git status --porcelain` empty afterwards. I could not reproduce the exit 1 two
agents reported tonight. The gate table in the PR body reports `1 — pre-existing`
for this row, and on a clean tree at this dune version that row is wrong. Worth
correcting the body rather than carrying a phantom known-failure forward.

## Gate exit codes (at `ccd45cbe`, each captured directly)

| Gate | Exit |
| --- | --- |
| `dune build` | 0 |
| `dune build @fmt` | 0 |
| `scripts/check-license-headers.sh` (+ `.test.sh`) | 0 / 0 |
| `scripts/check-opam-clean.sh` | **0** (see above) |

`git status --porcelain` empty throughout.

---

## Minor, not blocking

**The sweep's vacuity guard is single-payload-scoped on GLSL.** It is genuinely
body-scoped now — I deleted the edge in each shape and got `5 of 5` backends
reporting in both, which settles the item that was vacuous before:

```
shape A, edge deleted: 8 problem(s) across 5 of 5 backend(s)
shape B, edge deleted: 7 problem(s) across 5 of 5 backend(s)
```

But `struct_family.body` slices from `struct <name> {` to the first `}`, and
`gen_variant_def_glsl` hoists a *multi-payload* constructor's payload into a
separate `<variant>_<C>_payload` struct emitted just before the variant struct. A
record referenced only from a multi-payload constructor would therefore sit
outside the sliced body. No false green today — both shapes use single payloads —
but if the sweep is ever extended to a multi-payload shape, the guard needs to
widen with it.

**A theoretical spurious cycle.** Two same-named *self-referential* declarations
would now each depend on the other and raise `Type_decl_cycle` where name-keying
emitted them. Unreachable — the PPX refuses a self-referencing record field at
alignment-resolution time and fusion concatenates PPX-produced lists — and the
`.mli`'s wording survives it (the self-edge at index `i` *is* dropped; the
reported cycle is between distinct declarations). Noting it only so it is on the
record.

---

## What I could NOT verify

- **No device result for CUDA, HIP, Metal or WGSL.** No such hardware here. Those
  four rest entirely on the 5-generator source sweep. The PR is honest about this.
- **The 13 prove-red mutations were not re-run individually**; I re-derived four
  of them (no-sort, both vacuity deletions, name-keyed identity via the unit test)
  and took the rest from `prove-red.sh` exit 0.
- **`opencode` did not run** and I have not attested it — it points at an
  unreachable local ollama here. I did not run `codex` either. The ledger entry
  for `ccd45cbe` names `claude` only, which is exactly what ran.
- **Whether `require_cross_edge`'s both-ends conjunct is load-bearing.** It makes
  the guard strictly stricter, and the only shape that would exercise it — a
  dangling reference to an undeclared type — is the known-unfixed gap and is not
  reachable from this test. Deliberately conservative; not proved necessary.

mathiasbourgoin:
## Follow-up — CodeRabbit's finding applied (`5971d29a`)

CodeRabbit reviewed after my push (it was rate-limited on the original head, which
is why the PR had no independent read until now) and posted one Trivial finding. It
is valid, and it is the same defect class as everything else here — a diagnostic
narrower than the design it reports on — so I fixed it rather than deferring.

**The defect.** Node identity in this pass is the POSITION, so a record and a
variant sharing a mangled name are two distinct nodes. That is the choice the whole
cross-kind fix rests on, and it is the one the cycle diagnostic did not carry: a
genuine cycle between such a pair produced a payload of two identical strings, and
the registered printer rendered

```
the type declarations [t; t] — they form a dependency cycle
```

which identifies neither declaration, and reads like a self-reference — the one
case this pass deliberately does NOT report as a cycle.

**Fix.** Each unplaced entry is tagged with its kind, so the message names a
`record t` and a `variant t`. The `.mli` said the exception carries "the unplaced
type names"; that phrasing is what went stale, so it now says the payload carries
kind-tagged declarations and explains why the kind is load-bearing rather than
decorative.

**No test covered the same-name cycle at all** — the two existing cycle tests use
distinct names, so neither could have caught this. Added
`test_same_name_cycle_names_both_kinds`; the two existing expectations are updated
for the new payload shape.

Committed first, then proved red. The mutation (drop the kind tag at the raise
site) trips the earlier cycle test first, so I relaxed those two expectations to
isolate the new case:

```
Fatal error: exception Failure("same-name cycle: the payload must distinguish
  the two same-named declarations, got [t; t]")
exit 2
```

— the exact `[t; t]` that motivated it. Restored, and the whole gate set re-run at
`5971d29a`: `dune build` 0, `dune runtest --force` 0, `dune build @fmt` 0,
`prove-red.sh` 0, the eight `check-*.sh` (+ their `.test.sh`) 0, and
`check-opam-clean.sh` **0** again. `git status --porcelain` empty. The e2e
reproducer still reports `ALL PASSED on 4 declaration-emitting device(s)`, 14/14.

One process note against my own work: my first attempt at this mutation silently
did not apply — ocamlformat had reflowed the raise site, so the anchor string no
longer matched — and my shell still printed `MUTATION LANDED` because I had chained
the check with `;` instead of `&&`. A Python `assert` on the anchor is what caught
it. A prove-red whose mutation never landed is a green that means nothing, which is
the failure mode this repo keeps cataloguing; flagging it because the same shape
would silently weaken any mutation table.

Verdict unchanged: **GO**.

mathiasbourgoin:
@coderabbitai review

Context for the re-read, since the standing `CHANGES_REQUESTED` is against `ccd45cbe` and the head is now `9152895d`.

**Your finding is fixed.** The cycle payload was rendering a same-named record/variant pair as `[t; t]`, naming neither declaration. `883574b2` tags each unplaced entry with its kind, so the message now names a `record t` and a `variant t`. The `.mli` said the exception carries "the unplaced type names" — that phrasing was the part that went stale, so it now states the payload carries kind-tagged declarations and why the kind is load-bearing rather than decorative.

No test covered the same-name cycle at all (both existing cycle tests use distinct names), so the fix is pinned by a new case, `test_same_name_cycle_names_both_kinds`, rather than left resting on the two you cited. Those two have their expectations updated for the new payload shape. Proved red by dropping the kind tag at the raise site, with the earlier expectations relaxed so the new case is the one that fires:

```
Fatal error: exception Failure("same-name cycle: the payload must distinguish
  the two same-named declarations, got [t; t]")
exit 2
```

**Two other things changed since you last saw this branch,** both worth a look:

- `9152895d` — `declaring_frameworks` named `"CUDA"`, a *family* name. `Device.framework` reports the resolved backend (`CUDA/PTX` / `CUDA/C`), so that key could never match and the declaration-emitting device count would have read 0 on a CUDA-only host, printing NOTHING VERIFIED for a run that genuinely exercised the ordering. Corrected to the reported vocabulary, with the family-name check from `test_record_field_store.ml` applied so it cannot rot. The dune rule now puts ZLUDA on the loader path, so CUDA/PTX ×2 enumerate and pass both shapes while the summary still reports 4 declaration-emitting devices, not 6 — the PTX exclusion is exercised rather than asserted.
- `bdca651a` — the e2e reproducer printed `ALL PASSED` and exited 0 with the ordering bug fully reintroduced, once its constructors were written literally and tag erasure removed the edge. It now checks the lowered IR for both cross-kind edges before opening any device.

The branch is also rebased onto current `origin/main` (#376 landed underneath it). The two conflicts were both pure append/append — `CHANGES.md` and `sarek/tests/e2e/dune` — resolved keeping both sides, and verified by diffing against both parents restricted to the files each side owns: zero deletions from main other than the eight-line stale backlog-203 paragraph this PR rewrites by design.

PR #398, opened 2026-07-31
fix(ppx): make the parser's accepted subset a refusal boundary, not a silent filter

mathiasbourgoin:
# The parser's accepted subset is now a refusal boundary (backlog-192 + backlog-208)

#390 fixed ONE silent drop: `Sarek_parse`'s `Pexp_match` arm read `pc_lhs` and
`pc_rhs` and never `pc_guard`, so a `when` guard vanished in the parser —
upstream of the typer, of every lowering pass and of backend selection — and the
kernel compiled and computed a different function than its source said.

That was not a one-off. This is the sweep: every arm and every record field
`sarek/ppx/Sarek_parse.ml` and its callee `sarek/ppx/Sarek_parse_helpers.ml` can
encounter, walked **from the ppxlib Parsetree type definitions outward** rather
than from a list of constructs anybody could think of, classified implemented /
refused / **silently dropped**, with every drop turned into a located refusal or,
where refusing would shrink the accepted subset rather than state it, into an
implementation.

**Thirty rows of the enumeration below have a drop as their before-status.** That
is a count of SITES, not of OCaml constructs: several rows cover a whole arm set
at once (`parse_type`'s last arm alone is eight `core_type` constructors, the
payload structure-item fold's is thirteen), so the number of individual
constructors involved is larger and this PR does not put a figure on it.

## The drops that can produce a wrong answer or a missing binding

| what was dropped | what happened | what it says now |
|---|---|---|
| `{ r with f = e }` — `Pexp_record`'s base, bound as `_base` | parsed as `{ f = e }`, base gone. **Not silent**: measured, OCaml rejects the re-emitted native literal with `Some record fields are undefined: y` on the user's own `with` line, never mentioning the `with` | a refusal naming the construct and telling you to name every field |
| `let x : t = e` — `pvb_constraint`, unread | since OCaml 5.1 the annotation lives there, not in the pattern. A kernel-local `let sum : float = …` was typed by inference with the declared width ignored; an annotated module constant hit "must have type annotations", a message false of the code in front of it | **implemented** — both spellings read |
| `let f x : t = e` — `Pexp_function`'s constraint slot, read as `_` | every kernel helper's declared result type discarded. There are 20-odd such helpers in this tree's own tests | **implemented** |
[...]

PR #399, opened 2026-07-31
fix(opam,gates): make .opam generation independent of which dune ran (backlog-213)

mathiasbourgoin:
# backlog-213 — `make opam` is not idempotent under the project's own switch

`scripts/check-opam-clean.sh` exits 1 on a clean tree for anyone using the documented setup. A gate that is red for everyone is functionally off.

## The variable is which dune is on PATH

Measured on `97a062a2`, both binaries named by absolute path:

| dune | how invoked | result |
| --- | --- | --- |
| 3.23.0 (`~/.opam/octez-setup/bin/dune`) | `dune build sarek.opam` | `"dune" {>= "3.15"}` |
| 3.24.1 (`<project>/_opam/bin/dune`) | `dune build sarek.opam` | `"dune" {>= "3.15" & >= "3.15"}` |

Not tree state, not branch content. Two reports of exit 1 were twice not reproduced because the reproduction used the other dune.

## Mechanism — from dune's own changelog and source, not inference

Dune injects `"dune" {>= <lang version>}` into every generated `.opam` from the `(lang dune X.Y)` stanza (`src/dune_rules/opam_create.ml`, `insert_dune_dep`, read in `_opam/.opam-switch/sources/dune.3.24.1/`). All seven packages **also** declared `(dune (>= 3.15))` in `depends`, so the two bounds met.

- dune **3.23** deduplicated them unconditionally — `CHANGES.md`, 3.23 "Fixed": *"Fix duplicate dune version bounds in generated opam files … the generated opam `depends` no longer contains redundant constraints like `{>= "2.7" & >= "2.7"}`"* (#3916, #11106).
- dune **3.24.1** gated that deduplication on the project's lang version — `CHANGES.md`, 3.24.1 "Fixed": *"Gate the `dune` version-bound deduplication in generated opam files (introduced in 3.23) on `(lang dune 3.23)`. Projects at earlier lang versions get the prior `And [...]` shape … restoring 3.22 behaviour"* (#14436).

This project is `(lang dune 3.15)` — the earlier side of that gate. Note 3.24.**0** still deduplicates; the divergence is 3.24.1-and-later.

## Fix, at the source

The seven redundant `(dune (>= 3.15))` entries are deleted from `dune-project`. The injected bound is unchanged, so **the published floor is unchanged**. The entire `.opam` diff is dune inserting `"dune"` at the head of `depends` instead of at the position the explicit entry held:

```
$ git diff origin/main HEAD -- '*.opam' | grep -E '^[+-]' | grep -v '^[+-][+-]' | sort | uniq -c
      7 +  "ocaml" {>= "5.4.0"}
      7 -  "ocaml" {>= "5.4.0"}
```

Seven files, one line moved each, nothing incidental. No generated file was committed for any other reason.

Rejected alternatives, recorded in the gate header because the lang bump is what anyone tripping rule 1 will reach for first: bumping to `(lang dune 3.23)` is a semantics switch across eight minor versions and raises the *effective* dune floor for consumers, while leaving the duplicated fact in place; pinning a dune version has nowhere to live that everyone reads (`.ocamlformat`'s pin works only because ocamlformat refuses to run on a mismatch — dune has no such handshake).

## Before / after, both dunes, by absolute path

| | dune 3.23.0 | dune 3.24.1 |
| --- | --- | --- |
| **before**, generate + `git diff` vs tracked | no diff | **DIFFERS** |
| **after**, generate + `git diff` vs tracked | no diff | no diff |
| **after**, output compared between the two dunes | byte-identical | byte-identical |
| **after**, `x-canary` re-check | removed → real regeneration | removed → real regeneration |
| `check-opam-clean.sh` (full gate) **before** | *(see below)* | **exit 1** |
| `check-opam-clean.sh` (full gate) **after** | *(see below)* | **exit 0** |

The canary is the anti-vacuity control: an `x-canary` line appended to `sarek.opam` is removed by both dune versions after as well as before, so "it passes now" is not a pass over a target that stopped regenerating.

### What could not be verified

**The full `check-opam-clean.sh` cannot complete under any dune 3.23.0 on this machine**, so the reported green under 3.23.0 was not reproduced and is not quoted anywhere as an observation of that gate. It runs `make opam` = a full `dune build @install`, and:

- in the project switch, `ctypes.0.24.0` installs a `(lang dune 3.24)` `dune-package` that 3.23 refuses to read;
- in the octez-setup switch, that switch's OCaml 5.3.0 / ppxlib cannot build the PPX.

Both used to leave through `set -e` with no message of the gate's own and an exit code indistinguishable from "the tree is dirty" — which is the sharper reason this took three attempts to pin, and is what the new INCONCLUSIVE path exists for. Both now exit 2 saying so. The cross-dune evidence above is therefore at the **generation** layer, which needs no dependency resolution and is exactly where the bug lived.

Also unverified: dune 4's behaviour (#14436 is a 3.x change), which is why a major lang version above 3 makes the new gate **refuse** rather than go inert.

## Making it not recur

Nothing checked that the two dune versions agree, and nothing named the toolchain a verdict was about.

**`scripts/check-dune-opam-portability.sh`** (new; CI build job + `make test-all`). Rule 1: below `(lang dune 3.23)`, no package may declare `dune` in `depends`, in any form. Rule 2: no generated `.opam` may carry a multi-term `"dune" {...}` constraint — rule 2 holds at *every* lang version, so rule 1 going inert above 3.23 never leaves the gate with no assertion. Both textual: no build, no second toolchain.

It parses rather than greps, and that is not gold-plating. The first draft *was* a pile of greps, and four independent review passes each found a different shape it walked past — `(dune :build)`, a `dune fmt`-wrapped `(dune\n (>= 3.15))`, an opam-formatter-wrapped constraint, and a quoted escape-spelled `"\x64une"` — all while its comment claimed it matched `(dune ...)`. A whole-file scan for `"dune"` also reached the `build:` field, where dune emits `["dune" "subst"] {dev}`; the constraint is now read out of a bracket-balanced `depends:` section. A stray `)` made an earlier revision **loop forever** (exit 124 under a 10s timeout) instead of refusing.

Red path is a prove-red-spec block, **13 mutations, each observed red with its exact exit code and message**: three shapes of re-added explicit entry (1), the escape-spelled name (1), the duplicated bound on one line and across two (1), a deleted *and* a duplicated lang stanza (2, both sides of the exactly-one guard), one `.opam` of seven declaring no dune dependency (2 — an at-least-one guard passed at 1-of-7, measured), a declared package with no `.opam` (2), a stray `)` (2), an unterminated string (2), and a rule-1 violation combined with a cannot-measure condition, asserting the finding is flushed rather than discarded by the refusal. `EXPECTED_SUBJECTS` 6 → 7; the suite is now 7 subjects / 46 mutations.

**`scripts/check-opam-clean.sh`** now names the dune binary and version in every verdict. Review found its verdict was wider than its code in two further ways:

- it diffed `sarek.opam` **alone** while printing a verdict about `make opam`. It now snapshots between the two runs (a target that oscillated back onto the tracked bytes read as convergent) and diffs the whole set. Positive control: a drift confined to `sarek-hip.opam`, which the old single-file diff passed, now fails naming that file.
- the set is **pinned** to the packages `dune-project` declares, via a new `--list-packages` mode on the portability gate so there is one parser and no second copy of the list. A glob or `git ls-files` shrinks silently when a file is deleted or `git rm --cached`ed, and the PASS line would just print a smaller denominator.

Everything that cannot measure exits 2 INCONCLUSIVE rather than 1: no dune, an unrunnable dune, a `make opam` that does not build, a missing or untracked generated `.opam`, a failed snapshot, `cmp`/`git diff` above 1. `opam` joins `.PHONY` — a file of that name would have made both invocations no-ops whose snapshots compare equal.

**`.github/workflows/ci.yml`** gains `git diff --exit-code -- '*.opam'` after the `dune build @install` the build job already ran. CI has been regenerating the `.opam` files and discarding the result for the life of the repo, which is why this was invisible here and surfaced only on a contributor's switch. It is one build compared against the committed bytes — not a two-run convergence check — and it is the only CI-side backstop, because **no workflow invokes `check-opam-clean.sh`**.

`kb/properties.md` gains `KB-GATE-DUNE-OPAM-PORTABILITY` and rewrites `KB-GATE-OPAM-CLEAN`, whose old reason ("fails loudly and locally") was true and beside the point — it was failing loudly and locally for every user of the project switch.

## PR #398 is undisturbed

`dune-project`'s ppxlib floor is untouched at `>= 0.37.0`, and `scripts/check-ppxlib-floor.sh` passes along with its `.test.sh`.

## Gates

Toolchain for every row: switch `/home/mathias/dev/SPOC/_opam`, dune **3.24.1**, OCaml **5.4.0**, ppxlib **0.38.0**, ctypes 0.24.0. Exit codes captured directly, never a prose "PASS".

| gate | exit |
| --- | --- |
| `dune build` | 0 |
| `dune build @fmt` | 0 |
| `scripts/check-opam-clean.sh` — **dune 3.24.1**, project switch | **0** |
| `scripts/check-opam-clean.sh` — **dune 3.23.0**, project switch | **2 INCONCLUSIVE** (cannot build; see above) |
| `scripts/check-opam-clean.sh` — **dune 3.23.0**, ambient octez-setup switch (OCaml 5.3.0) | **2 INCONCLUSIVE** (cannot build the PPX) |
| `scripts/check-dune-opam-portability.sh` | 0 |
| `scripts/check-ppxlib-floor.sh` | 0 |
| `scripts/check-ppxlib-floor.test.sh` | 0 |
| `scripts/prove-red.sh` | 0 — 7 subjects, 46 mutations |
| `scripts/check-kb-properties.sh` | 0 — 43 declarations |
| `scripts/check-cited-paths-exist.test.sh` | 0 |

Review: two `codex` passes and three fresh-read/architect passes; every finding from all five is addressed above or in the code. The dominant defect class they found — a claim wider than its code — accounted for every single finding again, in prose I had written, including two comments in one file that contradicted each other about what had been measured.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved package metadata consistency by ensuring generated package files remain synchronized and converge reliably across repeated builds.
  * Added validation for Dune compatibility and dependency declarations to prevent portability issues across supported toolchain versions.

* **Chores**
  * Integrated portability and package-cleanliness checks into the standard test workflow.
  * Enhanced validation reporting with clearer toolchain details, package coverage, and distinct failure outcomes.


mathiasbourgoin:
All five addressed in `e3611c1d`. Two of them found real defects **in the fix itself**, and one of those is the sharpest instance of this repository's house defect class in the entire change. Taking them in order of how badly I got them wrong.

---

## 1. CRITICAL — dead diagnostics under `set -e`. Correct, and it is the defect this PR exists to remove, reintroduced inside the code written to remove it.

`scripts/check-opam-clean.sh` ran

```bash
cmp -s -- "$SNAPSHOT_DIR/$f" "$f"
case $? in ...
```

with `set -e` active — as the comment two lines above the site itself says. The shell terminates **at** the `cmp`. The `case` never runs. Every message on both paths was dead code and the gate exited with a bare status.

Reduced to nine lines, this prints nothing at all and exits 1:

```
$ ./setest.sh; echo "EXIT=$?"
EXIT=1
```

The reason that is worth stating plainly rather than quietly patching: **this PR's entire justification** is that the old gate "left through `set -e` carrying no message of its own and an exit code indistinguishable from *the tree is dirty*". I wrote that sentence, added an INCONCLUSIVE path to fix it, and then reintroduced the identical construct one layer down in the same file. The defect class named in the PR body — a claim wider than its code — was in the code enforcing the claim.

**Fixed** at both sites with `rc=0; cmd || rc=$?`, which `set -e` does not kill.

### Why three review passes and my own positive controls missed it

The controls asserted **exit codes**. An unreached diagnostic still produces a plausible exit code. Forcing each branch with PATH stubs against the pre-fix form:

| forced condition | pre-fix exit | pre-fix message | post-fix exit | post-fix message |
| --- | --- | --- | --- | --- |
| `cmp` differs (stub `make` diverges the two runs) | 1 | **MISSING** | 1 | PRESENT |
| `cmp` exits 2 (stub `cmp`) | 2 | **MISSING** | 2 | PRESENT |
| `git diff` exits 128 (stub `git`) | **128** | **MISSING** | 2 | PRESENT |

**Two of the three had the correct exit code with a dead message.** No exit-code assertion could ever have caught this — which is exactly your point, and the reason it survived. The third leaked git's raw 128 as the gate's status.

There is a second, more embarrassing mechanism: the drift control that *did* exercise the `git diff` path passed on an earlier revision, because at that point the code was `if ! git diff ...; then`. I converted it to the bare form in a later pass and re-ran only the **green** path afterwards. The control was run before the change that broke it.

All four red paths are now controlled on **message text**, not status: a drift confined to `sarek-hip.opam`, a non-converging stub `make`, and `cmp`/`git` stubs forced above 1. The pre-fix counter-proof above is part of that record, so "reached" is distinguished from "right exit code" by construction.

### Re-audit for further instances

Every `$?` and every status read in both touched scripts, as you asked, since one instance rarely travels alone:

- `check-opam-clean.sh` — `set -e` active. Two `$?` sites, both the broken shape, both fixed. Every other status-consuming construct was already in a form `set -e` spares: `if ! cmd`, `cmd || inconclusive ...`, `$(...) || ...`, `grep -q` inside `if`.
- `check-dune-opam-portability.sh` — `set -uo pipefail`, **no `-e`**, and no `$?` reads at all.

No third instance. No `$?` after a pipe anywhere.

---

## 3. MINOR — you are right about 3.23.1, and I was wrong twice.

I cited the installed sources and treated that as settling it. It did not, because I never checked which section the entry was in. The headings in `_opam/.opam-switch/sources/dune.3.24.1/CHANGES.md`:

```
line   3:  3.24.1 (2026-07-21)
line  37:  3.24.0 (2026-06-21)
line 132:  3.23.1 (2026-05-14)
line 164:  3.23.0 (2026-05-04)
```

The gating entry is at **line 143 — inside 3.23.1**, verbatim:

> **3.23.1 (2026-05-14)** › ### Fixed
> - Gate the `dune` version-bound deduplication in generated opam files (introduced in 3.23) on `(lang dune 3.23)`. Projects at earlier lang versions get the prior `And [...]` shape — e.g. `{>= "3.17" & >= "3.20"}` — restoring 3.22 behaviour and avoiding a silent change to opam output on dune-binary upgrade. (#14436, @robinbb)

and the dedup entry is at line 252, inside **3.23.0**:

> **3.23.0 (2026-05-04)** › ### Fixed
> - Fix duplicate dune version bounds in generated opam files. When users declare `(dune (>= X.Y))` matching or exceeding the `(lang dune X.Y)` version, the generated opam `depends` no longer contains redundant constraints like `{>= "2.7" & >= "2.7"}`. (#3916, #11106, @robinbb)

Neither the 3.24.0 nor the 3.24.1 section mentions deduplication or opam generation at all. I read the entry near the top of the file and assumed the section that is also near the top. Your web query beat my "stronger" local evidence because I misread the local evidence.

**This makes the mechanism cleaner, not murkier.** For a project at `(lang dune 3.15)`:

- **< 3.23.0** — no deduplication → `{>= "3.15" & >= "3.15"}`
- **3.23.0** — deduplication, ungated → `{>= "3.15"}`
- **≥ 3.23.1** — deduplication gated on `(lang dune 3.23)` → `{>= "3.15" & >= "3.15"}` again

So **3.23.0 is the only release in existence that collapses the pair for this project**, and the two binaries I measured sit either side of that one-patch window. The tracked `.opam` files on `main` were therefore generated by a 3.23.0 specifically.

It also retracts a correction I volunteered unprompted: I said "3.24.0 still deduplicates". It does not — it is after 3.23.1 and inherits the gating.

`DEDUP_LANG = (3, 23)` is unaffected; the *lang* threshold was always right. Corrected in `dune-project`, both script headers, the two runtime messages that named versions, `ci.yml`, and both KB rows — the portability row now records the misattribution and how it happened, since that is the durable part.

---

## 2. MAJOR — untracked `.opam` has no CI coverage. Correct. Adding the assertion inline; deliberately **not** wiring the script, and here is why.

`git diff` compares tracked files only, so a newly declared package whose generated `.opam` was never committed produces no diff and passes in silence. The build step now also fails on:

```bash
git ls-files --others --exclude-standard -- "*.opam"
```

being non-empty, with the offending paths printed as a `::error::`.

On the preferred option — wiring `check-opam-clean.sh` into CI — you said verify, do not assume, and **I could not verify it**, so I did not do it. It shells out to `make`, runs the build twice more inside the container, and its INCONCLUSIVE exit 2 has never been observed there. Confirming that behaviour needs a real run of the `spoc-ci` image with a full opam install, which I cannot do from here. Adding an unverified exit-2 path to CI in order to close a silent-pass would trade one unmeasured claim for another — the precise trade this PR is about. The rationale is recorded in the step's comment and in the KB row so the better end state is not lost.

---

## 4 & 5. MINOR — two unguarded exceptions. Correct, and fixing the second one properly took two attempts.

Both raised bare Python exceptions, leaving exit **1** — this gate's code for "the tree is wrong" — on input dune cannot read, which the contract puts at **2**. Both sat next to branches that already refused correctly, which is what made them inconsistent rather than merely absent.

- **`int()` on a truncated numeric escape** (`"\xZZ"`, `\x` at EOF, `\12` with two digits): all three numeric forms now share one guarded decode that refuses with `malformed numeric string escape`.
- **`text.index("[")` on a bracketless `depends:`**: switched to `find` with a `< 0` refusal, as you suggested — **and that was still wrong**. Unbounded `find` always succeeds, because every opam file has a `build: [` further down, so the guard was unreachable and the *build* list was silently parsed as the depends list. prove-red caught it by refusing to credit a mutation that exited 2 with an unrelated message. The `[` must now be the next non-whitespace token after `depends:`.

Both are new prove-red mutations. The suite is **7 subjects, 48 mutations**, every one observed red with its exact exit code and message.

---

## Gates

Switch `/home/mathias/dev/SPOC/_opam`, dune **3.24.1**, OCaml **5.4.0**, ppxlib **0.38.0**, ctypes 0.24.0. Exit codes captured directly.

| gate | exit |
| --- | --- |
| `dune build` | 0 |
| `dune build @fmt` | 0 |
| `scripts/prove-red.sh` | 0 — 7 subjects, 48 mutations |
| `scripts/check-dune-opam-portability.sh` | 0 |
| `scripts/check-opam-clean.sh` — dune 3.24.1 | 0 |
| `scripts/check-ppxlib-floor.sh` (+ `.test.sh`) | 0 / 0 |

Both dunes re-verified on the final tree: generation byte-identical, `git diff` clean under each, and the `x-canary` anti-vacuity control removed by both. `git diff origin/main HEAD -- '*.opam'` is still exactly seven files × one moved `"ocaml"` line — no regenerated `.opam` committed for any other reason.

## One thing I still cannot verify

`check-opam-clean.sh` cannot complete under any dune 3.23.0 on this machine and exits 2 INCONCLUSIVE both ways. The coordinator has since established the cause is **tree state, not toolchain**: a pre-#398 tree builds fine at exit 0 under ambient 3.23.0, while current `main` fails with `no constructor "Ptyp_open"`. The script's header records the two build reasons rather than a dune-version reason.

mathiasbourgoin:
## The guard that went red was wrong, not the tree. Option 3, established by counting.

`build` now passes at `c43ffc8c` (7m49s), with every step executed rather than skipped.

The answer is **not** "the fix is incomplete" and **not** "the committed `.opam` files are stale". It is the third possibility, and the specific shape is the one this whole PR is about: **my assertion was wider than its intent.**

### What the guard actually listed

```
$ awk '/not committed/,0' ci.log | grep -oE '[^ ]+\.opam$' | wc -l
826
$ ... | grep -c '\.opam-ci/'
826
$ ... | grep -cvE '/'          # repository-root .opam files
0
```

**826 files named. All 826 under `.opam-ci/`. Not one from this repository.** They are the `.opam` files of dependencies' own source trees — `ocamlbuild.0.16.1/ocamlbuild.opam`, `ppxlib.0.38.0/ppxlib.opam`, `yojson.3.0.0/yojson.opam`, and so on.

The cause: `OPAMROOT` in that step is `/work/.opam-ci`, i.e. **inside the workspace**, and `.opam-ci` is not in `.gitignore`. And a bare `*.opam` git pathspec is matched with `FNM_PATHNAME` **off**, so `*` crosses `/` and the pattern means "any `.opam` at any depth" — not "at the root", which is what the comment beside it claimed and what I intended.

So: an assertion wider than its code, in the guard I added to close a review finding about assertions wider than their code, on a PR whose stated subject is that defect class. That is the fourth instance in this change and the second one I introduced while fixing the previous one.

### The fix, and both polarities measured

Both pathspecs are now `:(top,glob)*.opam` — `:(glob)` stops `*` at `/`, `:(top)` anchors to the repository root. Exercised locally against the exact assertion text, not a paraphrase of it:

| condition | want | got |
| --- | --- | --- |
| clean tree + planted `.opam-ci/.../ppxlib.opam` | 0 | **0** — dependency `.opam` now invisible |
| uncommitted root `sarek-newpkg.opam` | 1 | **1**, names the file |
| drifted tracked `sarek.opam` | 1 | **1** |
| clean tree, no opam root | 0 | **0** |

I did **not** add `.opam-ci/` to `.gitignore`, which would also have silenced it: a pathspec that is only honest because an ignore rule happens to cover the difference is two copies of one fact, which is exactly the trap `backlog-213` exists to close.

## What the red run proved that I could not prove locally

This is the more interesting result, and it settles your options 1 and 2 by measurement rather than by preference.

The step is an `&&` chain: `dune build @install && git diff --exit-code … && <untracked check>`. The error printed was **the untracked message**, which sits last. So `git diff --exit-code` had already **exited 0** — and there are no `diff --git` hunks anywhere in the 4,021-line log.

**CI's own dune regenerated all seven `.opam` files and produced byte-identical output to what is committed.** That is third-party confirmation of the fix's premise from a toolchain I do not control, which neither of my two local dunes could give.

So:

- **Option 1, fix incomplete — refuted.** Generation is dune-version-independent for CI's dune too.
- **Option 2, committed files stale — refuted.** They are canonical. `git diff origin/main HEAD -- '*.opam'` remains exactly 7 files × one moved `"ocaml"` line, and that is the complete set of differences CI sees.
- **No dune version needs pinning in CI.** Nothing to escalate.

### CI's dune version, read rather than assumed

**3.24.1.** There is no `installed dune.X` line, because dune came from the restored switch cache — so I took it from `dune-build-info`, which is version-locked to dune: `∗ install dune-build-info 3.24.1`, `∗ installed dune-build-info.3.24.1`. The first log also showed a `sources/dune.3.24.0` directory, which is a stale leftover in the cached switch and would have been the convenient wrong answer; on the passing run the only sources dir is `dune.3.24.1`. Same version as my project switch, so the local reproduction already covers it and no third bisect was needed.

## Toolchains

| where | switch | dune | OCaml | ppxlib | ctypes |
| --- | --- | --- | --- | --- | --- |
| local | `/home/mathias/dev/SPOC/_opam` | **3.24.1** | 5.4.0 | 0.38.0 | 0.24.0 |
| local, second binary | `~/.opam/octez-setup` | **3.23.0** | 5.3.0 | — | — |
| **CI** `build` job | `.opam-ci/5.4.0` | **3.24.1** | 5.4.0 | 0.38.0 | 0.24.0 |

## Gates, local, all exit codes captured directly

`dune build` 0 · `dune runtest --force` 0 · `dune build @fmt` 0 · `prove-red.sh` 0 (7 subjects, 48 mutations) · **`check-opam-clean.test.sh` 0 (7/7)** · `check-opam-clean.sh` 0 · `check-dune-opam-portability.sh` 0 · `check-ppxlib-floor.sh` 0 (+`.test.sh` 0) · `check-kb-properties.sh` 0 (+`.test.sh` 0) · `check-license-headers.sh` 0 · `check-cited-paths-exist.sh` 0 (+`.test.sh` 0) · `check-workflow-steps.sh` 0 · `check-workflow-toolchain-env.sh` 0 · `check-review-finding-tests.js 399` 0.

CI on `c43ffc8c`: `build` pass, `check-generated-code` pass, `formal-proofs` pass, `pocl-runner-probe` pass. Steps confirmed executed, not skipped: *No explicit dune dependency in dune-project*, *Prove-red*, *Resolved review findings carry a test*, *Harness self-tests*, *Build SPOC packages*. `scripts/check-opam-clean.test.sh` reported `OK: 7/7` inside CI.

## Also in this round

`scripts/check-opam-clean.test.sh` is new — the Critical `set -e` finding was previously covered only by shell I ran by hand and deleted. Seven cases, each asserting the **promised message** rather than a status, with `make`, `dune`, `cmp` and `git` replaced by PATH stubs so it needs no switch. Reverting either `|| rc=$?` turns three of the seven red; observed. `KB-GATE-OPAM-CLEAN` therefore carries a real `red_path` for the first time instead of a `reason` for having none.

Four entries added to `scripts/review-finding-escapes.tsv`. The gate passes on the new test path alone, but its own ceiling note says a touched test path is not per-finding attribution, and that test covers exactly one of the five findings — so the other four are recorded individually rather than letting one file blanket all five. The `ci.yml` entry now records that its original reasoning ("the step's own failure is the test") was tested by reality and held, in the most literal way available.

No `.opam` file was committed for any reason other than the intended constraint change; the deliberate-exception path you offered was not needed.

PR #400, opened 2026-07-31
fix(interp,native,ppx,ptx): a shared array of a record aliased every slot on one backend, refused on another, and did not compile on two more

mathiasbourgoin:
backlog-206. `let%shared (s : tri) = 4l` then `s.(i).a <- e`. The item filed it as a Native/Interpreter divergence. Measured on 9 devices before touching anything, it was four answers, none of them right, from three unrelated causes — and one of those causes is not a shared-memory bug at all.

## Measured before

4 threads, `if tid = 0l then s.(tid).a <- 7.0`, want `7 0 0 0`:

| Backend | Before |
|---|---|
| Interpreter ×2 | **raised** `Expected record, got: assignment target of .a (got unit)` |
| Native | `7 7 7 7` — accepted the store and wrote **every** slot |
| CUDA/PTX ×2 (ZLUDA) | **raised** `PTX codegen: unsupported construct: btype of custom type` — names no array, no type, no state space |
| OpenCL ×2 | device compiler: `error: unknown type name 'Test_shared_record_slots_tri'` |
| Vulkan ×2 | glslang: syntax error at the shared declaration |

## Make it work, not refuse it — and the evidence that decides it

The brief's fork was make-it-work versus a loud located refusal. The measurement that settles it is one the item does not mention: **the same kernel written with a whole-slot store (`s.(tid) <- {a = ...}`) compiled and ran correctly on 7 of the 9 devices throughout the defect.** A shared array of a record is a working, exercised feature; only the *field-store* shape was broken. A blanket refusal at the `let%shared` boundary would have deleted working functionality on seven devices — the narrowing correction overshooting, which is how this repo's dominant defect recurs.

That whole-slot result is also the tell for cause (3). The record literal is what registered the type; without it nothing did.

CUDA/PTX keeps its refusal, because that one is a real emitter constraint rather than unimplemented work: PTX has no struct type, and `Sarek_ir_ptx_mem.emit_agg_elem_addr` already refuses state-space aggregates on purpose ("aggregate elements are supported in global vectors only"). What changed there is that the message now names the array, the element type and the state space, and the test asserts the refusal on content instead of tolerating any exception.

## The three causes

**1. Native — the aliasing the item described.** `alloc_shared_with_key` filled the array with `Array.make size default`, so every slot of a boxed element type was the same allocation. It now takes a per-slot **thunk** and calls `Array.init`. A thunk rather than a value because nothing in the runtime can copy an `'a` — the PPX is the only place that knows how to build one, and passing a value puts the aliasing back whatever the allocator does with it. The identical `Array.make` sits one line away in the `create_array n Local` path and is fixed with it.

**2. Interpreter — no record to store into.** `EArrayCreate` / `DShared` mapped a record element type to `VUnit`. `Sarek_ir_interp_value.alloc_kernel_array` now builds a zeroed `VRecord` (or a constructor-0 `VVariant`) per slot. That replaces three identical copies of the old init table plus a fourth, narrower one on `Sarek_ir_interp`'s `DShared` path, which mapped only `TInt32`/`TFloat32`; unifying them also widens that path's `TInt64`/`TFloat64`/`TBool`/`TFloat16`/`TUint8` from `VUnit` to typed zeros. That is a fix, but it is outside this shape, so it is stated in CHANGES rather than folded into the word "copies".

**3. OpenCL and Vulkan — a type-collection gap, not a shared-memory gap.** `register_types_from_typ` ran over PARAMETER types only. In a kernel whose record appears nowhere but the shared declaration, nothing put it in `state.types`, so the emitted source said `__local Test_..._tri s[4];` with no such struct above it. The traversal is now top-level and runs at both kernel-array declaration sites.

Variants are deliberately **not** covered by (3), and the code says where the seam is: `TEConstr` registers the type, so the common shape works, but copying a variant in from a vector parameter registers nothing and would still fail on the struct-emitting backends. That shape was reasoned about, not executed, and it is left open rather than fixed blind — a `TVariant` arm also changes what is emitted for ordinary variant parameters, which nothing here measures.

## After — 27 device rows, plus 3 with no device at all

`sarek/tests/e2e/test_shared_record_slots.ml`, 9 devices × 3 shapes (field store, whole-slot store, `create_array n Local`), exit 0:

```
deviceless: kern_types = [Test_shared_record_slots.tri]
deviceless: PTX refused, named
deviceless: OK (no device needed for any of the above)
shared-record field-store [Interpreter] ×2: OK
shared-record field-store [Native]: OK
shared-record field-store [CUDA/PTX] ×2 (ZLUDA): refused, named (expected)
shared-record field-store [OpenCL] ×2: OK
shared-record field-store [Vulkan] ×2: OK
... slot-store ×9, local-array ×9 ...
9 device(s) exercised (frameworks: CUDA/PTX, Interpreter, Native, OpenCL, Vulkan)
```

**A slot is never read unless something wrote it.** An earlier draft asserted `0` for the untouched slots and read back `[7; 0.973611; 1210; 1]` on an RX 7900 XTX under OpenCL — correct hardware behaviour failing a wrong assertion. `__local`/`shared`/`threadgroup` memory is uninitialised on every device. The CPU backends do zero their slots; that is an implementation detail and is deliberately not asserted.

**CI has no GPU, so the two backend-only halves are also checked with no device**, first thing, on the same kernel the device loop runs: `kern_types` contains the record type, the OpenCL and GLSL text introduce the struct above the line that declares the array, and the PTX generator raises a message naming the array and the type. Without this, reverting either half left CI green.

## Prove-red — 10 mutations, each asserted to have landed

Committed first, then mutated; `git diff --numstat` checked non-empty every time (one restore silently reverted uncommitted work, which is why the counts are checked rather than assumed). Each mutation keeps the function and breaks the wiring or one arm, so nothing reads red because it failed to link.

| # | Mutation | Went red as |
|---|---|---|
| 1 | `Array.init` → `Array.make` in `alloc_shared_with_key` | `test_cpu_runtime` exit 1, `FAIL slot 1 untouched: Expected 0, Received 7`; e2e Native `@0 own slot .a: got 3 want 0` (all four outputs equal) |
| 2 | `Array.init` → `Array.make` in `alloc_kernel_array` | interp unit exit 1, `slots are independent allocations`; e2e Interpreter ×2 `got 3 want 0` |
| 3 | record default back to `VUnit` | `a record element type must produce a VRecord, got VUnit (the pre-fix behaviour)`; e2e `Expected record, got: assignment target of .a (got unit)` |
| 4 | unwire `register_types_from_typ` at the `let%shared` site | OpenCL ×2 + Vulkan ×2 `unknown type name 'Test_shared_record_slots_tri'` |
| 5 | unwire the PTX named refusal | `refused, but the message does not name the shared array and its record type. Got: ... btype of custom type` |
| 6 | `Array.init` → `Array.make` in `create_array n Local` | Native `@0 local slot 0 .a: got 10 want 0` — slots 0 and 1 aliased |
| 7 | PTX local site passes `~what:"shared"` | `refused the local array ... does not name the local array and its record type` — proves the predicate distinguishes the two state spaces rather than accepting either |
| 8 | (4) again, **with no GPU on the path** | deviceless `kern_types = []` |
| 9 | (8) after fixing the ordering check | deviceless, 3 failures: `kern_types`, OpenCL and GLSL |
| 10 | (5) **with no GPU on the path** | deviceless `PTX refused but the message names neither the shared array nor its record type` |

Mutation 8 is why 9 exists. The first version of the source-ordering check **could not fail**: it compared the first occurrence of the type name against the offset of `s[4]`, and the declaration line is `__local Test_..._tri s[4];`, so the name is on that line whether or not a struct was ever defined. Measured — with the registration reverted the `kern_types` assertion went red and the ordering one stayed green on both dialects. The anchor is now the start of the declaration line.

## Two review passes, 15 findings, all this repo's dominant class

A cross-runtime pass (codex) and an unscoped fresh read. Every finding was a claim wider than its code; the fixes are commits 3 and 4. The substantive ones:

- The tolerance comparisons **accepted NaN** — `Float.abs (got -. want) > 1e-4` is false for NaN, and NaN is exactly what a read of uninitialised device memory returns, which is the failure mode this file exists to catch. One `close_enough` with an `is_finite` guard now carries all five sites.
- The local-array and slot-store cases **accepted any CUDA/PTX exception**, so the named refusal could regress to the old anonymous one and stay green.
- The header said a Metal/HIP/WGSL failure "is not tolerated". It cannot be: this file initialises five plugins and WGSL is not on the `Device.init` list, so no such device can enumerate here on any host.
- `expects_refusal` was described as "asserted below" against a framework "starting with CUDA". Neither: the list prints, and the match is exact equality on `"CUDA/PTX"`.
- The `.mli` said the thunk runs "once per slot"; a cache hit returns the stored array and never runs it.
- The PTX message claimed the same kernel runs on four other backends, which is not guaranteed for the unregistered-variant shape above.
- `TUint8 -> VInt32 0l` is a guess, not a convention — `TUint8` occurs nowhere else in the interpreter — and the docstring's silence read as "nothing to say".
- CHANGES gave Vulkan OpenCL's error message, and two `kb/research` line references drifted when the allocator moved.

## Gates

Project switch `/home/mathias/dev/SPOC/_opam`, dune 3.24.1, OCaml 5.4.0, ppxlib 0.38.0. Exit codes captured directly, never a prose PASS.

| Gate | Exit |
|---|---|
| `dune build` | 0 |
| `dune build @fmt` | 0 |
| `make test_negative` | 0 |
| `scripts/check-opam-clean.sh` | **1** |

`check-opam-clean` is **pre-existing and unchanged by this branch**, verified rather than assumed: a detached worktree at `origin/main` (97a062a2) gives the identical **1** under the project switch (`FAIL: sarek.opam differs from the tracked version after 'make opam'`) and **2** under ambient dune 3.23.0, where the tree no longer builds at all (`Ptyp_open` / ppxlib 0.35 — the floor #398 raised). No `.opam` is committed here.

No new gate script, so no new `kb/properties.md` `gate-red-path` row is required.

## Not verified

- **Metal, HIP, WGSL, and NVIDIA hardware.** None exist on this host. WGSL is not even on this file's `Device.init` list and the Metal/HIP plugins are not initialised by it, so those legs cannot run here on any host — the header says so rather than promising non-tolerance. They get source-level coverage only: `Sarek_ir_metal` and `Sarek_ir_wgsl` were read, not executed.
- **The unregistered-variant shape** (`s.(tid) <- v.(tid)` from a variant-element vector parameter). Reasoned about from the code, not executed. Left open and documented at the seam.
- **opencode.** On this host it is on `PATH` but non-functional: the review invocation produced zero bytes in 25 minutes, and a 120-second smoke test with a one-line prompt also timed out with no output. The cross-runtime pass was codex only, plus an unscoped fresh read. Recorded as `GO codex,claude`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed shared record-array allocation and initialization to prevent element aliasing across slots.
  * Improved field-store handling for record types in arrays.
  * Fixed vector record-field stores and record detachment issues.

* **New Features**
  * PTX backend now provides contextual error messages for unsupported aggregate shared-memory accesses, naming the memory space, array, and element type.

* **Tests**
  * Added comprehensive end-to-end tests for shared and local record arrays across backends.
  * Added unit tests validating shared array allocation and initialization.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -132,8 +132,9 @@
   `assignment target of .a (got unit)`; Native accepted the store and wrote
   EVERY slot (`7 7 7 7`, want `7 0 0 0`); CUDA/PTX ×2 raised
   `unsupported construct: btype of custom type`, naming neither the array nor
-  the type; and OpenCL ×2 and Vulkan ×2 failed inside the DEVICE compiler with
-  `unknown type name`. Three separate causes. (1) Native's
+  the type; and OpenCL ×2 and Vulkan ×2 failed inside the DEVICE compiler,
+  OpenCL with `unknown type name 'Test_..._tri'` and Vulkan with a glslang
+  syntax error at the shared declaration. Three separate causes. (1) Native's
   `alloc_shared_with_key` filled the array with `Array.make size default`, so
   every slot of a boxed element type was the same allocation — it now takes a
   per-slot thunk and calls `Array.init`, and the identical `Array.make` in the
@@ -141,8 +142,14 @@
   record element type to `VUnit`, so there was nothing to store into — it now
   builds a zeroed `VRecord` — or, for a variant with at least one
   constructor, a constructor-0 `VVariant` — per slot in
-  `Sarek_ir_interp_value.alloc_kernel_array`, which replaces four copies of the
-  old init table. (3) OpenCL and Vulkan had no shared-memory gap at all:
+  `Sarek_ir_interp_value.alloc_kernel_array`. That replaces three identical
+  copies of the old init table in `Sarek_ir_interp_eval` plus a fourth, NARROWER
+  one on `Sarek_ir_interp`'s `DShared` kernel-parameter path, which mapped only
+  `TInt32` and `TFloat32` and everything else to `VUnit`. Unifying them
+  therefore also widens that path: `TInt64`, `TFloat64`, `TBool`, `TFloat16` and
+  `TUint8` go from `VUnit` to their typed zeros there. That is a fix, but it is
+  a behaviour change outside backlog-206's shape and it is stated rather than
+  folded into the word "copies". (3) OpenCL and Vulkan had no shared-memory gap at all:
   `register_types_from_typ` ran over PARAMETER types only, so a record named
   nowhere but the shared declaration was never emitted as a `struct`; it is now
   top-level and runs at both kernel-array declaration sites. The tell was that
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -126,6 +126,36 @@
 
 ### Fixed
 
+- A shared-memory array of a record type behaved four different ways across
+  nine devices, none of them right. `let%shared (s : tri) = 4l` followed by
+  `s.(i).a <- e`, measured on this host: the Interpreter ×2 raised
+  `assignment target of .a (got unit)`; Native accepted the store and wrote
+  EVERY slot (`7 7 7 7`, want `7 0 0 0`); CUDA/PTX ×2 raised
+  `unsupported construct: btype of custom type`, naming neither the array nor
+  the type; and OpenCL ×2 and Vulkan ×2 failed inside the DEVICE compiler with
+  `unknown type name`. Three separate causes. (1) Native's
+  `alloc_shared_with_key` filled the array with `Array.make size default`, so
+  every slot of a boxed element type was the same allocation — it now takes a
+  per-slot thunk and calls `Array.init`, and the identical `Array.make` in the
+  `create_array n Local` path is fixed with it. (2) The interpreter mapped a
+  record element type to `VUnit`, so there was nothing to store into — it now
+  builds a zeroed `VRecord` (or a constructor-0 `VVariant`) per slot in
+  `Sarek_ir_interp_value.alloc_kernel_array`, which replaces four copies of the
+  old init table. (3) OpenCL and Vulkan had no shared-memory gap at all:
+  `register_types_from_typ` ran over PARAMETER types only, so a record named
+  nowhere but the shared declaration was never emitted as a `struct`; it is now
+  top-level and runs at both kernel-array declaration sites. The tell was that
+  the same kernel written with a whole-slot store (`s.(tid) <- {...}`) compiled
+  and ran correctly on all seven non-PTX devices throughout, because the record
+  literal registered the type — so a whole-slot-store test would have been green
+  the whole time, and `test_shared_record_slots.ml` exercises both shapes.
+  CUDA/PTX still refuses, and the test asserts the refusal rather than
+  tolerating it: PTX has no struct type and
+  `Sarek_ir_ptx_mem.emit_agg_elem_addr` refuses state-space aggregates on
+  purpose. What changed there is that the message names the array, the element
+  type and the state space. Metal, HIP and WGSL were not measured — no such
+  device on this machine — and nothing is claimed about them.
+
 - A `[@@sarek.type]` record with a record-typed field had its inner struct
   emitted AFTER the struct that referenced it, so no kernel touching a nested
   record compiled on any shader backend. Both declaration-emission loops walked
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -139,7 +139,8 @@
   per-slot thunk and calls `Array.init`, and the identical `Array.make` in the
   `create_array n Local` path is fixed with it. (2) The interpreter mapped a
   record element type to `VUnit`, so there was nothing to store into — it now
-  builds a zeroed `VRecord` (or a constructor-0 `VVariant`) per slot in
+  builds a zeroed `VRecord` — or, for a variant with at least one
+  constructor, a constructor-0 `VVariant` — per slot in
   `Sarek_ir_interp_value.alloc_kernel_array`, which replaces four copies of the
   old init table. (3) OpenCL and Vulkan had no shared-memory gap at all:
   `register_types_from_typ` ran over PARAMETER types only, so a record named
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -140,9 +140,16 @@
   per-slot thunk and calls `Array.init`, and the identical `Array.make` in the
   `create_array n Local` path is fixed with it. (2) The interpreter mapped a
   record element type to `VUnit`, so there was nothing to store into — it now
-  builds a zeroed `VRecord` — or, for a variant with at least one
-  constructor, a constructor-0 `VVariant` — per slot in
-  `Sarek_ir_interp_value.alloc_kernel_array`. That replaces three identical
+  builds a zeroed `VRecord` — or, for a variant, a `VVariant` carrying the
+  first NULLARY constructor (the first constructor with zeroed payloads if
+  there is no nullary one), which is the same constructor Native's
+  `default_value_for_type` puts in that slot — per slot in
+  `Sarek_ir_interp_value.alloc_kernel_array`. The tag is
+  `Hashtbl.hash ctor mod 256`, the encoding `EVariant` and the two matchers
+  already use and which is now the single `variant_tag_of_ctor`; a positional
+  index there produced a `VVariant` no arm could match, so reading a default
+  variant slot raised `Pattern match failure in SMatch` on the Interpreter
+  while Native answered the nullary constructor. That replaces three identical
   copies of the old init table in `Sarek_ir_interp_eval` plus a fourth, NARROWER
   one on `Sarek_ir_interp`'s `DShared` kernel-parameter path, which mapped only
   `TInt32` and `TFloat32` and everything else to `VUnit`. Unifying them
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -141,9 +141,10 @@
   `create_array n Local` path is fixed with it. (2) The interpreter mapped a
   record element type to `VUnit`, so there was nothing to store into — it now
   builds a zeroed `VRecord` — or, for a variant, a `VVariant` carrying the
-  first NULLARY constructor (the first constructor with zeroed payloads if
-  there is no nullary one), which is the same constructor Native's
-  `default_value_for_type` puts in that slot — per slot in
+  first NULLARY constructor, the same one Native's `default_value_for_type`
+  picks, falling back to the first constructor with zeroed payloads where
+  there is no nullary one (a case in which Native has no default at all and
+  raises, so the interpreter is more defined there rather than agreeing) — per slot in
   `Sarek_ir_interp_value.alloc_kernel_array`. The tag is
   `Hashtbl.hash ctor mod 256`, the encoding `EVariant` and the two matchers
   already use and which is now the single `variant_tag_of_ctor`; a positional
--- a/kb/research/obj-usage/README.md
+++ b/kb/research/obj-usage/README.md
@@ -34,7 +34,7 @@ No active `Obj.field`, `Obj.set_field`, `Obj.tag`, `Obj.size`, or related low-le
 - `sarek/plugins/interpreter/Interpreter_plugin.ml:93`: `Obj.magic` from `EXEC_VECTOR` to `Kernel_arg.Vec`.
 - `sarek/plugins/native/Native_plugin.ml:264-269`: `Obj.obj`, `Obj.repr`, and `Obj.obj` for custom vector access.
 - `sarek/sarek/Kirc_kernel.ml:251-256`: same custom vector access bridge.
-- `sarek/sarek/Sarek_cpu_runtime_types.ml:126` (`alloc_shared_with_key`): former `Obj.obj (Obj.repr arr)` in generic shared-memory custom arrays — now typed via `Type_id.Refl` (resolved; module split out of `Sarek_cpu_runtime.ml`).
+- `sarek/sarek/Sarek_cpu_runtime_types.ml:151` (`alloc_shared_with_key`): former `Obj.obj (Obj.repr arr)` in generic shared-memory custom arrays — now typed via `Type_id.Refl` (resolved; module split out of `Sarek_cpu_runtime.ml`).
 - `sarek/sarek/Sarek_type_helpers.ml:65-66`: `Obj.obj (Obj.repr ...)` in untyped helper dispatch.
 - `sarek/sarek/Sarek_ir_interp.ml:239` (`vector_to_array`), `274` (`array_to_vector`): former fallback vector element casts — no `Obj` present now (interpreter module split; bridge stayed in main).
 - `sarek/plugins/native/Native_plugin.ml:384-385`: legacy `Obj.t array` direct execution type.
--- a/kb/research/obj-usage/shared-memory-type-ids.md
+++ b/kb/research/obj-usage/shared-memory-type-ids.md
@@ -9,9 +9,9 @@ repointed to current locations and retained for history.
 
 ## Obj Sites (former)
 
-- `sarek/sarek/Sarek_cpu_runtime_types.ml:126` (`alloc_shared_with_key`): the former
+- `sarek/sarek/Sarek_cpu_runtime_types.ml:151` (`alloc_shared_with_key`): the former
   `Obj.obj (Obj.repr arr)` cast is gone — now a typed `Type_id.equal`/`Refl` match.
-- `sarek/sarek/Sarek_cpu_runtime.mli:84`: now documents typed custom allocation (no
+- `sarek/sarek/Sarek_cpu_runtime.mli:101`: now documents typed custom allocation (no
   `Obj.t`).
 
 ## Assumed Invariant
--- a/sarek/codegen/Sarek_ir_ptx_stmt.ml
+++ b/sarek/codegen/Sarek_ir_ptx_stmt.ml
@@ -42,12 +42,10 @@ let refuse_aggregate_state_space_array ~what ~(name : string) (elt : elttype) :
             PTX has no struct type and this backend addresses aggregates only \
             in global vectors, so an aggregate %s array cannot be declared. \
             Pass the data as a vector parameter, or declare one scalar %s \
-            array per field. (The same kernel runs on the Interpreter, Native, \
-            OpenCL and Vulkan backends, which do have an aggregate %s array.)"
+            array per field."
            what
            name
            tyname
-           what
            what
            what)
   | TInt32 | TInt64 | TFloat16 | TFloat32 | TFloat64 | TUint8 | TBool | TUnit
--- a/sarek/codegen/Sarek_ir_ptx_stmt.ml
+++ b/sarek/codegen/Sarek_ir_ptx_stmt.ml
@@ -13,6 +13,47 @@ open Sarek_ir_ptx_types
 open Sarek_ir_ptx_mem
 open Sarek_ir_ptx_expr
 
+(** Refuse a [.shared]/[.local] array whose element type is a record or a
+    variant, BY NAME (backlog-206).
+
+    The refusal itself is not new and is not being lifted here: PTX has no
+    struct type, so an aggregate element would need byte-offset addressing in
+    the state space, which {!Sarek_ir_ptx_mem.emit_agg_elem_addr} refuses on
+    purpose ("aggregate elements are supported in global vectors only"). What
+    was wrong is what the user saw. The declaration path reached
+    [ptx_btype_of_elttype], whose refusal read "PTX codegen: unsupported
+    construct: btype of custom type" — no array, no type, no mention of shared
+    memory, and identical for the two unrelated call sites that reach it.
+    Measured verbatim on two ZLUDA devices for [let%shared (s : tri) = 4l].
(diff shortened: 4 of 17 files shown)


mathiasbourgoin:
## Independent adversarial review — GO, with two defects fixed on the branch

I did not write this code. Everything below was executed on this host, on a worktree of my own (`/mnt/ssd-external-2to/wt-review-400`), against the pre-fix tree as well as the head. Toolchain for every row: project switch `/home/mathias/dev/SPOC/_opam`, dune 3.24.1, OCaml 5.4.0, ppxlib 0.38.0.

Two commits pushed: `977bf5cd` (the defect) and `c65cd7ef` (four claims wider than their code — in my own fix, found by the cross-runtime pass).

---

## 1. The load-bearing judgement holds. I re-executed it on the pre-fix tree.

The make-it-work-not-refuse-it decision rests on one measurement the item does not mention: that the whole-slot store worked on 7 of 9 devices throughout the defect. I did not take that on trust. I built the PR's own test file **on `origin/main` (97a062a2)**, with only `check_deviceless` disabled so the run reaches the device loop, and ran it under ZLUDA:

```
shared-record slot-store  [Interpreter] x2:                       OK
shared-record slot-store  [Native]:                               OK
shared-record slot-store  [CUDA/PTX] x2 (ZLUDA):                  refused (anonymous)
shared-record slot-store  [OpenCL]  x2:                           OK
shared-record slot-store  [Vulkan]  x2:                           OK
```

**Confirmed: 7 of 9.** A blanket refusal at `let%shared` would have deleted working functionality on seven devices. The refusal option was correctly discarded.

The same run reproduces the whole BEFORE table, independently, verbatim:

| Backend | pre-fix, my run |
|---|---|
| Interpreter ×2 | `Expected record, got: assignment target of .a (got unit)` |
| Native | `@0 own slot .a: got 3 want 0` — all four outputs equal, the aliasing |
| CUDA/PTX ×2 | `unsupported construct: btype of custom type` |
| OpenCL ×2 | `error: unknown type name 'Test_pre206_probe_tri'` |
| Vulkan ×2 | glslang `syntax error, unexpected IDENTIFIER` at the shared declaration |

And the pre-fix local-array row is worth recording, because it is *not* symmetric with the shared one: OpenCL ×2 and Vulkan ×2 were **OK** pre-fix for `create_array n Local`, because that kernel's `arr.(2l) <- {...}` literal registers the type. Only the Interpreter and Native were broken there. Consistent with cause (3) being a type-collection gap.

**The converse — does the fix now accept anything it should refuse?** No, and this is structural rather than measured: `refuse_aggregate_state_space_array` refuses exactly `TRecord | TVariant`, and `ptx_btype_of_elttype` (Sarek_ir_ptx_types.ml:434) already refused exactly `TRecord | TVariant` on the same path. The accept/refuse partition is byte-identical; only the message changed. The underlying constraint is real and pre-existing: `Sarek_ir_ptx_mem.emit_agg_elem_addr:171` refuses `space <> None` on purpose, and that file is not in this diff. The refusal does name all three things — measured on a record I invented:

```
shared-memory array 's' has element type 'Test_rev400_probe.outer', a record or
variant. PTX has no struct type and this backend addresses aggregates only in
global vectors, ...
```

---

## 2. Defect: a default variant slot that no arm could match. Fixed.

**The interpreter's variant tag is not a positional index.** `EVariant` encodes `Hashtbl.hash ctor mod 256` (`Sarek_ir_interp_eval.ml:176`) and `EMatch`/`SMatch` select an arm with exactly that predicate (`:200`, `:386`). This branch's new `default_value_of_elttype` stored a literal `0`, so a default variant slot matched **no arm at all**.

Executed, `type c2 = A of float32 | B` with `[@@sarek.type]`, the type registered by a `TEConstr` so the registration gap below is not in play, reading a slot no thread wrote:

```
before:  [Interpreter] RAISED Pattern match failure in SMatch
         [Interpreter] RAISED Pattern match failure in SMatch
         [Native]      2 2 2 2      (= B)
after:   [Interpreter] 2 2 2 2
         [Interpreter] 2 2 2 2
         [Native]      2 2 2 2
```

Three things were wrong at once, which is why it survived a codex pass and an unscoped fresh read:

- the value (a literal `0`);
- the docstring, which said the slot gets "its FIRST constructor" and that `0` "is what the zeroed tag field of zeroed storage decodes to" — neither is true of this representation;
- **the new unit test, which pinned the bug.** `check int "tag" 0 tag` asserts the wrong convention, and it never runs the value through a matcher, so it could not fail. A gate that cannot fail in the direction that matters.

Fixed in `977bf5cd`:

- the tag goes through a single `variant_tag_of_ctor`, which `Sarek_ir_interp_eval`'s three sites now call as well — they were three copies of the expression, which is precisely how a fourth copy went unnoticed;
- the constructor chosen is the first **nullary** one, which is what `Sarek_native_helpers.default_value_for_type` picks, so the two CPU backends put the same constructor in a freshly declared shared array. A CPU-backend disagreement about that is the exact shape of divergence backlog-206 was filed as;
- the test now asserts the **matcher's predicate**, in both polarities (the tag must select the chosen arm and must *not* select the other), plus the nullary preference and the no-nullary fallback.

Prove-red, committed first, each mutation asserted to have landed with `git diff --numstat`:

| Mutation | Went red as |
|---|---|
| tag back to literal `0` | `test_interp_shared_array_slots` exit **1**, 2 failures, `Received: 'false'` in the matchable case |
| drop the nullary preference (`let nullary = None`) | exit **1**, `FAIL B, the nullary constructor, not A which comes first` |

---

## 3. Defect: the seam comment mislocated the variant gap. Fixed.

The branch documents the unverified shape rather than fixing it blind, which is right. I executed it. It is broken — the reasoning was correct about the outcome — but **the comment locates the seam in the wrong place, and the stated reason for deferring is wrong.**

`s.(tid) <- v.(tid)` from a `color vector`, shared array: `kern_types = []`, OpenCL ×2 `unknown type name 'Test_rev400_probe_color'`, Vulkan ×2 glslang syntax error. As predicted.

Then I removed the shared array entirely — a kernel whose only variant occurrence is the vector **parameter**, no `let%shared` anywhere, no `TEConstr`:

```ocaml
fun (v : color vector) (out : float32 vector) ->
  match v.(tid) with Red -> out.(tid) <- 0.0 | Value x -> out.(tid) <- x
```

```
kern_types = []
[OpenCL] x2: unknown type name 'Test_rev400_probe_color'   <- on the PARAMETER declaration
[Vulkan] x2: glslang syntax error
```

So the gap is **variant registration in general**, not a shared-memory seam. The shared-array shape is one instance of it. And the deferral rationale — "adding a `TVariant` arm changes what gets emitted for ordinary variant PARAMETERS too, and this change has no measurement for that" — is wrong where it matters: that kernel does not compile, so there is nothing working to regress. (Add a `TEConstr` and it compiles and runs; the emitter needs `kern_variants`, which only `TEConstr` fills.)

Two *real* obstacles, both measured, are named in the comment instead:

- this function writes `state.types`, the **record** table (`→ kern_types`), while a variant belongs in `state.variants` (`→ kern_variants`, Sarek_lower_ir.ml:1672/1687). A naive `TVariant` arm here would declare a variant as a record struct;
- registration alone is **not sufficient**. With the type unregistered the emitter also drops the match-arm payload binder and emits, in the unregistered case, `out[tid] = x;` with no `x` declared — versus `float x = c.data.Value_v;` when a `TEConstr` is present. The binder comes from `kern_variants` too.

---

## 4. Uninitialised memory and NaN: clean.

Every slot the file reads is written first. `k`: all four threads write `.a` and `.b` of their own slot, then read own `.a` and neighbour `.b` across the barrier. `k_slot`: all four slots written whole. `k_local`: slots 0 and 1 field-stored before either is read; slot 2 is written and never read. `.c` is never written and never read.

Tolerance: all five value comparisons in the file go through `close_enough`, which is `Float.is_finite got && Float.abs (got -. want) <= 1e-4`. There is no surviving bare `Float.abs (…) > 1e-4` a NaN would satisfy — grepped, `Float.abs` occurs exactly once, inside `close_enough`.

---

## 5. The vacuous gate: I looked for the second one, by mutation.

The near-miss was a source-ordering check that compared the type name against the offset of `s[4]`, on a line that contains the type name. It is now anchored to the start of the declaration line. Reverting the registration proves it fails when the struct is *absent*; it does not prove it is sensitive to **ordering**, which is the property claimed. So I mutated the emitter instead: moved `gen_c_type_decls` from before the kernel signature to after the closing brace, **in the OpenCL backend only**. The struct is then present, and after its use.

```
deviceless: kern_types = [Test_shared_record_slots.tri]     <- still green
DEVICELESS FAILED: OpenCL uses the type "Test_shared_record_slots_tri" on the
  shared-array declaration line but nothing above that line introduces it
(GLSL: green — only the mutated dialect fired)
exit 1
```

That is the decisive result: **presence-only would have stayed green, and only the OpenCL row fired.** The vacuity fix is real and the check is genuinely ordering-sensitive. Full mutation ledger, all `git diff --numstat`-verified:

| # | Mutation | Gate | Exit | Went red as |
|---|---|---|---|---|
| 1 | unwire `register_types_from_typ` at the `let%shared` site | e2e | 1 | 3 deviceless failures: `kern_types = []`, OpenCL, GLSL |
| 2 | **struct emitted AFTER the kernel, OpenCL only** | e2e | 1 | OpenCL ordering only; `kern_types` and GLSL green |
| 3 | `Array.init` → `Array.make` in `alloc_shared_with_key` | unit + e2e | 1 / 1 | `FAIL slot 1 untouched`; Native `@0 own slot .a: got 3 want 0` |
| 4 | `Array.init` → `Array.make` in `alloc_kernel_array` | unit + e2e | 1 / 1 | `FAIL slot 1 untouched`; Interpreter `got 3 want 0` |
| 5 | record default back to `VUnit` | unit + e2e | 1 / 1 | `must produce a VRecord, got VUnit (the pre-fix behaviour)`; `assignment target of .a (got unit)` |
| 6 | unwire the PTX named refusal | e2e | 1 | `PTX refused but the message names neither the shared array nor its record type: … btype of custom type` |
| 7 | `Array.init` → `Array.make` in `create_array n Local` | e2e | 1 | Native `@0 local slot 0 .a: got 10 want 0` |
| 8 | tag back to literal `0` (my fix) | unit | 1 | matchable case, `Received: 'false'` |
| 9 | drop the nullary preference (my fix) | unit | 1 | `FAIL B, the nullary constructor, not A which comes first` |

---

## 6. `Array.init` per-slot thunk: independent, including one level down.

Constructed the case that would catch a re-alias one level down — a record with a nested *mutable* record component, `outer = {mutable lo : inner; mutable hi : float32}`, `inner = {mutable p; mutable q}` — and stored through the nested field, `s.(tid).lo.p <- src.(tid)`, both in shared memory and in `create_array n Local`:

```
nested-shared [Interpreter] x2 OK   [Native] OK   [OpenCL] x2 OK   [Vulkan] x2 OK
nested-local  [Interpreter] x2 OK   [Native] OK   [OpenCL] x2 OK   [Vulkan] x2 OK
              [CUDA/PTX] x2 refused, naming 'Test_rev400_probe.outer' (both state spaces)
```

The thunk is genuinely re-evaluated: `default_value_for_type` builds a nested record *literal*, so each call allocates the whole tree fresh; nothing closes over a shared mutable value. The interpreter side recurses through `default_value_of_elttype` per field, per slot.

---

## 7. Type registration: no double-registration, and the record path really is reached only through the shared array.

`register_types_from_typ` guards with `if not (Hashtbl.mem state.types name)`, and `register_tuple_type` does the same, so running it at three sites (parameters, `TECreateArray`, `TELetShared`) cannot duplicate an entry. The deviceless assertion is the direct evidence that a type reachable *only* through the shared declaration is now registered on the struct-emitting backends: `kern_types = [Test_shared_record_slots.tri]` for a kernel with no parameter of that type and no record literal, and `[]` under mutation 1.

**One claim I could not reproduce.** The `TECreateArray` comment says a `create_array n Local` with a primitive-component tuple element type now registers the synthesized `_tup_*` record and "the emitted kernel gains an unreferenced struct typedef it did not have before … inert output". I built that shape and diffed the generated OpenCL against the pre-fix tree: **byte-identical**, typedef present on both, and *referenced* (`(_tup_float32_float32){…}`) rather than unreferenced — the tuple write registers the type anyway. So the predicted change did not occur in the only shape I could express. Not a regression (the whole shape is broken identically on both trees — `int arr[4]` from the `elttype_of_typ` collapse), but the sentence describes a change I cannot make happen. Worth narrowing or dropping.

---

## 8. Coverage boundary: the header states exactly what it can.

The file initialises Native, Interpreter, CUDA, OpenCL, Vulkan. `Device.init` names `Metal` and `HIP` — whose plugins are not linked or initialised here — and does **not** name WGSL. So no Metal, HIP or WGSL device can enumerate on any host, and the header says that rather than promising non-tolerance. Checked the correction did not overshoot the other way: the NOT-MEASURED-HERE list keys are `Interpreter`, `Native`, `CUDA/PTX`, `OpenCL`, `Vulkan` — backend names, matching `Device.framework`, and it omits Metal/HIP/WGSL consistently with the header. `expects_refusal` is exact equality on `"CUDA/PTX"`, not `String.starts_with` on the family name; the family/backend confusion that cost two PRs today is not present here.

`kb/research/obj-usage/` citations: `Sarek_cpu_runtime_types.ml:151` is `let alloc_shared_with_key`, `Sarek_cpu_runtime.mli:101` is `val alloc_shared_with_key`. Both land.

**Nit, not fixed.** `ptx_refusal_is_named` matches `has "tri"` — a three-character substring of the type name rather than the type name. It happens not to collide with anything else in the message, but `"tri"` occurs in plenty of English words a future message might use ("distribution", "electric"). `has record_type_name` or the mangled tag would be the same assertion with none of the slack.

---

## 9. Gates

Project switch `/home/mathias/dev/SPOC/_opam`, dune 3.24.1, OCaml 5.4.0, ppxlib 0.38.0. Each exit code captured directly.

| Gate | Exit |
|---|---|
| `dune build` | 0 |
| `dune build @fmt` | 0 |
| `make test_negative` | 0 |
| `scripts/prove-red.test.sh` | 0 |
| `scripts/check-opam-clean.sh` | **1** |
| `test_shared_record_slots.exe` under ZLUDA, 9 devices × 3 shapes + 3 deviceless | 0 |

`check-opam-clean` **1** is pre-existing, and I verified it myself rather than accepting the PR's word: a separate worktree at `origin/main` (97a062a2), untouched, gives the identical **1** under the project switch (`FAIL: sarek.opam differs from the tracked version after 'make opam'`) and **2** under ambient dune 3.23.0 / OCaml 5.3.0 / ppxlib 0.35.0, where the tree does not build at all (`There is no constructor "Ptyp_open" within type "core_type_desc"`). No `.opam` is committed on this branch.

The two ocamlformat doc warnings now printed for `Sarek_ir_interp_value.ml` (`'{v … v}' should begin on its own line`, lines 246–298) are pre-existing, in `detach_record`'s untouched docstring — confirmed by running `ocamlformat` directly on `git show HEAD:` of that file, same two warnings, exit 0.

---

## 10. My own fix got the same treatment, and needed it

A codex pass over `977bf5cd` found four claims wider than their code — in the fix for a claim wider than its code. All four corrected in `c65cd7ef`:

- "THE definition, and the only one" of the tag was false: `Sarek_ppx.ml:1535` emits `Hashtbl.hash "C" mod 256` into generated user code. Now named as the one copy outside the interpreter, at **both** sites, and honestly described as a duplication rather than a boundary — the generated code already reaches `Sarek.Sarek_value` through a forwarding alias, so it *could* be a call.
- "The two CPU backends now agree" was too broad. It holds where a nullary constructor exists (the measured case). Where none does, Native recurses on the argument type and has no arm for a tuple, so `C of float32 * float32` as the only constructor reaches `failwith "Cannot create default value for this type"` and never initialises, while the interpreter zeroes each component. That is the interpreter being strictly more defined, not agreement — now stated in the docstring, at the test case, and in CHANGES.
- the "matchable" case claimed value and matcher "cannot drift apart". It does not execute `EMatch`/`SMatch` — no kernel runs in that file, by design. The coupling is by construction; a matcher that stopped calling the helper would leave it green. Said, with the executed evidence named for what it is.
- the nullary case said the Native agreement was "pinned". Only the interpreter side is.

---

## What I could NOT verify

- **Metal, HIP, WGSL, NVIDIA hardware.** None on this host, and the file cannot enumerate the first three on any host. Source-level only.
- **`opencode`**: on `PATH`, non-functional here (zero bytes). Not attested. The cross-runtime pass was **codex** only, plus my own read.
- **The `_tup_*` typedef claim** in the `TECreateArray` comment: measured and did not reproduce (§7). I could not construct a shape where the new registration adds a typedef that was absent before.
- **`mod 256` tag collisions.** The encoding is not injective; two constructors of one type whose names collide modulo 256 select each other's arms. Pre-existing, out of scope, now named at one site instead of four.
- **The variant registration fix itself.** I measured that it is needed and where the real obstacles are (§3), and deliberately did not attempt it: routing to `state.variants` changes what is emitted for every variant kernel and needs its own evidence, not a review's side-effect.

**Verdict: GO.** Attested `GO codex,claude` for `c65cd7ef`.

PR #402, opened 2026-07-31
fix(gates): the machine-identifier gate never looked for absolute home paths (backlog-216)

mathiasbourgoin:
# backlog-216 — the machine-identifier gate never looked for absolute home paths

## The gap, measured

`scripts/check-no-machine-identifiers.sh` exits **0** on a tree with
`/home/<user>/dev/SPOC/_opam` committed in `kb/sarek/ppx/parser.md:106`.

Proven on this branch, on one tree, by reintroducing that exact line:

| gate | exit | message |
|---|---|---|
| `origin/main` version | **0** | `OK -- 1 tracked payload(s), no machine identifier in payloads, paths or producers` |
| this branch's version | **1** | `::error::tracked file(s) contain an absolute home path naming a user account:` / `kb/sarek/ppx/parser.md:106: /home/mathias/` |

This is a **scope** gap, not a bypass like the one #389 closed. The gate's three
rules cover benchmark payload fields, benchmark result filenames, and
hostname-reading producer source. None reads prose. So the class backlog-168
existed to eliminate — personal machine identifiers in tracked files, which in
that case reached a published gh-pages site — was closed only on the three
surfaces in scope then, while the gate's *name* claimed the class.

That mismatch is the dominant defect class in this repo, so the widening ships
with its limits written down rather than implied.

---

## 1. The sweep, in full

Measured **before** any rule was written, because the rule's shape had to follow
the data and not the other way round: a gate that lands with a known backlog of
failures is a gate somebody switches off.

### Absolute home paths — 26 files (0 macOS-spelled)

**USE — fixed in this PR (20 files)**

| file | occurrences | what it was | fix |
|---|---|---|---|
| `formal/**/.{16 names}.aux` | 16 files, 1 each | Rocq `COQAUX1` side files: tracked **build artifacts**, each embedding the absolute path of the `.v` it was produced from | untracked + gitignored; the autopilot workflow that staged them no longer does |
| `docs/plans/phase2-playground-mvp-2026-06-03.md` | 4 | `opam exec --switch=<home path>` in the Quality-gates block | `--switch=.` from the repo root, with a line saying so |
| `docs/plans/ptx-spike-findings.md` | 3 | absolute paths to files inside this repo | repo-relative |
| `formal/convergence-safety/report/PHASE1A_PLAN.md` | 1 | absolute path to a `.v` in this repo | repo-relative |
| `formal/convergence-safety/formal-verif-autopilot.workflow.js` | 3 | hardcoded `SPOC`, `SKILL_ROOT`, `METHODOLOGY_BACKLOG` | env vars with `$HOME`-relative defaults |

**DESCRIBE — exempted (6 files)**

| file | root | why it is a mention, not a use |
|---|---|---|
| `Dockerfile` | `/home/opam/` | the opam base image's own account; fixed by the IMAGE, and `OPAMROOT`/`WORKDIR` must name it absolutely |
| `.github/workflows/ci.yml` | `/home/rocq/` | a comment recording *why* a `container:` job fails — the rocq image's switch belongs to its own user |
| `scripts/check-cited-paths-exist.sh` | `/home/mathias/` | the worked example in the comment documenting what that gate's regex deliberately does **not** match; the path is the sentence's subject |
| `scripts/check-cited-paths-exist.test.sh` | `/home/u/` | fixture text for that case |
| `scripts/prove-red-fixtures/dune-verbose-{command-exit,signal}-log.txt` | `/home/user/` | captured `dune --verbose` output kept byte-for-byte as a parser fixture; already anonymised at capture, and editing it would stop it being a capture |

**A seventh describe-case landed on `main` while this branch was in flight** —
`scripts/check-dune-opam-portability.sh:19`, a measurement transcript naming one
dune by a personal opam root. **Fixed, not exempted**, because the very next line
of the same transcript already wrote `<project>/_opam/bin/dune`; consistency
inside one code block decided it. The gate catching a fresh instance on rebase
is the item working.

### Portable spellings — 106 occurrences, all legitimate, all OUT of scope

Measured 2026-07-31: **48** `~/` occurrences in 32 files, **58** `$HOME`/`${HOME}`
occurrences in 21 files.

**Decision: out of scope, and this is a decision rather than an oversight.** Both
harms are absent — those spellings name no user and re-run correctly for
everyone. And a rule starting with a hundred known failures does not survive
contact with CI. The consequence is stated in the rule's own header: it cannot
see a portable reach into a home directory, **which is exactly the shape it wants
people to write.**

---

## 2. The mechanism, and why

**Scoped exemptions** — the precedent `scripts/cited-paths-exempt.tsv` already
sets (#396) — in `scripts/home-path-exempt.tsv`:
`citing/file::home-root<TAB>reason`.

**Why not a use-versus-describe pattern.** It does not fit the sweep. The
legitimate cases are `/home/opam/` *used for real* in a Dockerfile and
`/home/mathias/` *described* in a comment about a regex, and nothing lexical
separates either from a design document naming the same paths. There is no
pattern; there is only a per-file judgement, so the mechanism has to record
per-file judgements. No third mechanism was invented.

**Scoped-ONLY — there is no bare form, and a bare row is `exit 2`.** Every
legitimate case above is legitimate *because of the file it sits in*. A bare row
would wave `/home/opam/` through a design document too.

**What it cannot catch — an exemption is a judgement, and a wrong one is
invisible.** The reason field is prose; nothing checks that it is *true*. The
mitigations are that a row is scoped to one file and one root, that a row
matching nothing is `exit 1` (so it cannot rot quietly), and that the reason is
mandatory (`exit 2` without one).

**The token is the home ROOT, not a whole path** — so one row covers the
Dockerfile's twelve lines. The cost: a username appearing anywhere other than the
second path component is not seen. A component like `-home-<user>-dev-...` is not
a match, and one such string in `formal-verif-autopilot.workflow.js` had to be
fixed by hand because no path rule can see it.

---

## 3. Portability, not identity — so the message says the fix

The `parser.md` case mattered because **the verification command could not be
re-run by anyone else**; the fix was naming the opam *switch* rather than its
path. So when the offending line looks like a switch (`--switch=`, `OPAMSWITCH`,
`opam switch`), the error adds:

```
  -> name the opam SWITCH, not its directory (`--switch=.` from the repo root)
```

Proven by mutation: reintroducing a switch path into
`docs/plans/phase2-playground-mvp-2026-06-03.md` produces that exact line;
removing the hint turns the harness case red.

---

## 4. What the rule cannot catch — the complete list

Written into the gate header as limits (a)–(d), the `kb/properties.md` row, and
the green line, so no reader has to infer it:

1. **`~/`, `$HOME`, `${HOME}`** — out by decision (§1).
2. **Accounts outside `[A-Za-z0-9_-][A-Za-z0-9._-]*`, or not terminated by `/`
   or end-of-token.** A name holding a `$` or `+` is **not a finding at all**.
   This is a trade, not an accident: `/home/<user>/` and `/home/$USER/` are how
   this repo — *including the comment documenting this rule* — writes a home path
   it does not mean literally. A class admitting `<` or `$` would make the rule's
   own documentation a violation of the rule. Requiring termination also stops a
   **truncated** root being reported, which mattered more than the miss: an
   exemption written for a truncated root would have aliased a genuine account of
   the shorter name.
3. **Only the two leading components** (§2).
4. **Scope is the INDEX**, so a fix must be staged before the rule sees it; and
   it reads neither git history nor files git treats as binary.
5. **Nothing else in the gate's name.** No hostnames in prose, usernames outside
   a path, IPs, MACs, serials or e-mail addresses. A bare hostname in prose is
   pinned **GREEN on purpose**, so that widening the rule has to turn a declared
   case red and be re-decided rather than the coverage claim drifting.

---

## 5. The recursive trap, handled rather than papered over

- The **exemption file is scanned by the rule**. Only the **first field** of a
  well-formed row is waived — by scanning the reason alone — so a home path in a
  *reason*, or in a comment shaped like a row, is a finding. Both pinned.
- The **harness's own fixtures** use home roots (`/home/fixture-user/`,
  `/Users/fixture-user/`, `/home/fixture-user2/`) that exist nowhere else,
  scoped to the harness file. A *real* absolute home path pasted into the test
  file is still red. A file-level exemption would have made the gate blind to its
  own test data.
- The second fixture root exists because the reason-leak and row-shaped-comment
  cases need a root the fixture's *own* row does not cover — otherwise those two
  cases would have passed for the wrong reason.
- The gate's own source is scanned on every harness case (`check()` stages it).
  It caught its own comments twice during development, and those examples were
  rewritten as placeholders rather than exempted.

---

## 6. Adversarial review: 17 defects found and fixed

Three rounds (unscoped fresh-read + cross-runtime). Every one is a
claim-wider-than-its-code or a gate-that-cannot-fail, and every one now has a
harness case:

| # | defect | fix |
|---|---|---|
| 1 | whole-line waiver in the exemption file hid a path in a **reason** | waive the first field only, on parser-recorded rows |
| 2 | `git grep ... \|\| true` → **vacuous pass** on any git error | `gg` helper, status captured, `>1` is exit 2 |
| 3 | `exit 2` inside a **command substitution** ends only the subshell | helper returns via `$GG_OUT`; never called in `$( )` |
| 4 | over-broad: `/srv/home/x`, URLs, `/home/../etc` fired | token-start boundary; first account char is not `.` |
| 5 | **truncated root** on `/home/a+lab` → exemption aliasing | account must run to `/` or end-of-token |
| 6 | two **adjacent roots** collapsed into one finding a single exemption then covered | IFS tokenising, not a consuming regex boundary |
| 7 | `file:line:text` parse defeated by a **colon in a tracked filename** | NUL-separated file list + in-script line numbering |
| 8 | `git show ":$path"` read `0:note.md` as **stage syntax** | `git show ":0:$path"` |
| 9 | exemptions read from the **working tree** while findings came from the index | both from the index |
| 10 | exemption key split on the **first** `::` | split on the last |
| 11 | **CRLF** checkout made every row malformed | strip trailing `\r` |
| 12 | trailing slash required → blind to 3 tracked slashless roots | slash optional, root canonicalised |
| 13 | `BASH_REMATCH[1]` was the alternation, not the root | root parenthesised as a whole |
| 14 | pre-existing rule 2 `grep \| grep \|\| true` — a broken external `grep` made it **report clean** | bash `=~`; no external filter in the gate at all |
| 15 | exit-2 paths **leaked a temp file** each | `trap ... EXIT` |
| 16 | last line lost when input had no trailing newline | `read ... \|\| [ -n "$x" ]` |
| 17 | unquoted heredoc ran `` `suffixed` `` as a command substitution on every harness run | quoted |

---

## 7. Both polarities proven

**Positive — a genuine violation fails.** 11 mutations, each asserted landed with
`git diff --numstat`, each producing its declared red case (not just a non-zero
exit):

| mutation | landed | red case |
|---|---|---|
| delete rule 4 | `0 228` | 38 cases |
| unscope exemptions | `3 3` | scoped exemption travels to another file |
| whole-line waiver | `2 2` | reason leak; row-shaped comment |
| report first root only | `1 0` | two adjacent roots are two findings |
| `git show ":$p"` short form | `1 1` | stage-number filename |
| exemption from working tree | `1 1` | unstaged exemption |
| remove CRLF strip | `0 1` | CRLF exemption file parses |
| remove the trap | `1 1` | exit-2 temp leak |
| restore rule 2's grep pipeline | `1 7` | broken external grep |
| split key on first `::` | `2 2` | filename containing `::` |
| remove the switch hint | `1 1` | switch-naming remedy message |

**Negative — every legitimate mention still passes.** With all rows present the
gate is green on the real tree. Dropping each of 6 exemption rows in turn was
measured: every one produces exit 1 **with the refusal message and the file
named**, so no row is decoration.

**Message-level, not exit-code-level.** The two headline proofs assert text:
`absolute home path naming a user account` + `kb/sarek/ppx/parser.md:106:
/home/mathias/`, and `name the opam SWITCH, not its directory`.

---

## 8. Gates

Switch `/home/mathias/dev/SPOC` (dune **3.24.1**, OCaml **5.4.0**, ppxlib
**0.38.0**) unless stated.

| gate | exit |
|---|---|
| `scripts/check-no-machine-identifiers.sh` | **0** |
| `scripts/check-no-machine-identifiers.test.sh` | **0** — all **113** cases |
| `scripts/check-kb-properties.sh` | **0** |
| `scripts/check-kb-properties.test.sh` | **0** |
| `scripts/prove-red.sh` | **0** — 8 subjects, 53 mutations |
| `scripts/check-license-headers.sh` | **0** |
| `scripts/check-cited-paths-exist.sh` | **0** |
| `scripts/check-cited-paths-exist.test.sh` | **0** |
| `scripts/check-dune-opam-portability.sh` | **0** |
| `scripts/check-opam-clean.sh` (project switch) | **0** — was 1 before #399; the backlog-213 failure is gone |
| `scripts/check-opam-clean.sh` (ambient dune **3.23.0**, OCaml **5.3.0**) | **2** — known; the tree does not build under 3.23.0 |
| `dune build` | **0** |
| `dune build @fmt` | **0** |
| `dune runtest --force` | **0** |

## 9. Wiring

- `kb/properties.md` → `KB-GATE-NO-MACHINE-IDENTIFIERS` updated to describe four
  surfaces and every limit above. `check.tool` and `check.red_path` unchanged.
- Both are invoked from a carrier: `.github/workflows/ci.yml:137`
  (`check-no-machine-identifiers.sh`) and `:298`
  (`check-no-machine-identifiers.test.sh`). The ci.yml comments were corrected —
  they said "three surfaces" and "Five independent red shapes".

## 10. Not verified

- **CI itself.** Every exit code above is local. The `.github/workflows/ci.yml`
  edits are comment-only, but the workflow has not run.
- **The `.*.aux` gitignore pattern in practice.** 16 files were untracked and the
  producer instruction was changed, but no Rocq build was run afterwards to
  confirm regenerated `.aux` files stay ignored.
- **The `formal-verif-autopilot.workflow.js` changes were not executed.** It
  carries a top-level `return`, so it only parses inside a function wrapper; that
  was checked against `origin/main` as a baseline (both parse identically
  wrapped, both fail identically unwrapped). The env-var behaviour and the two
  conditional prompt blocks are unexercised — there is no harness for CWR
  workflows in this repo.
- **`/Users/<user>/` on a real macOS tree.** Pinned by a fixture only; the sweep
  found zero tracked instances.
- **The exemption *reasons*.** Prose, unchecked (§2).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection and reporting of accidentally committed machine-specific home-directory paths.
  - Added fail-closed behavior and clearer remediation guidance when portability checks cannot complete.
  - Updated automation to work without developer-specific filesystem paths.

- **Tests**
  - Expanded coverage for home-path leaks, exemptions, staged content, line endings, and failure scenarios.

- **Documentation**
  - Updated portability, workflow, and verification guidance to use repository-relative paths and local environment settings.

- **Chores**
  - Removed generated proof-build metadata and prevented it from being reintroduced.


mathiasbourgoin:
Thanks — four fixed in `d66f7430`, two answered below. Finding 1 is the one that mattered: it is a defect this branch already fixed once in the same rule and I left the second instance.

## 1 — MAJOR, `check-no-machine-identifiers.sh:129`, `git ls-files` failing still reported clean · **FIXED**

Correct, and correctly diagnosed as "the `|| true` shape the `gg` helper was added to remove, moved one call site along". Your note that no existing shim reached it was also right — the harness could not have caught it.

Rather than fix the one call site, every git invocation in the file now routes through one of **two** fail-closed helpers, and there are exactly two because a NUL-separated result cannot survive a bash variable:

- `gg` — string result in `$GG_OUT`, `>1` is exit 2.
- `git_to_file <dest> <max-ok-status> <git args...>` — so `1 = no matches` stays fine for `grep`, while any non-zero from `ls-files` is exit 2.

Neither is called inside a command substitution, for the reason already in the comment above `gg`.

**Proven on the message, not the exit code**, because a vacuous pass and a real pass differ only in output. One tree, one `git` shim that fails only `ls-files`, one tracked `myhost_…json`:

```
-- fixed gate:
::error::git ls-files failed (exit 128) -- this check cannot report a pass
-- origin/main gate, same shimmed tree:
OK -- 0 tracked payload(s), no machine identifier in payloads, paths or producers
```

Harness case added in the shape you asked for (`red: a failing git ls-files says so instead of passing`); it asserts the `cannot report a pass` text **and** the absence of the green line. Mutation: restoring the process substitution → landed `1 4`, that case red.

## 4 — MINOR, `kb/properties.md:102`, two claims wider than the code · **FIXED, both by correcting the code**

You gave the choice of narrowing the sentence or making it true. Both are now true:

- *"Every git grep routes through a helper"* — the candidate scan's direct `git grep -lI --cached -z` with an inline `$?` now goes through `git_to_file`. The row is reworded to *"EVERY git invocation … one of two fail-closed helpers"*, which is stronger and is what the code does.
- *"no external process failing can empty a result"* — this was the same defect as finding 1; fixed there, so the sentence holds. The row now also names the `ls-files` process-substitution instance explicitly, since a row that only states the rule and not the instance it got wrong is the thing that let this through.

## 3 — MINOR, `:354`, CRLF drops a root at end of line · **FIXED**

Right, and worse than a dropped line: as you noted, the slashless shapes the rule was *widened for* — a `chown` argument, a `--switch=` value — are exactly the ones sitting at end of line, so this was a whole class of input on which the rule could not fail. `text="${text%$'\r'}"` in the scan loop, matching the exemption parser.

Pinned with a fixture whose **tracked content** is CRLF (`check()` cannot produce that, so it is built directly) carrying both slashless shapes, asserting both line numbers are named. Mutation: remove the strip → landed `0 1`, that case red.

## 2 — MINOR, count drift 78 vs 106 · **FIXED, by deleting the number**

You are right that the copies must agree, since the figure is the stated evidence for keeping portable spellings out of scope. I did not reconcile them — reconciling would have created a fourth copy to drift. The number is **removed from all prose** and the red path re-measures it from the tree:

```
PASS green: 109 tracked portable spellings (48 tilde, 61 HOME), gate still 0
```

It reads 109, not 106, and two of the new occurrences are this branch's own files — which is the argument for not writing it down. The case is anti-vacuous: a measurement under 50 **fails**, because if the tree carried none, the gate's green would say nothing about the decision.

## 5 — MAJOR, `formal-verif-autopilot.workflow.js:54`, fail-closed `SPOC_ROOT` · **the check is added; the mechanism is not the one you proposed, deliberately**

Your diagnosis stands: `process.cwd()` makes a wrong `SPOC_ROOT` **wrong rather than absent**, `FORMAL`/`PLAN_FILE` become well-formed strings naming a tree that does not exist, and an agent then reports the emptiness as project state.

I first implemented it as you specified — `import fs`, `fs.existsSync(PLAN_FILE)`, top-level `throw`. Then I reverted it, because **this file is a CWR workflow body, not a module** (it carries a top-level `return`, so `node --check` only accepts it inside a function wrapper). Adding an `import` and a top-level `throw` introduces runner semantics I have no way to test in this repo — there is no CWR harness — and shipping an unverifiable `throw` into the one path that runs before every tick trades your bug for a worse one.

So the file's runtime semantics are now **identical to `origin/main` apart from string construction**, asserted rather than claimed:

```
$ git diff origin/main -- …/formal-verif-autopilot.workflow.js | grep -E '^\+' | grep -E 'import |require\(|throw |fs\.'
NONE
```

The validation lives in the **Adapt prompt**, where it is data and can be read: a PRE-FLIGHT block that checks `${PLAN_FILE}` and `${SKILL_ROOT}` exist, stops with `HARD: SPOC_ROOT resolved to … which has no formal/convergence-safety tree — set SPOC_ROOT to the repository root`, is told not to create the file or proceed with an empty plan, and must report both checks so a pass is visible rather than assumed.

The prologue is now sliceable, and I executed it in isolation across three configurations — including the `HOME` unset + `FORMAL_APPARATUS_ROOT` set recovery you flagged as impossible in the previous revision:

```
SPOC_ROOT=/repo HOME=/h        → SKILL_ROOT=/h/.claude/skills/formal-apparatus
SPOC_ROOT unset, cwd=/tmp      → FORMAL=/tmp/formal/convergence-safety
HOME unset, APPARATUS_ROOT=/app → SKILL_ROOT=/app          (recovery works)
```

and no configuration produces a home path in any output.

**Still not verified, stated plainly:** whether the runner surfaces a prompt-level HARD stop the way the file's other `HARD:` paths assume. That is a property of CWR, not of this diff, and it is why the check is a prompt rather than a throw.

## 6 — MINOR, `check-dune-opam-portability.sh:22`, `opam exec --switch` · **no invalid usage introduced**

Your own analysis chain confirms `opam exec --switch=SWITCH -- COMMAND` is valid, and `opam exec --help` here agrees. Line 22 is the line this branch changed, and it does not use `opam exec` at all:

```
#     $ <ambient-switch>/bin/dune build sarek.opam                  # 3.23.0
```

It is a recorded transcript invoking each `dune` **by absolute path**, and the sentence above it says why: *"with each dune binary invoked by an absolute path to its own switch — no dependency resolution involved"*. Rewriting it as `opam exec --switch=… -- dune` would reintroduce resolution and so misdescribe what was measured. The change replaced a personal opam root with a `<ambient-switch>` placeholder, matching the `<project>/_opam/bin/dune` form the next line of the same transcript already used — which is what made this a fix rather than an exemption.

---

## Gates after `d66f7430`

Switch `/home/mathias/dev/SPOC`, dune **3.24.1**, OCaml **5.4.0**, ppxlib **0.38.0**.

| gate | exit |
|---|---|
| `check-no-machine-identifiers.sh` | 0 |
| `check-no-machine-identifiers.test.sh` | 0 — **116** cases (was 113) |
| `check-kb-properties.sh` + `.test.sh` | 0 / 0 |
| `prove-red.sh` | 0 — 8 subjects, 53 mutations |
| `check-license-headers.sh` | 0 |
| `check-cited-paths-exist.sh` + `.test.sh` | 0 / 0 |
| `check-dune-opam-portability.sh` | 0 |
| `check-opam-clean.sh` (project switch) | 0 |
| `dune build` / `@fmt` / `runtest --force` | 0 / 0 / 0 |

Both new mutations asserted landed via `git diff --numstat` and each produced its declared red case.

PR #403, opened 2026-07-31
fix(codegen,wgsl): refuse whole-value array/vector equality (backlog-217)

mathiasbourgoin:
## Summary

- backlog-217: the WGSL emitter accepted whole-value equality (`=`/`<>`) on a `TVec`/`TArray`-typed operand (a kernel-vector parameter or a local array) and printed `(a == b)`. That is exactly what backlog-194/#401 left unrefused on purpose — `Sarek_typer.reject_aggregate_equality` / `Sarek_types.is_uncomparable_operand_typ` deliberately exclude `TVec`/`TArray`, because the same comparison is a legal pointer-equality check in the four C-family emitters (clang -x cl / glslangValidator both accept `(a == b)` on a vector/array param at exit 0).
- WGSL does not generalise the same way. naga (30.0.0) has no equality operator on `array<T>`/`array<T, N>` and rejects the identical shape with `Incompatible operands: Equal(Array { .. }, _)`.
- Not closed by #401 — confirmed by running the reproducer through the full frontend pipeline before writing the fix: `Sarek_transpile.of_source WGSL "fun (a : float32 vector) (b : float32 vector) ... -> if a = b then ..."` still emitted `(a == b)`, and feeding that output to `naga` failed to parse with the exact error above.
- Fix: `sarek/codegen/Sarek_ir_wgsl.ml` refuses `EBinop (Eq | Ne, ..)` in the ONE emitter this is true for, via a new `is_array_shaped_operand` predicate that only matches a bare `EVar` of `TVec`/`TArray` type. Reading an element (`EArrayRead`) still lowers to ordinary scalar equality and is unaffected. The four C-family emitters are untouched.
- Added `sarek/tests/codegen_golden/test_wgsl_array_equality_refusal.ml` (4 Alcotest cases): WGSL refuses vector-parameter equality, WGSL refuses local-array equality, WGSL still accepts scalar element equality (over-refusal guard), and the four C-family emitters (OpenCL/CUDA/Metal/GLSL) still *emit* the operator unchanged — that last case is a regression guard on the emitters' output, not a compiler-acceptance check; it invokes no compiler and exercises no Vulkan backend. The compiler-acceptance claim above (clang -x cl / glslangValidator both exit 0) was established by hand, separately, before this fix.
- Added a KB entry in `kb/sarek/tests/README.md` documenting the asymmetry and the fix.

## Evidence

No WGSL device on this host and WGSL is not on the e2e `Device.init` list — evidence here is generated WGSL source plus `naga` as a validator, not a device run.

Toolchain: switch `/home/mathias/dev/SPOC/_opam`, dune 3.24.1, OCaml 5.4.0, ppxlib 0.38.0.

Reproduction against naga (before the fix), via the real frontend pipeline:
```
$ naga /tmp/.../vec_eq_real.wgsl
Could not parse WGSL:
error: Incompatible operands: Equal(Array { base: [0], size: Dynamic, stride: 4 }, _)
```
Confirmed clang -x cl and glslangValidator (`-V`) both accept the equivalent OpenCL/GLSL output at exit 0 for the same kernel.

## Gates (this branch, single commit)

All exit codes captured directly (never through a pipe).

| Gate | Exit | Notes |
|---|---|---|
| `dune build` | 0 | |
| `dune runtest --force` | 0 | 0 FAIL across the full suite; new `wgsl-array-equality-refusal` test wired into the default `codegen_golden` runtest alias, 4/4 pass |
| `dune build @fmt` | 0 | clean after `dune promote` on the two new/changed files (pre-existing doc-comment warnings elsewhere are unrelated and untouched) |
| `make test_negative` | 0 | |
| `scripts/prove-red.sh` | 0 | 8 subjects / 53 mutations, all as declared |
| `scripts/check-kb-properties.sh` (+ `.test.sh`) | 0 / 0 | |
| `scripts/check-ppx-construct-names.sh` (+ `.test.sh`) | 0 / 0 | |
| `scripts/check-license-headers.sh` | 0 | |
| `scripts/check-cited-paths-exist.sh` (+ `.test.sh`) | 0 / 0 | |
| `scripts/check-opam-clean.sh` | 0 | converged over two `make opam` runs, matches PR #399's fix |

## Prove-red (this fix specifically)

Fix committed first (`b77645ce`), then mutated and restored:
1. Flipped `is_array_shaped_operand`'s `true` branch to `false` (mutation confirmed landed via `git diff --numstat` before trusting the result: 1 line changed).
2. Re-ran the new test: exit 1, both refusal cases failed with the exact expected assertion text (`"array/vector equality was EMITTED by WGSL, not refused (backlog-217) — naga will reject the output with \"Incompatible operands: Equal(Array …)\""`) — anchored on message text, not just exit code.
3. Restored the fix; `git diff --numstat` empty; re-ran: exit 0, 4/4 pass.

## Review

Cross-runtime review recorded per the local Sarek review gate: `codex exec` (GO, no blocking findings) + Claude (GO). Recorded in the local ledger before push.

## What I could NOT verify

- No WGSL device on this host — everything above is generated-source + naga validation, never a live WebGPU run.
- Did not verify against every naga version, only 30.0.0 (the one installed here).


mathiasbourgoin:
Addressed the Minor finding (`test_wgsl_array_equality_refusal.ml:15`): the claim was wider than what the test checks, not the test's fault.

Fixed the claim rather than the test, per the note — no validator invocations added:

- File header and `test_c_family_still_accepts`'s comment now state plainly that the test only asserts the four emitters still *emit* `==`/`!=` unchanged (i.e. that the WGSL-only refusal did not widen onto them); it invokes no compiler, no naga, no validator.
- The compiler-acceptance claim (clang -x cl / glslangValidator -V, both exit 0) is now attributed to the manual measurement made by hand before writing the fix, cited as such, not implied to come from this test.
- Removed "Vulkan" from the test-case name and comments — no Vulkan backend is exercised here; GLSL acceptance via glslangValidator is a separate, already-correctly-stated claim.
- Same correction applied to the `kb/sarek/tests/README.md` entry, which had the identical overclaim.
- PR body above corrected likewise (see the updated description).

No assertion logic changed, so no re-prove-red was needed (only re-ran the test itself: still 4/4 green, unchanged behavior).

Gates re-run on the new commit (`691ebc3d`), same toolchain (switch `/home/mathias/dev/SPOC/_opam`, dune 3.24.1, OCaml 5.4.0, ppxlib 0.38.0):
- `dune build @fmt`: 0 (clean after `dune promote` on the two touched files)
- `dune runtest --force`: 0 (0 FAIL)

Cross-runtime review recorded GO (codex + claude) for `691ebc3d` in the local Sarek review-gate ledger before push. Not merging, no `--admin`, no CodeRabbit re-review requested.

PR #405, opened 2026-07-31
backlog-214: five emitters accepted an SoA request and answered with AoS source

mathiasbourgoin:
## backlog-214 — five emitters accepted an SoA request and answered with AoS source

`Framework_sig.generate_source` offers `?soa_params:string list` to every backend. Only the CUDA/PTX emitter lowers a named vector parameter to N per-leaf base pointers plus one shared length. Five source-generating backends bound the parameter away and returned their ordinary packed-AoS source:

| plugin | before | after |
| --- | --- | --- |
| `sarek-cuda/Cuda_ptx_plugin.ml` | `?(soa_params = [])` → emitter | unchanged — this is the one that implements it |
| `sarek-cuda/Cuda_c_plugin.ml` | `?soa_params:_` | refuses a non-empty list |
| `sarek-hip/Hip_plugin.ml` | `?soa_params:_` | refuses |
| `sarek-opencl/Opencl_plugin.ml` | `?soa_params:_` | refuses |
| `sarek-vulkan/Vulkan_plugin.ml` | `?soa_params:_` | refuses |
| `sarek-metal/Metal_plugin.ml` | `?soa_params:_` | refuses |

The launch side expands an SoA-dispatched vector into N `RSA_Buffer`s plus one `RSA_Vector_Length`, so AoS source under an SoA argument list is an ABI mismatch no compiler catches.

**The defect is not that this happens — it cannot happen today. It is that the guarantee lived in exactly one place, on the caller side, and the emitters accepted the parameter without complaint.** `Execute.soa_dispatch` restricts SoA to a `"CUDA/PTX"` device and `Soa_launch` gates on `PTX` being in the backend's `supported_source_langs`. Those two predicates were the whole thing standing between an SoA launch and a mismatched signature. `?soa_params:_` is a silent filter where a refusal belongs — the same class as backlog-192, closed for the parser in #398.

Each of the five now calls the shared `Backend_error.reject_soa_params ~backend`, which raises `Plugin (Feature_not_supported _)` naming the framework and the requested parameter, and returns `()` for `[]` so the caller-side fast path is byte-for-byte unchanged. `~backend` is the plugin's own `name`, not its `*_error` module's: `Cuda_error` says `"CUDA"` for both CUDA plugins, and CUDA/PTX is precisely the one that does support SoA.

### What I verified before building, since three of four items filed that day were mis-filed

- **The ignoring is real, and the filed line numbers were exact** — `Cuda_ptx_plugin.ml:37`, `Opencl_plugin.ml:190`, `Vulkan_plugin.ml:190`, `Metal_plugin.ml:183`, `Cuda_c_plugin.ml:36`, `Hip_plugin.ml:33`, all unmoved by #375.
- **No backend could honour it instead.** None of `Sarek_ir_cuda`, `Sarek_ir_opencl`, `Sarek_ir_metal` or `Sarek_ir_glsl` contains any SoA lowering; `soa_params` appears in the PTX emitter and nowhere else. Refusing is not an overshoot here. (HIP reuses the CUDA-C generator verbatim, so one lowering would retire two of the five refusals — noted for backlog-215.)
- **No caller can bypass `Execute`'s gate.** Three call sites: `Execute.ml:944` (gated), `Soa_launch.ml:327` (gated, by a different predicate that happens to select the same single backend — only `Cuda_ptx_plugin` advertises `PTX`), and `Kirc_kernel.source_for_backend`, which omits the argument entirely. **`run_source` never calls `generate_source` at all** — it takes a source string. On CodeRabbit's adjacent #375 point: `run_source`'s transfer and expansion both default to `~soa_abi:false`, so its two halves agree on the packed ABI. **This change is therefore prophylactic, and the PR does not claim otherwise.**

### Scope

Emitter-side refusal only. The caller-side gate is untouched and extending SoA to another backend stays backlog-215.

Native, Interpreter and WebGPU return `None` unconditionally, generate no source, and carry no refusal. **WebGPU is the one to watch** — it is a JIT backend whose WGSL codegen is not yet wired, and when it is wired it inherits the original defect. Flagged in `Execute.ml` and `Framework_sig.ml` rather than pre-emptively patched.

**Nothing in this PR makes a future backend refuse by construction.** A new plugin that omits the call still silently emits AoS for an SoA launch. `Framework_sig`'s doc now states the obligation, and that is a comment, not a gate. A source-level gate over the plugin directories would close it; it is not here, because it is a new gate needing its own `kb/properties.md` row, red path and carrier.

### Evidence

Six Alcotest binaries, 24 cases, no GPU, all under the default `runtest` alias so CI runs them. `sarek-hip/test` gets its first `(test)` stanza — everything else there is an `e2e-hip` executable — because this check needs neither a device nor ROCm.

Both polarities per backend, and each separation was proved by **mutating the code and requiring the expected MESSAGE**, with `git diff --numstat` asserted non-empty first so a pattern that failed to match could not read as separation holding:

| mutation | expected | observed |
| --- | --- | --- |
| revert each of the five to `?soa_params:_` | that backend's refusal case red | 5/5 red on `expected a refusal, got N bytes of AoS source` |
| add the refusal to CUDA/PTX | no-overshoot case red | red on `CUDA/PTX raised a plugin-level refusal` |
| break the PTX leaf-pointer naming | the direct record case red | red on `a record vector still lowers to per-leaf pointers` |
| make the refusal fire on `[]` too | the fast-path case red | 5/5 red on `empty list still generates AoS source` |
| refuse only names that are *not* the record param | the record case red, scalar case green | 5/5 red on the record case only |
| move the Vulkan refusal below codegen | the ordering case red | red on `the refusal precedes codegen` |

The last two exist because of review. The refusal tests originally probed only with a scalar-element vector — a name the PTX emitter would itself reject as ineligible — so the pair could not distinguish "refuses any non-empty list" from "refuses only ineligible names while still ignoring the eligible ones", and the second is the pre-fix behaviour with a coat of paint. Each test now also names a flat two-leaf record, the shape a real SoA launch names.

`test_ptx_soa_not_refused.ml` is the other direction of the change: CUDA/PTX must **not** acquire the refusal. It asserts directly that a `point3d` vector still comes back as PTX carrying `param_sarek_soa_pts_{x,y,z}` + `param_sarek_pts_length` and not the packed `param_pts`.

### One thing deliberately refused although it would work

A record with a **single leaf**. `Soa.plan` permits it, and at N = 1 both ABIs are one binding plus one length, so the AoS source would bind correctly — codex proposed carving it out. It is refused anyway, and the reason is recorded next to the code: that correctness is a coincidence of the leaf count, not a property of the emitter. The same call means something different the moment the record gains a second field, and a carve-out that silently changes meaning under an unrelated edit is worse than a refusal. No in-tree caller can reach it either.

### Five review rounds, every finding one class

Every finding across all five rounds was **a claim wider than its code**. Not one was a wrong algorithm. Three of the five rounds found instances introduced by the *previous* round's correction, which is the part worth reading:

1. The refusal message described a "packed parameter signature (one pointer plus one length)". A GLSL compute entry point has no parameter list — SSBO bindings plus a push-constant length. And "silently wrong data, not a crash" was asserted uniformly and is wrong for Vulkan, which carries a second source-derived count (`validate_buffer_indices`, read from the GLSL `binding = N` declarations) that catches the mismatch late as a buffer-count rejection.
2. The five refusal tests probed only with a scalar-element vector — a name the PTX emitter would itself reject as ineligible — so they could not distinguish "refuses any non-empty list" from "refuses only ineligible names while still ignoring the eligible ones", and the second is the pre-fix behaviour with a coat of paint. Each now also names a flat two-leaf record. Also: the refusal deliberately covers a **single-leaf record**, which would in fact bind correctly (see above).
3. Narrowing (1) for OpenCL and Vulkan **left CUDA/C and HIP stated absolutely**, including in the `Framework_sig` doc comment that is the public contract for `?soa_params` — and dropped **Metal** from an enumeration that reads exhaustive, five backends named and four accounted for.
4. Round 3's replacement was then over-specific in the other direction, and its Metal fold carried a claim about Metal that Metal cannot produce. I found both myself, by asking the two questions a fix of that shape invites.
5. A fresh read found sharper versions of both and I took those. The mechanism sentence — "the length slot reads a pointer and the next pointer parameter reads 8 bytes out of a 4-byte cell" — holds only at N = 2 with a pointer parameter declared after the vector, which is exactly the probe kernels' shape. At **N = 3** the third leaf lands in the following pointer slot: a *valid* device pointer to the wrong buffer, silent corruption, nothing to trap on. With the vector declared **last**, nothing after it shifts into a pointer slot at all. So the claim is now the general form — every parameter declared after the vector receives the wrong entry — with the three shapes as conditioned examples under it.

   And the trap half is **false on Metal**, not merely unestablished: Metal never receives a caller-supplied address. Buffers go through `setBuffer` as MTLBuffer objects, scalars through `setBytes` as a driver-allocated inline copy, against `device T* x [[buffer(k)]], constant int &sarek_x_length [[buffer(k+1)]]`. Under the shift a length index reads a buffer's first 4 bytes as the length and a pointer index gets inline `setBytes` data indexed out of bounds, or another buffer — wrong data and out-of-bounds reads inside valid allocations, or an encode Metal's own API validation rejects. Nothing that traps in the CUDA sense.

Also worth recording: I first *refuted* a reviewer's report of literal space runs in the failure messages, using the pre-ocamlformat literal. Re-checked against the rendered output of the compiled binary, the reviewer was right and I was wrong. The messages now read `expected a refusal, got 155 bytes of AoS source`.

The commit bodies of the first two commits overstate the CUDA/C and HIP symptom in the way (3) describes. Left in place and corrected in the later bodies, so the sequence of corrections stays legible rather than looking as though it never happened.

### Known pre-existing inconsistency, deliberately not fixed here

`Backend_error.reject_soa_params` now cites `Execute.expand_to_run_source_args` as its authority for the shift. **`Execute.ml:270` and `Execute.ml:933` still describe that same shift as "not a crash but silently wrong data"** — the exact claim rounds 3 and 5 established is too narrow. Both predate this branch (`9d0ac351`, `5178b553`) and neither is touched by it, so following the citation currently lands on text that contradicts the citing text. Flagged so the next reader does not conclude the new text is the wrong half; being filed as its own item.

### Gates

Project switch `/home/mathias/dev/SPOC/_opam` — dune **3.24.1**, OCaml **5.4.0**, ppxlib **0.38.0**. Each exit code captured directly.

| gate | exit |
| --- | --- |
| `dune build` | 0 |
| `dune runtest --force` | 0 (see the flake note below) |
| `dune build @fmt` | 0 |
| `make test_negative` | 0 |
| `scripts/prove-red.sh` (8 subjects / 53 mutations) | 0 |
| `scripts/check-kb-properties.sh` / `.test.sh` | 0 / 0 |
| `scripts/check-ppx-construct-names.sh` / `.test.sh` | 0 / 0 |
| `scripts/check-cited-paths-exist.sh` / `.test.sh` | 0 / 0 |
| `scripts/check-opam-clean.sh` | **0** |
| `scripts/check-alcotest-registration.js` | 0 |
| backlog-214 prove-red rounds a / b / c | 0 / 0 / 0 |

`check-opam-clean.sh` exits **0**, not the 1 tracked as backlog-213 — #399 landed while this branch was parked and fixed the dune-3.24.1 `"dune" {>= "3.15" & >= "3.15"}` duplication. The working tree is clean after every gate, so no regenerated `.opam` is committed.

**One `dune runtest --force` run went red and I am reporting it rather than only the green one.** The failure was `test_record_variant_decl_order` (backlog-211, untouched by this branch) on `[CUDA/PTX] AMD Ryzen 9 7950X [ZLUDA]` — the ZLUDA-on-CPU device — with the other ZLUDA device passing the same case. Measured rather than argued: standalone on this branch 3/3 exit 0; standalone on `origin/main` built in a throwaway worktree, exit 0; full `dune runtest --force` re-run on this branch, exit 0. So it is a pre-existing flake under parallel device contention on this machine, which has ZLUDA and two Vulkan/OpenCL devices. CI has no such device and never runs that arm.

**Not verified:** nothing here runs on a GPU. The refusals and the PTX text assertions are host-side codegen, which is the whole point — but the per-backend *symptom* descriptions (illegal-address trap on CUDA/C and HIP, `CL_INVALID_ARG_INDEX` on OpenCL, the Vulkan buffer-count rejection) are read from the binding code and were never executed, because reaching them requires defeating the caller-side gate this PR deliberately leaves alone. They are stated as what the code does, not as observed behaviour. Ambient dune 3.23.0 / ppxlib 0.35.0 cannot build this tree at all (`Ptyp_open` needs ppxlib ≥ 0.37), so no row was measured there.

Rebased onto `9a3a8f79`. No file on this branch is touched by any of the five PRs that merged while it was parked, so the clean rebase cannot have hidden a rerere-applied resolution — verified by intersecting the two name lists, which is empty.

### DIFF
--- a/docs/optimization/tier1b-emitter-soa-handoff.md
+++ b/docs/optimization/tier1b-emitter-soa-handoff.md
@@ -201,8 +201,13 @@ codegen-level `~soa_params` knob:
 > in exactly one place: the caller-side predicate. The five source-generating
 > backends without an SoA lowering (CUDA/C, HIP, OpenCL, Vulkan, Metal) bound
 > `?soa_params:_` away, so a caller that reached them with a non-empty list got
-> AoS source for an SoA argument list — silently wrong data, not a crash. They
-> now raise `Backend_error.reject_soa_params` instead. The `"CUDA/PTX"` gate is
+> AoS source for an SoA argument list — an ABI mismatch no compiler catches,
+> whose consequence is per backend: rejected at bind or launch on OpenCL
+> (`clSetKernelArg` / `CL_INVALID_ARG_INDEX`) and Vulkan
+> (`validate_buffer_indices`, count read from the GLSL `binding = N`
+> declarations), silently misinterpreted data on CUDA/C and HIP, which bind
+> positionally into an unchecked argument array. They now raise
+> `Backend_error.reject_soa_params` instead. The `"CUDA/PTX"` gate is
 > unchanged and is still the thing that keeps the refusal unreached; what
 > changed is that it is no longer the only thing. Extending SoA to a further
 > backend remains separate work (backlog-215): each such backend removes its own
--- a/docs/optimization/tier1b-emitter-soa-handoff.md
+++ b/docs/optimization/tier1b-emitter-soa-handoff.md
@@ -205,8 +205,11 @@ codegen-level `~soa_params` knob:
 > whose consequence is per backend: rejected at bind or launch on OpenCL
 > (`clSetKernelArg` / `CL_INVALID_ARG_INDEX`) and Vulkan
 > (`validate_buffer_indices`, count read from the GLSL `binding = N`
-> declarations), silently misinterpreted data on CUDA/C and HIP, which bind
-> positionally into an unchecked argument array. They now raise
+> declarations); on CUDA/C, HIP and Metal, which all bind positionally with no
+> arity check against the compiled kernel, the shift can misinterpret data or
+> trap on an illegal address — `Backend_error.reject_soa_params` carries the
+> per-backend detail, and none of the three is described by "silently wrong
+> data" alone. They now raise
 > `Backend_error.reject_soa_params` instead. The `"CUDA/PTX"` gate is
 > unchanged and is still the thing that keeps the refusal unreached; what
 > changed is that it is no longer the only thing. Extending SoA to a further
--- a/docs/optimization/tier1b-emitter-soa-handoff.md
+++ b/docs/optimization/tier1b-emitter-soa-handoff.md
@@ -205,11 +205,11 @@ codegen-level `~soa_params` knob:
 > whose consequence is per backend: rejected at bind or launch on OpenCL
 > (`clSetKernelArg` / `CL_INVALID_ARG_INDEX`) and Vulkan
 > (`validate_buffer_indices`, count read from the GLSL `binding = N`
-> declarations); on CUDA/C, HIP and Metal, which all bind positionally with no
-> arity check against the compiled kernel, the shift can misinterpret data or
-> trap on an illegal address — `Backend_error.reject_soa_params` carries the
-> per-backend detail, and none of the three is described by "silently wrong
-> data" alone. They now raise
+> declarations); on CUDA/C, HIP and Metal, which all bind positionally with
+> nothing comparing the expanded list against the compiled kernel, the shift goes
+> unchecked and every declared slot from the vector onward is fed a value of the
+> wrong kind — `Backend_error.reject_soa_params` carries the per-backend detail,
+> and none of the three is described by "silently wrong data" alone. They now raise
 > `Backend_error.reject_soa_params` instead. The `"CUDA/PTX"` gate is
 > unchanged and is still the thing that keeps the refusal unreached; what
 > changed is that it is no longer the only thing. Extending SoA to a further
--- a/docs/optimization/tier1b-emitter-soa-handoff.md
+++ b/docs/optimization/tier1b-emitter-soa-handoff.md
@@ -197,6 +197,17 @@ codegen-level `~soa_params` knob:
   documented fallback that guarantees "never wrong data". `of_ctypes_ptr` /
   `sub_vector` stay AoS-only.
 
+> **Update (backlog-214).** The sentence above described a guarantee that lived
+> in exactly one place: the caller-side predicate. The five source-generating
+> backends without an SoA lowering (CUDA/C, HIP, OpenCL, Vulkan, Metal) bound
+> `?soa_params:_` away, so a caller that reached them with a non-empty list got
+> AoS source for an SoA argument list — silently wrong data, not a crash. They
+> now raise `Backend_error.reject_soa_params` instead. The `"CUDA/PTX"` gate is
+> unchanged and is still the thing that keeps the refusal unreached; what
+> changed is that it is no longer the only thing. Extending SoA to a further
+> backend remains separate work (backlog-215): each such backend removes its own
+> refusal when its emitter gains the lowering.
+
 This is pure plumbing (no new coalescing) — the roadmap (§1.2c/e) rates it the
 bulk of the SoA cost; it was descoped here so the emitter could ship complete
 and proven. When it lands, extend `test_soa_emitter_equiv` to drive SoA through
--- a/sarek-cuda/Cuda_c_plugin.ml
+++ b/sarek-cuda/Cuda_c_plugin.ml
@@ -33,8 +33,13 @@ module Backend : Framework_sig.BACKEND = struct
      node. That error is allowed to PROPAGATE so callers surface the name rather
      than the opaque "generate_source returned None" — catching it here and
      returning [None] hid it identically to the Vulkan swallow (PR #259). *)
-  let generate_source ?block:_ ?soa_params:_ (ir : Sarek_ir_types.kernel) :
+  let generate_source ?block:_ ?(soa_params = []) (ir : Sarek_ir_types.kernel) :
       string option =
+    (* Refused rather than bound away: this emitter has no SoA lowering, and
+       [name] is passed so the message says CUDA/C — the sibling CUDA/PTX
+       backend does honour it, and [Cuda_error]'s own backend string is "CUDA"
+       for both, which would not distinguish them (backlog-214). *)
+    Sarek_backend_error.Backend_error.reject_soa_params ~backend:name soa_params ;
     Some (Sarek_ir_cuda.generate_with_types ~types:ir.kern_types ir)
 
   let execute_direct ~native_fn:_ ~ir:_ ~block:_ ~grid:_ _args =
--- a/sarek-cuda/test/dune
+++ b/sarek-cuda/test/dune
@@ -102,3 +102,22 @@
  (libraries sarek_cuda sarek.codegen spoc.ir alcotest str unix)
  (instrumentation
   (backend bisect_ppx)))
+
+; backlog-214 - the CUDA/C plugin refuses a non-empty ~soa_params instead of
+; answering an SoA argument list with AoS source. Pure codegen, no device.
+
+(test
+ (name test_cuda_c_soa_refusal)
+ (modules test_cuda_c_soa_refusal)
+ (libraries sarek_cuda sarek_ir sarek_backend_error alcotest)
+ (instrumentation
+  (backend bisect_ppx)))
+
+; backlog-214, other polarity - CUDA/PTX must NOT have acquired that refusal.
+
+(test
+ (name test_ptx_soa_not_refused)
+ (modules test_ptx_soa_not_refused)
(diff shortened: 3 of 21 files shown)


mathiasbourgoin:
## Review round 5 applied — both findings, plus the MINOR-3 note

Commit `ec96793e`. Text only; `reject_soa_params` and all five plugin call sites untouched.

**NEW-1 — the mechanism was one case stated as the consequence.** Taken in full, including the counter-case as measured: at N = 3 the args are `leaf1,leaf2,leaf3,len,out_buf,out_len` against declared `(S* pts, int pts_len, T* out, int out_len)`, putting `leaf3` in the `out` slot — a *valid* device pointer to the wrong buffer, silent corruption, nothing to trap on — and `len` in `out_len`, coincidentally correct. And with the SoA vector declared last, nothing after it shifts into a pointer slot at all.

`Backend_error.ml` now leads with the general form as the claim — **every parameter declared AFTER the vector receives the wrong entry** — and demotes all three shapes (N = 2 with a following pointer, N = 3, vector-last) to conditioned examples under it, labelled as examples rather than the consequence. The CUDA/C and HIP test headers say the same. `Framework_sig.ml` and the handoff doc already hedged correctly and were left alone, as you asked.

**NEW-2 — the trap half is false on Metal.** Verified against the code you cited before taking it: `Metal_api` binds MTLBuffer objects via `setBuffer` and scalars via `setBytes` as a driver-allocated inline copy, and `Sarek_ir_metal.gen_buffer_param` declares `device T* x [[buffer(k)]], constant int &sarek_x_length [[buffer(k+1)]]`. No caller-supplied address ever crosses. Metal is now its own bullet in `Backend_error.ml` and its own paragraph in the test header, claiming only wrong-buffer and out-of-bounds-within-a-valid-allocation, plus the possibility that Metal's own API validation rejects the encode — explicitly **not** the trap. The CUDA/C + HIP bullet keeps the trap as one outcome among several, not the outcome.

Your note that the bad fold traced to your own earlier recommendation is recorded in the commit body, because the boundary call (nothing checks the arity) was right and carrying the CUDA mechanism across with it was not — those are separable and the log should say which one was wrong.

**MINOR-3 — noted, not fixed.** In the PR body under "Known pre-existing inconsistency, deliberately not fixed here": `Execute.ml:270` and `:933` still call the same shift "not a crash but silently wrong data", both predating this branch (`9d0ac351`, `5178b553`), so following the new citation lands on text that contradicts the citing text. Flagged so the next reader does not conclude the new text is the wrong half.

### Also worth flagging, because I would rather report it than only report the green run

One `dune runtest --force` went **red**: `test_record_variant_decl_order` (backlog-211, untouched here) on `[CUDA/PTX] AMD Ryzen 9 7950X [ZLUDA]`, the ZLUDA-on-CPU device, while the other ZLUDA device passed the same case. Measured rather than reasoned about:

| check | result |
| --- | --- |
| standalone on this branch, 3 runs | exit 0, 0, 0 |
| standalone on `origin/main`, built in a throwaway worktree | exit 0 |
| full `dune runtest --force` re-run on this branch | exit 0 |

A pre-existing flake under parallel device contention on this machine, which has ZLUDA plus two Vulkan and two OpenCL devices. CI has no such device and never runs that arm.

### Gates re-run after the fix

Project switch `/home/mathias/dev/SPOC/_opam` — dune **3.24.1**, OCaml **5.4.0**, ppxlib **0.38.0**.

`dune build` 0 · `dune runtest --force` 0 · `dune build @fmt` 0 · `make test_negative` 0 · `prove-red.sh` 0 (8 subjects / 53 mutations) · `check-kb-properties.sh` 0 · `check-ppx-construct-names.sh` 0 · `check-license-headers.sh` 0 · `check-cited-paths-exist.sh` 0 · `check-opam-clean.sh` **0** · `check-test-alias-coverage.sh` 0 · `check-alcotest-registration.js` 0 · `check-dune-dir-visibility.sh` 0. Six binaries, 24 cases, green. Working tree clean after every gate, so no regenerated `.opam` is committed.

Comment lines are all within 80 now; the three still over are ocamlformat's own output on code, matching the rest of the tree.

PR #199, opened 2026-06-14
refactor(codegen): split Sarek_ir_ptx.ml into 5 modules + PTX snapshot test

mathiasbourgoin:
## Summary

- Split 962-line `Sarek_ir_ptx.ml` spike into `Sarek_ir_ptx_{types,mem,expr,stmt,kernel}.ml` (all ≤500L)
- Add `.mli` interface file per module to enforce encapsulation; thin re-export façade kept for callers
- Add `test_ptx_snapshot` — CPU-only alcotest test that validates `demo_vector_add_ptx ()` against a PTX string validated by `ptxas --gpu-name sm_86`

## Module responsibility

| Module | Responsibility | Lines |
|---|---|---|
| `Sarek_ir_ptx_types` | reg_alloc, env, emit/emit_label helpers | 113 |
| `Sarek_ir_ptx_mem` | elt_shift, array read/write emission | 67 |
| `Sarek_ir_ptx_expr` | emit_expr, emit_binop, emit_cast, emit_intrinsic | 454 |
| `Sarek_ir_ptx_stmt` | emit_stmt, emit_assign_lv | 138 |
| `Sarek_ir_ptx_kernel` | generate, generate_raw, emit_params, emit_locals | 168 |

## Test plan

- [ ] `dune build` exits 0
- [ ] `dune build @sarek/tests/runtest` exits 0 (all existing + snapshot test)
- [ ] `dune fmt` clean
- [ ] Each module ≤500L (verified: max 454)
- [ ] Each module has `.mli`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Split PTX code generation into dedicated modules for shared PTX utilities, register/environment handling, expression lowering, statement lowering, array load/store helpers, and kernel emission.
  * Kept the external PTX generation entry points consistent while reorganizing the underlying implementation.
  * Added improved kernel emission, including `.param`/`ld.param` handling and configurable PTX header (compute target + PTX version).
* **Tests**
  * Updated the CPU-only PTX unit test to validate generated output using canonical marker checks instead of exact full-string snapshot matching.

PR #413, opened 2026-07-31
docs(design): refuse the scatter capability token (backlog-221)

mathiasbourgoin:
A design record whose conclusion is **REFUSE — do not build.** Docs-only:
`docs/design/soa-scatter-freshness-token.md`, and nothing else. (Against `main`
this branch also shows the shared `Soa_vector` work from `c93cf495`/`e9f8d612`;
that is #220's, already merged in PR #408.)

## The decision

A `Synced.t` capability token for `Soa_vector.scatter`. A phantom type on the
vector cannot work — auto-sync is a runtime mutable flag, and tracking mutable
per-value state statically needs affinity/uniqueness, which upstream OCaml 5.4
lacks. The tractable shape would carry the vector's identity plus its mutation
generation. The record concludes that every sound version collapses into the
runtime check already present.

**The argument was attacked and held.** The reviewer tried four encodings the
document does not name — scoped/CPS `with_host_fresh`, generative-functor
branding, effect handlers, witness-threading — and broke none of them: a scoped
region does not stop a `launch` inside it; branding fixes *identity*, not
*temporal* freshness; effect handlers give dynamic enforcement, i.e. a runtime
check.

## Invalidation set — complete, and smaller than it looks

Host-freshness is `location ∈ {CPU, Both, Stale_GPU}`; invalidation is the
transition into `Stale_CPU`. Repo-wide there are **exactly two** producers:
`Soa_vector.ml:222` (end of `soa_to_device`) and `Execute.ml:1038`
(`mark_vectors_stale`). The document names exactly those two — **no sixth event.**

Worth recording because it is counter-intuitive: `set`/`unsafe_set` and `fill`
write `Stale_GPU`, `free` writes `CPU`, and host→device/cross-device transfers
write `Both`. **None of them invalidates host freshness.** A census of all 26
production `location <-` writes reproduces exactly (Stale_GPU 14, Both 7, CPU 3,
Stale_CPU 2, GPU 0), repo-wide and `sarek/`-scoped alike.

## Known findings, recorded not fixed

`roster-review` GO, `roster-qa` GO, 15 findings, **0 HIGH**. Carry these:

- **`:301` says "Three things would reverse the refusal." Only two do.** Item 2
  (more callers reaching the hazard) is a cost sensitivity, not a reversal — and
  the document says so itself twice, at `:310-311` and `:317-319` ("more callers
  without affinity gives a more widely used unsound token"). Unlike the sibling
  record #199, the insufficiency is **disclosed inline**, so the defect is the
  label rather than a hidden concession. Better phrasing: one reverses
  (affinity), one dissolves the question (immutable `location`), one changes the
  cost case.
- **`:156` states the universal unscoped** — "Every sound version … collapses" —
  while `:262-263` states the same conclusion **correctly self-scoped**: "at any
  boundary examined". The document already contains the honest form of its own
  claim, 106 lines later.
- **`:145` "It has exactly two options"** — there is a third (thread the witness
  through the record field so the caller at `Execute.ml:310` produces it), which
  the document itself contemplates at §3 and costs at §6.
- **`:235` "Its siblings are all live"** is false and self-contradicting:
  `soa_aos_stride` has zero read sites — the same pattern used to call
  `soa_scatter` dead — and the next sentence admits `soa_gather` is dead. 3 of 8
  siblings are dead.
- **`:65`** cites `Spoc_core_base.mli:91` for the mutable `location` field; `:91`
  is `| GPU of Ops.device_t`, the variant is at `:89` and the field at `:192`.
  Invisible to the citation gate, which checks that a line exists and never what
  it says.

## Not verified

The §5 hole was confirmed **statically** — `Stale_CPU, true → () → ensure_cpu_sync
→ Vector_transfer.ml:45-49 ignore (cb.sync vec) → Transfer.ml:719 returns false
with no `to_cpu` → scatters stale bytes` — but **not re-executed**; its recorded
runtime measurement and positive control were not reproduced. Cross-runtime was
breaker-skipped (codex and opencode both `skipped-degraded` from review-phase
degradation), so this is single-runtime verification — a coverage gap, not a
pass. Gates run were only the three that mean anything for a docs change, and
**none of them is evidence about the document's reasoning.**

### DIFF
--- a/docs/design/soa-scatter-freshness-token.md
+++ b/docs/design/soa-scatter-freshness-token.md
@@ -1,0 +1,327 @@
+# A capability token for `Soa_vector.scatter` — refused, and what to build instead
+
+**Task:** backlog-221 — "can an auto-sync phantom type state make the `scatter`
+hazard unrepresentable, rather than merely refused at runtime?"
+
+**Date:** 2026-07-31.
+
+**Base:** the backlog-220 runtime refusal, on branch `backlog/220-scatter-refuse`.
+At the time of writing that branch is local and unpushed, so it is named rather
+than cited by sha — the sha would not resolve in a fresh clone, and an unlanded
+branch may still be amended.
+
+**Verdict: REFUSE.** No token, no phantom type. The exploration did not produce a
+design worth building, and it did produce one thing worth having: **the shipped
+refusal has a hole**, measured in §5. Fixing that hole is a one-line predicate
+widening and is worth more than every type-level option considered here.
+
+---
+
+## 1. The question was asked about the wrong state variable
+
+The framing that opened the item was:
+
+> "once auto-sync is off the sync token is valid, we must make sure it becomes
+> invalid if auto-sync is on again"
+
+That describes a token whose validity is keyed to a **flag flip**. The mechanism
+is not keyed to a flag flip. `Soa_vector.scatter` refuses on a **conjunction**
+(`sarek/core/Soa_vector.ml:109-110`):
+
+```ocaml
+match (Vector.location t.aos, Vector.auto_sync t.aos) with
+| Vector.Stale_CPU _, false -> raise (Soa.Unsupported ...)
+```
+
+and `Vector_transfer.ensure_cpu_sync` (`sarek/core/Vector_transfer.ml:44-51`) is
+a no-op unless *both* halves say so. In `CPU`, `Both` and `Stale_GPU` the host
+copy is already the freshest data, so scatter is safe **regardless of the flag**.
+
+So a witness proving "auto-sync is on" is the wrong token in both directions:
+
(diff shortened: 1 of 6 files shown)


mathiasbourgoin:
Superseded by a rebased branch. #220 merged by **rebase**, which minted new shas on `main` (`86153a2a`, `7c2c676e`) for the two commits this branch also carried (`c93cf495`, `e9f8d612`) — so the branch went DIRTY despite the file content being identical. Counting file sets caught the content overlap but not the commit-identity overlap.

Reopened from `main` with only #221's unique commit; the design document is sha256-identical to the reviewed version and the branch adds nothing else. `attest.sh --carry` was tried first and correctly refused (two dropped commits), so the new sha carries its own attestation.

PR #418, opened 2026-08-01
docs(formal): headline the shipped-artefact theorem count; make the split mechanical (backlog-201)

mathiasbourgoin:
> **`ci.yml` change AUTHORISED by Mathias, 2026-08-01.** This PR adds a line to
> `.github/workflows/ci.yml`, which `.claude/rules/escalation.md` names as requiring
> explicit human confirmation. That confirmation was given. Review and QA are both GO.

Headlines the **shipped-artefact** theorem count and lists model-only proofs
separately and labelled — not two numbers side by side, not one combined figure.

**94 shipped / 186 model-only / 280 total**, and they genuinely reproduce: a real
end-to-end `check-formal-proofs.sh` run on a clean tree, exit 0, all three
projects rebuilt from source and `coqchk`-re-checked, printed `TOTAL: 94` /
`model-only: 186` / `total: 280`. Toolchain: Rocq 9.1.1, dune 3.24.1, OCaml
5.4.0, ppxlib 0.38.0, ctypes 0.24.0, Python 3.14.6.

**#95's three-disagreeing-ledgers reconciliation was re-verified, not assumed** —
independently reproduced `79 + 111 + 90 = 280` with the axiom allowlist at 6
sanctioned / 6 found, exact match, every `proof-notes.json` anchor real. Sets are
clean: `50+44=94`, `29+46+111=186`, no double-count, no drop.

## What review found, and what was fixed

Round 1 returned NO-GO with 5 HIGH. Two mattered a great deal:

- **The new gate false-passed.** `strip_ocaml_comments` stripped comments but not
  **string literals**, so a bare string mention sustained a production-link claim.
  Proven by mutation: replacing the sole live call left the gate at **exit 0 still
  printing 94**; blanking the string turned it red. Now
  `strip_ocaml_comments_and_strings`, with the pre-fix false pass reproduced and
  the post-fix catch verified, and the opposite polarity checked (comment-plus-live
  and string-plus-live still count as live).
- **`rocq-value-ledger.md:44` said "None of the three figures is hand-typed:" over
  a fabricated transcript.** Only the `280` line matched what the script actually
  prints; **two of three lines were hand-composed inside a block presented as
  generated output** — the dominant defect class in its purest form, inside the
  document whose whole thesis is that the numbers are not hand-typed. Replaced
  with verbatim real output plus an honest "snapshot, not gate-covered" caveat.

Also: a **309-line gate with no red-path companion and no `kb/properties.md` row**,
against that file's own line 20 — *"A gate that cannot fail is worse than no
gate"*. Now `check-production-link.test.sh` (13 cases) plus a
`KB-GATE-PRODUCTION-LINK` row; `check-kb-properties.sh` goes 44 → 45 declarations.
**Case 3 is the regression pin and genuinely flips red when the string-stripping
fix is reverted** — verified by swapping in the pre-fix script body: 4 of 13 cases
fail, including case 3 exactly.

## Disclosed, not implemented

`formal/type-safety` has **no extraction-drift check**. `_CoqProject` lists only
the 12 theories; the extraction drivers are absent and the drift byte-compare is
guarded by `basename == codegen-ptx`. So `TypeSafetyModel.ml`/`UnifyModel.ml` —
the models its 44 theorems are differential-tested against — are committed OCaml
the verified rebuild never reproduces: a theorem can change, be re-proven,
re-counted and go coqchk-green while the test compares against the *older*
extraction.

Building that check is a real feature (new Rocq extraction drivers, build wiring,
drift-compare logic) and out of scope here, so it is **disclosed** — a
`no_anti_drift_check` field in `production-link.json` and an inline `STATUS.md`
scoreboard cell, both where a reader learns the 44-count, closing the original
complaint that README/ledger separated the two projects only on
exhaustive-vs-sampled.

## Copy reduction

94 appeared **9×** and 186 **4×**, none gate-covered. Now **6×** and **3×** — the
standing rule here is *prefer deleting a number to reconciling its copies*, since
reconciling just creates one more to drift. Also deleted README's false "so it
cannot go stale" (no gate reads any doc), the unimplemented theorem-count clause
from three prose blocks, and widened `type-safety/STATUS.md` to 3 columns so its
sampling caveat **renders at all** — it had 3 cells in a 2-column table.

`convergence-safety` stays consistent with backlog-228: `modules: {}`, both
theorem groups model-only, **0 of 111** production-linked, `known_gap` matching
`test_convergence_live.ml:35` (it calls the real `check_expr` but is a
fixed-pattern suite, not differential). No linkage is implied that does not exist.

## Interaction worth carrying

This change makes `check-formal-proofs.sh` the advertised way to reproduce the
headline figures — which **increases traffic to the script whose exit trap
destroys uncommitted work in `formal/`** (backlog-229, not yet merged).
`check-production-link.py` standalone reproduces 94/186 with **no rebuild and no
trap**; the docs should prefer it.

Disjointness with `backlog/229-formal-dirty-trap` is verified: #201 appends 10
lines after line 428 (from `a093b7b3`); #229 edits the pre-flight block and
`restore_tree`. No overlap.

`roster-review` GO · `roster-qa` GO.

## Summary by CodeRabbit

* **Documentation**
  * Clarified formal-proof coverage, distinguishing production-linked theorems from model-only results.
  * Added validation methods, limitations, reproduction instructions, and known convergence-safety linkage gaps.
  * Added provenance and evidence details for PTX layout and type-safety verification.

* **Quality Improvements**
  * Added automated validation for declared links between formal proofs and shipped code.
  * Added safeguards against invalid claims, missing evidence, and misleading references.
  * Added comprehensive coverage for successful and failure scenarios.


mathiasbourgoin:
### CodeRabbit review response — 4 of 10 addressed in `b4bd4ffa`, 6 deferred with reasons

**Fixed**

1. **MD040** — `text` added to the fence in `docs/formal/rocq-value-ledger.md`. Fence *content* verified byte-identical and re-checked against a live `check-production-link.py` run; that block is a real transcript (an earlier round replaced a fabricated one there), so it was important not to disturb it.
2. **`#n-03` anchor** — added before the N-03 heading, matching the existing `<a id="n-01"></a>` pattern. Confirmed the `#n-03` link actually exists (line 77), so the anchor is not decorative.
3. **`convergence-safety/production-link.json`** — *"must appear as a live call, comments stripped"* → *"live reference, comments and string literals stripped"*. Verified against the code: the matcher is `re.escape(call) + r"\b"` with **no call-position requirement**, so "call" was genuinely too strong; and `strip_ocaml_comments_and_strings` strips both nested `(* … *)` and `"…"` contents with escape handling, so the new text is **stronger**, not an underclaim. Checked the set — `codegen-ptx` and `type-safety` already carried the correct wording, so this was a single stale copy, not three.
4. **`convergence-safety/STATUS.md`** — found while fixing (3): it claimed the gate *enforces* `known_gap`, but `known_gap` never appears in `check-production-link.py`. Reworded to say the gate enforces the split via `modules` (empty here → 0 shipped-linked) and treats `known_gap` as documentary. The other two `STATUS.md` files make no such claim.

**Deferred — 6 findings** (`type-safety/production-link.json:17`, `check-production-link.py:182/240/245/284`, `check-production-link.test.sh:154` + `kb/properties.md:83`)

Each needs new checker logic — schema/type validation, duplicate-short-name rejection, path-traversal containment, OCaml quoted-string stripping — **plus** new red-path cases, per this repo's own `KB-GATE-SELF` doctrine that a gate not proven able to fail is not trusted. That would move `check-production-link.test.sh` off the **13/13** this change is pinned to, which is a fair signal they belong to a separate change rather than a wording round.

Worth stating explicitly: the tempting in-scope shortcut for `type-safety/production-link.json:17` — dropping `TypeSafetySpec`/`UnifySpec` from `modules` — would take the shipped-linked total **from 94 to 50**, silently gutting the figure this whole change exists to publish. That gap is already disclosed in the manifest's own `no_anti_drift_check` field as known, deferred work.

**Unchanged and re-verified:** `TOTAL: 94` / `model-only: 186` / `total: 280` reproduce from a live run; 94 appears 6× and 186 3× (unchanged after the earlier deliberate copy reduction). Gates: `check-kb-properties.sh` 0 (45 declarations) · `check-production-link.test.sh` 13/13 · `check-cited-paths-exist.sh` 0.

mathiasbourgoin:
### Five of the six deferred findings are now implemented, not deferred — `884a5f08`

Reversing the earlier deferral: all five tractable Major findings are fixed with red-path coverage. Test count **13 → 26**.

| Finding | Fix |
|---|---|
| `:182` non-code / non-exact matches | `strip_ocaml_comments_and_strings` now also blanks `{\|…\|}` and `{tag\|…\|tag}` (raw contents, exact matching close). The trailing `\b` is replaced by a **both-sided** lookaround over `[A-Za-z0-9_']`, so `infer'` and `infer_x` no longer satisfy a claim for `infer`. |
| `:240` schema validation | `schema != 1` (including absent) → **exit 2**; `modules` not a JSON object → **exit 2**, replacing an unhandled `AttributeError`. |
| `:245` ambiguous bare names | Duplicate `module_short_name()` in a ledger → **exit 2**, naming both colliding full paths, instead of silently overwriting. |
| `:284` `test_file` confinement | Resolved via `realpath` and confined with `commonpath` to `formal/<project>/`; `../`, absolute paths **and symlinks pointing outside** → exit 1. |
| `test.sh:154` / `kb/properties.md:83` | 13 new cases; `KB-GATE-PRODUCTION-LINK` row rewritten to match the code exactly. |

**Exit-code split:** exit 2 = precondition failure, nothing can be classified yet (schema, `modules` type, ambiguous short names). exit 1 = a specific claim was checked and found false (live-reference failures, path confinement).

**The headline numbers did not move.** Verified the strict way rather than by re-running the new script alone: the **pre-change** script was run against the same tree and the full outputs diffed **byte-for-byte, including every per-module row** (`PtxLayout` 50, `TypeSafetySpec` 17, `UnifySpec` 27). Nothing was reclassified in either direction — which matters, because an over-strict identifier boundary would silently push a genuinely-linked module into model-only and two such errors could cancel while the totals still read 94/186/280.

**Adversarially checked, not just unit-tested:** an internal `|}` or wrong-tag `|bar}` does not terminate a quoted string early; an unterminated `{|` swallows to EOF without hanging; nested `(* … *)` and `\"`-escaped strings are unregressed; the `.` in `Sarek_typer.infer` is `re.escape`d so `Sarek_typerXinfer` does not match; a symlink **inside** the project pointing outside is caught by `realpath`. Prove-red reproduced independently for `:182` and `:284`.

Gates: `check-production-link.py` exit 0 (94/186/280) · `check-production-link.test.sh` 26/26 · `check-kb-properties.sh` exit 0 (45 declarations) · `check-cited-paths-exist.sh` exit 0. Python 3.14.6.

---

### The sixth finding (`formal/type-safety/production-link.json:17`) is still open, deliberately

`TypeSafetyModel.ml` / `UnifyModel.ml` are counted as shipped-linked but are never regenerated or drift-checked from the Rocq sources, so **44 of the 94 may be linked to a stale extraction**. The manifest's own `no_anti_drift_check` field already admits this.

The two remedies are not equivalent, and neither is a review-response edit:

- **build it** — extraction drivers for `type-safety` plus generalising the drift byte-compare beyond its `basename == codegen-ptx` guard; keeps 94 honestly;
- **exclude them** — one manifest edit, and the published figure drops **94 → 50**.

That is a call about what number the project publishes, so it is left to a human rather than taken because it is the cheaper edit. `modules` is confirmed untouched in this range.

PR #419, opened 2026-08-01
fix(scripts): FORMAL_ALLOW_DIRTY no longer loses your work; restore_tree restores the build-owned set (backlog-229)

mathiasbourgoin:
> `ci.yml` change **authorised by Mathias, 2026-08-01**. `.claude/rules/escalation.md`
> names workflow edits as requiring explicit human confirmation; it was given.

`scripts/check-formal-proofs.sh` installs an exit trap that runs `git checkout -- .`
inside each `formal/` project. It fired **even under `FORMAL_ALLOW_DIRTY=1`** — the
flag bypassed only the pre-flight check, not the cleanup. **It silently destroyed
an agent's uncommitted edits during this campaign.** The false claim was in the
**flag name**, and the cost was deleted work.

## Why the first fix was not enough — twice

The original fix made the trap's `git checkout -- .` conditional on the flag. Review
found that reasoning rested on an enumeration produced by
`grep 'git checkout|git restore|git clean'` — **an instrument that by construction
can only find what it already expects.** Reality: **nine** working-tree-mutating
sites, **six** touching tracked content:

| site | tracked? | restored under the flag, before this PR? |
|---|---|---|
| `rm -f .lia.cache .nia.cache` | **1 tracked** | no — runs before the flag check |
| `git checkout -- .` | — | the only one counted |
| `find … -exec rm -f {} +` | **145 tracked** | **no** |
| `rocq makefile -o CoqMakefile` | yes | **no** |
| `make -f CoqMakefile` (extraction rewrites) | yes | **no** |
| `canonicalize-extraction.py` (in-place) | yes | **no** |

So under the flag, **145 tracked artefacts stayed deleted** — and those ` D ` entries
survive the pre-flight's `grep -v '^?? '`, meaning **the next unflagged run was
refused.** An absorbing state, verbatim the failure the trap's own comment says it
exists to prevent. Reproduced in a fixture.

Then the *second* fix repeated the pattern: it justified an unconditional
`rm -f .lia.cache .nia.cache` by calling them **"untracked residue."**
`git ls-files formal` says otherwise — **`formal/convergence-safety/.lia.cache` is
tracked** (committed `9655c874`), and `ConvergenceSemantics.v` uses `lia`, so a build
regenerates it. Same absorbing state through a **one-file door instead of 145**.

**Check the tree, not the comment.**

## What this does

`restore_tree` now restores the **build-owned path set** — `*.vo`, `*.vok`, `*.vos`,
`*.glob`, `*.CoqMakefile.d`, `CoqMakefile`, `CoqMakefile.conf`, `extraction/*.ml|mli`,
`.lia.cache`, `.nia.cache` — **unconditionally, flag or no flag.** That set is exactly
what the script itself deletes and regenerates, so it is disjoint from a developer's
source edits **by construction**, without needing to classify a diff. The flag now only
widens *entry*; the trap's *scope* is identical either way.

**One `git checkout --` call per glob, deliberately.** A single call with any unmatched
pathspec exits non-zero and applies **none** of the patterns — verified empirically: a
combined call on a repo where only `.lia.cache` is tracked exits 1 and restores nothing,
while the per-pattern loop restores it correctly with `.nia.cache` matching nothing.

**Also CRITICAL, and it would have turned CI red:** wiring the new test into `ci.yml`
without a `gate-red-path` row violated `KB-GATE-INVENTORY`. `check-kb-properties.sh`
exited **1** on the branch and **0** at base — and that gate runs *before* the harness
step in the same job, so the branch would have failed CI **and the new test would never
have run.** Fixed by giving `KB-GATE-FORMAL-ADMIT` a `red_path`; the rule credits a
script declared as either `tool` or `red_path`, confirmed by reading the checker.

## Messages corrected

*"your uncommitted edits are left in place rather than discarded at exit"* was
**demonstrated false**: a tracked file holding `MY-PRECIOUS-EDIT` came out
`BUILD-OVERWROTE-THIS`. The flag protects edits from **the trap**, not from **the
build**. Likewise *"restores nothing"* was contradicted two lines later by the
unconditional cache `rm`. All now state the mechanism, and the **"145-artefact" count
was deleted rather than bumped to 146** — a number repeated in prose will drift.

The one deliberate width: `extraction/*.ml|mli` is restored unconditionally, so a
hand-edit there is lost. That is **stated**, not silent.

## Tests

9 cases. Case-d pins the tracked-cache path: a flagged run must restore it (asserted by
**content** and by absence of a `^ D` porcelain line, not exit code), and the next
unflagged run must not be refused. That last assertion is genuinely non-vacuous — both
runs exit 1 because the stub dies at `rocq makefile`, but **neither prints the
dirty-tree error**, so it cannot pass by accident.

Prove-red: removing the two cache globs takes the suite 9/0 → **6 passed / 3 failed**,
with `<MISSING>` content, the ` D ` porcelain line, and the refusal all present.

## Verified, not assumed

Flag spelling is **exact-complementary** over every string value — no spelling passes the
entry gate that the trap then rejects. Disjoint from `backlog/201-theorem-count`, which
appends after line 428 while this edits the pre-flight block and `restore_tree`. The
`ci.yml` line needs **no Rocq** (the stub satisfies `--version` and dies later) and **no
GPU**, and creates no new required check.

Gates: `check-kb-properties.sh` 0 (44 declarations) · `check-formal-proofs.test.sh` 0
(9/0) · `check-provered-kb-link.sh` 0 · `check-license-headers.sh` 0 ·
`check-no-machine-identifiers.sh` 0 · `prove-red.sh` 0 (9 subjects, 58 mutations).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved formal-proof checks to safely restore generated build files and tracked cache files without overwriting unrelated local changes.
  * Clarified behavior when the working tree contains changes, including an opt-in mode for preserving existing edits.

* **Tests**
  * Added automated coverage for clean and modified working trees, build-generated changes, cache restoration, expected results, and failure handling.

* **Documentation**
  * Updated formal verification guidance to describe dirty-tree restoration and test coverage.


mathiasbourgoin:
Superseded by a rebased branch. Eight merges landed on `main` after this branch was cut from `ed245581`, and `kb/properties.md` conflicted — #226 and #222 edited `KB-GATE-CITED-PATHS` and `KB-GATE-LICENSE` while this branch edits `KB-GATE-FORMAL-ADMIT`.

Rebased with `rerere` disabled (its cache is shared across worktrees and can replay a resolution from another branch silently). The conflict was resolved mechanically after **proving the two sides changed disjoint rows** — no hand-editing of conflict markers. Verified afterwards: the three script/workflow files are sha256-identical to the reviewed `f2be4bd6`, the `FORMAL-ADMIT` row matches the reviewed version, `LICENSE`/`CITED-PATHS` match `main`, and no other row differs from `main`. 44 rows, valid JSON, unique ids.

